Issue #27, opened 2021-12-07
ENHANCEMENT: Implement velocity reading into trajectory converters


RastislavTuranyi:
Currently, MDANSE saves velocities into NetCDF file only when converting from DL POLY files. However, trajectories generated with other software also contain velocity data, which is not being saved. Because of this, interpolation has to be used when computing functions that depend on velocities, and current correlation function cannot be computed at all since it does not implement interpolation. Therefore, the following steps should be undertaken:

1. Check which trajectories contain velocity data.
2. Extend the relevant converters to save velocities.

MBartkowiakSTFC:
At the moment, velocities are written into the output trajectory by the following converters:
All the universal ones: ASE, MDAnalysis, MDTraj
Some of the specific ones: Castep, CP2K, Gromacs, LAMMPS, Forcite.

If we discover that some of the other MD engines can write velocities into their output files, new issues will be created to deal with it.

PR #39, opened 2022-03-03
Fix velocity reading in CASTEP converter

### LINKED ISSUE #38: BUG: Fix velocity reading into CASTEP converter
Specific implementation of #27 for the CASTEP converter, except in this case the reading and saving of velocities and gradients from md files is already supported. However, this data is for some reason inaccessible once the created NetCDF file is loaded into MDANSE.

RastislavTuranyi:
**Description of work.**

Fixes #38, making progress on #27 

Currently, CASTEP converter correctly reads velocities and forces from MD files, but it does not save them into NetCDF file correctly. The cause of this has been identified as the fact that forces are saved as [energy] gradients, though the exact reason why MMTK has issues with gradients is unknown. By saving forces as forces, the same way it is done in the DL_POLY converter which works well, the CASTEP converter now correctly saves the read velocities into the .nc file, which then can be used in analyses.

**To test:**

1. Convert an MD file using the CASTEP converter.
2. Use that converted NetCDF file for the Current Correlation Function (or other ones which require velocities) analysis.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

- Do changes function as described? Add comments below that describe the tests performed?
- Do the changes handle unexpected situations, e.g. bad input?
- Has the relevant (user and developer) documentation been added/updated?

### DIFF
--- a/Src/Framework/Jobs/Castep.py
+++ b/Src/Framework/Jobs/Castep.py
@@ -190,7 +190,7 @@ class CASTEPConverter(Converter):
         self._universe.initializeVelocitiesToTemperature(0.)
         self._velocities = ParticleVector(self._universe)
 
-        self._gradients = ParticleVector(self._universe)        
+        self._forces = ParticleVector(self._universe)
 
         # A MMTK trajectory is opened for writing.
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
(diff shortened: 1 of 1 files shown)


Src/Framework/Jobs/Castep.py
```diff
@@ -225,13 +225,13 @@ def run_step(self, index):
         
         data = {"time": timeStep}
 
-        # Retrieve the positions multiplied by Units.Bohr*Units.Hartree/Units.hbar and save them as universe velocities
+        # Retrieve the velocities multiplied by Units.Bohr*Units.Hartree/Units.hbar and save them
```
sanghamitra-mukhopadhyay, line 228: Check the unit of velocities and whether it is has been correctly read and saved

RastislavTuranyi: Hi, I have kept all the units as they were originally, but since you asked; according to https://www.tcm.phy.cam.ac.uk/castep/MD/node13.html, all the variables in MD files are in Hartree units. Therefore,
- to convert positions, the given value has to be multiplied by the value of Bohr radius, which is what happens on line 151 `config[0:self["n_atoms"], :] *= Units.Bohr`;
- to convert velocities, the given values have to be multiplied by the value of Bohr radius and Hartree energy as well as divided by reduced Planck's constant, which is what happens on line 152: `config[self["n_atoms"]:2*self["n_atoms"], :] *= Units.Bohr*Units.Hartree/Units.hbar`;
- to convert forces, the given values have to be multiplied by the value of Hartree energy and divided by Bohr's radius, which is what happens on line 153: `config[2*self["n_atoms"]:3*self["n_atoms"], :] *= Units.Hartree/Units.Bohr`.

 Please let me know if there is any mistake anywhere!

Issue #46, opened 2022-03-22
BUG: Conversions that convert velocities generate unexpected and broken variables in NetCDF trajectory


RastislavTuranyi:
When trajectory conversion is performed using a converter that can convert velocities on a file that contain velocities, the following variables exist in the resulting NetCDF trajectory:

- description
- box_size
- step
- time
- configuration
- velocities
- temperature
- kinetic_energy

The first 6 are expected and should indeed appear, however, temperature and kinetic energy are often not even present in the initial trajectory, and there is no code in MDANSE that parses them even if they were, so they should not appear in the NetCDF trajectory. Even worse, these two unexpected variables are broken: when plotted, there is no x-axis variable, only units of au, which makes no sense. The data itself appears approximately correct, as can be seen below (unexpected graph above, actual temperature calculation using the same trajectory below)
![image](https://user-images.githubusercontent.com/78977041/159481107-89f0b9ad-5522-40a9-a1fe-e0584071dede.png)
![image](https://user-images.githubusercontent.com/78977041/159481161-830ef666-c98c-4584-a43c-a296632def2c.png)

PR #50, opened 2022-04-05
Only the configuration (and velocities and gradients when available) …

eurydice76:
I fixed the issue #46. @RastislavTuranyi could you please double check that everything is OK as I am a bit rusted in MDANSE !

### DIFF
--- a/Src/Framework/Jobs/Castep.py
+++ b/Src/Framework/Jobs/Castep.py
@@ -227,7 +227,7 @@ class CASTEPConverter(Converter):
 
         # Retrieve the velocities multiplied by Units.Bohr*Units.Hartree/Units.hbar and save them
         self._velocities.array = config[nAtoms:2*nAtoms, :]
-        self._universe.setVelocities(self._velocities)
+        data["velocities"] = self._velocities
 
         # Retrieve the forces multiplied by Units.Hartree/Units.Bohr and save them
         self._forces.array = config[2 * nAtoms:3 * nAtoms, :]
--- a/Src/Framework/Jobs/Castep.py
+++ b/Src/Framework/Jobs/Castep.py
@@ -195,8 +195,10 @@ class CASTEPConverter(Converter):
         # A MMTK trajectory is opened for writing.
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
 
+        data_to_be_written = ["configuration","time","velocities","gradients"]
+
         # A frame generator is created.
-        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["configuration","velocities","gradients"],
+        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, data_to_be_written,
                                                                                      0, None, 1)])
     
     def run_step(self, index):
--- a/Src/Framework/Jobs/Castep.py
+++ b/Src/Framework/Jobs/Castep.py
@@ -196,7 +196,7 @@ class CASTEPConverter(Converter):
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
 
         # A frame generator is created.
-        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["all"],
+        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["configuration","velocities","gradients"],
                                                                                      0, None, 1)])
     
     def run_step(self, index):
@@ -231,7 +231,7 @@ class CASTEPConverter(Converter):
 
         # Retrieve the forces multiplied by Units.Hartree/Units.Bohr and save them
         self._forces.array = config[2 * nAtoms:3 * nAtoms, :]
-        data["forces"] = self._forces
+        data["gradients"] = self._forces
 
         # Store a snapshot of the current configuration in the output trajectory.
         self._snapshot(data=data)                                          
--- a/Src/Framework/Jobs/DCDConverter.py
+++ b/Src/Framework/Jobs/DCDConverter.py
@@ -300,7 +300,7 @@ class DCDConverter(Converter):
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
         
         # A frame generator is created.        
-        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["all"], 0, None, 1)])
+        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["configuration"], 0, None, 1)])
 
     def run_step(self, index):
         """
--- a/Src/Framework/Jobs/DCDConverter.py
+++ b/Src/Framework/Jobs/DCDConverter.py
@@ -299,8 +299,10 @@ class DCDConverter(Converter):
         # A MMTK trajectory is opened for writing.
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
         
+        data_to_be_written = ["configuration","time"]
+        
         # A frame generator is created.        
-        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["configuration"], 0, None, 1)])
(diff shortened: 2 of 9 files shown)


eurydice76:
Now that velocities and gradients are set in CASTEP, DL_POLY, FORCITE we should check that the they are in properly set with MDANSE units. @gonzalezma could you please check ?

Src/Framework/Jobs/Castep.py
```diff
@@ -231,7 +231,7 @@ def run_step(self, index):
 
         # Retrieve the forces multiplied by Units.Hartree/Units.Bohr and save them
         self._forces.array = config[2 * nAtoms:3 * nAtoms, :]
-        data["forces"] = self._forces
+        data["gradients"] = self._forces
```
RastislavTuranyi, line 234: Hi, are you certain this should be gradients? My understanding of MMTK is that this variable denotes the energy gradients, not forces.

gonzalezma: Hi both. From what I see, in MMTK "gradients" stands for the energy gradients at each atom position and as @eurydice76 says, the force is the negative gradient of the potential energy. From this point of view, I would imagine that data["gradients"] = -self._forces. But do we use "gradients" in any analysis? And "forces"? Can we have both and do we need both?

RastislavTuranyi (changes requested):
Firstly, thanks for making all the changes! However, you have forgotten to include time when setting the `SnapshotGenerator`. In all files, `dataToBeWritten = ["configuration"]` has to be changed to `dataToBeWritten = ["configuration", "time"]`.

Other than that, everything seems OK, though I have a small gripe with the name `dataToBeWritten` since it does not comply with PEP8 and as such is unpythonic. What do you think?

Src/Framework/Jobs/Castep.py
RastislavTuranyi: I see, thank you for the answer! To the best of my knowledge, no current analysis uses gradients or forces. Also, we technically cannot have a variable called 'forces' since MMTK does not recognise such variable, and so essentially does not write it into the NetCDF trajectory.

eurydice76:
Thanks for the feedback. You right, time should be included. My bad. Being a PEP specialist, could you tell me what should be the variable name ?

RastislavTuranyi:
[PEP8](https://peps.python.org/pep-0008/#function-and-variable-names) states the following:

```
Function and Variable Names
Function names should be lowercase, with words separated by underscores as necessary to improve readability.

Variable names follow the same convention as function names.

mixedCase is allowed only in contexts where that’s already the prevailing style (e.g. threading.py), to retain backwards compatibility.
```

 so I think it should be `data_to_be_written`.

eurydice76:
OK, let's change this

Src/Framework/Jobs/DCDConverter.py
```diff
@@ -300,7 +300,7 @@ def initialize(self):
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
         
         # A frame generator is created.        
-        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["all"], 0, None, 1)])
+        self._snapshot = SnapshotGenerator(self._universe, actions=[TrajectoryOutput(self._trajectory, ["configuration"], 0, None, 1)])
```
RastislavTuranyi, line 303: This needs time as well.

Src/Framework/Jobs/Discover.py
```diff
@@ -244,15 +244,14 @@ def initialize(self):
...
-        self._universe.setVelocities(self._velocities)
         
         self._universe.foldCoordinatesIntoBox()
             
         # A MMTK trajectory is opened for writing.
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w', comment=self._hisfile["title"])
 
         # A frame generator is created.
-        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["all"], 0, None, 1)])
+        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["configuration","velocities"], 0, None, 1)])
```
RastislavTuranyi, line 254: This needs time as well.

Src/Framework/Jobs/Gromacs.py
```diff
@@ -66,7 +66,7 @@ def initialize(self):
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
 
         # A frame generator is created.
-        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["all"], 0, None, 1)])
+        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["configuration"], 0, None, 1)])
```
RastislavTuranyi, line 69: This needs time as well.

Src/Framework/Jobs/LAMMPS.py
```diff
@@ -163,15 +163,15 @@ def initialize(self):
         self._nameToIndex = dict([(at.name,at.index) for at in self._universe.atomList()])
 
         # A frame generator is created.
-        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["all"], 0, None, 1)])
-
-        # Estimate number of steps if needed
-        if self.numberOfSteps == 0:
-            self.numberOfSteps = 1
-            for line in self._lammps:
+        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["configuration"], 0, None, 1)])
```
RastislavTuranyi, line 166: This also needs time.

Src/Framework/Jobs/VASP.py
```diff
@@ -181,7 +182,7 @@ def initialize(self):
         self._trajectory = Trajectory(self._universe, self.configuration['output_files']['files'][0], mode='w')
 
         # A frame generator is created.
-        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["all"], 0, None, 1)])
+        self._snapshot = SnapshotGenerator(self._universe, actions = [TrajectoryOutput(self._trajectory, ["configuration"], 0, None, 1)])
```
RastislavTuranyi, line 185: This also needs time.

RastislavTuranyi (changes requested):
Thank you for the changes, but you forgot to add "time" to converters that do not convert velocities/forces.

Src/Framework/Jobs/Discover.py
eurydice76: Sorry for the bug. I have been too fast. I should have double checked. Let's fix this.

Src/Framework/Jobs/Castep.py
eurydice76: Thtats' why I turned forces to gradients in order that MMTK can "recognize" them. However, as you said Rastislav, no analysis uses the force/gradient. So we should perhaps not worry too much about this right now.

RastislavTuranyi (approved):
I have added the "time" to Gromacs and fixed typos in some variable names, and with that I think everything should be OK. Not sure if anyone has noticed anything else, but it seems all good to me now.

RastislavTuranyi (changes requested):
Actually, I have just realised that we need to work on the tests before this can be merged. When all tests are run, the following ones fail (also the PDF, but that is unrelated to this PR):

- CASTEP
- DFTB
- Discover
- DMOL
- Forcite

This is because the reference trajectories used for these tests contain the variables which have been removed in this PR. That said, I am not sure what is the best way to solve this; simply generate new reference trajectories using a build from this PR? Perhaps a better solution would be to alter the BuildJobTests.py file in the `__create_test()` method so that the variables that have been removed in this PR are skipped during the testing. Or possibly there is another, better, solution. What do you think?

eurydice76:
Oups I am sorry Rastislav but I merged before you sent me the message.

RastislavTuranyi:
Apologies too for noticing only so late. I think we could still use this branch to fix the tests and merge again, or we could either create a new branch or use the [fix_tests](https://github.com/ISISNeutronMuon/MDANSE/tree/fix_tests) branch which I am using some of the other tests that are broken.

Also, what do you think would be the best way to fix these tests?

eurydice76:
Perhaps using the fix_tests dedicated branch is the way to go.

PR #51, opened 2022-04-05
added info method to NetCDFInputData

### LINKED ISSUE #45: ENHANCEMENT: include more information in Data Info
Currently, the Data Info window contains the following information:

- Path to the trajectory
- Number of steps
- Universe
- Direct cell
- Reciprocal cell
- Molecules

This is all useful, but there is other information that would be useful to know about a given trajectory/analysis results. So far, I have thought about the following data to add, but if anyone has any suggestions, please share!

- variables stored in the trajectory

eurydice76:
Fixes #45 

@RastislavTuranyi I added some data info for MMTK trajectory and NetCDF file object as required in issue #45. Is it OK for you ?


RastislavTuranyi (approved):
Hi, yes, this is gives exactly the result I had in mind, thank you! If there is nothing else you or anyone else would like to include, I am happy with this.

Issue #42, opened 2022-03-04
ENHANCEMENT: Increase the default size of the variable box in 2D/3D Plotter


RastislavTuranyi:
At the moment, the variable box in 2D/3D Plotter is only large enough to fit one row by default, which makes selecting a variable tedious, especially in analyses with lots of variables like PDF. Therefore, the default size should be increased at the file box's expense.

Furthermore, the only way to change the size of the variable box is to resize the whole Plotter window. Manual resizing of these boxes should be implemented. This would work similarly to how the various panels in the main window can be resized.

eurydice76:
I have not practised `wxPython` for years but I will give a try because I experienced the described behavior and it is very annoying.

eurydice76:
I added a splitter window between the dataset and the data lists and also increased a bit the minimal size of the mainn frame. What do you think ?

RastislavTuranyi:
Looking at the GUI, the initial size is increase is great, and the ability to resize is also useful. However, do you think it would be possible to limit the ability to resize to prevent the two parts of the Data panel from overlapping? Currently, it is possible to essentially delete the 'variables' part of the Data panel, like so:
![image](https://user-images.githubusercontent.com/78977041/162001210-31215791-31d9-4cb2-9ab2-fc9da006c550.png)
To reproduce, just resize all the way down, and then maximize the plotter window. This completely hides the 'variables' part and also hides the ability to resize.

Similar effect can be achieved by resizing the 'variables' part all the way up, which hides the 'filename', though this time the ability to resize remains intact. Still, I think this could cause confusion.
![image](https://user-images.githubusercontent.com/78977041/162002197-b444f510-66dc-4285-851b-176627b994a9.png)

eurydice76:
ok, I will try to solve this

eurydice76:
I could resolve the bottom to up problem, but the problem still exists for the top to bottom direction. I suspect a bug on wxPython or at least something I misunderstand from their API.

eurydice76:
I think it is OK now. The dataset and data panels now have a minimal size. Does it fulfill your expectations ?

RastislavTuranyi:
I have some comments that I would like to discuss, but I think it would be easier do it using the functionalities that a PR provides. Can you please open one? Thanks!

eurydice76:
I just created the PR

PR #55, opened 2022-04-07
42 improve pane layout in plotter

eurydice76:
In order to solve the issue #42. I added in case of standalone mode a splitter window which contains the dataset and data panel. A minimum size of 100 has been set for each of these panels in order to avoid then to collapse.

### DIFF
--- a/Src/GUI/Plugins/PlotterPlugin.py
+++ b/Src/GUI/Plugins/PlotterPlugin.py
@@ -110,26 +110,30 @@ class DataPanel(wx.Panel):
         
         sizer0 =  wx.BoxSizer(wx.VERTICAL)
         
-        splitterWindow = wxsplitter.MultiSplitterWindow(self.setup)
-        splitterWindow.SetOrientation(wx.VERTICAL)
-
         if self.standalone:
+            splitterWindow = wx.SplitterWindow(self.setup, style=wx.SP_LIVE_UPDATE)
+            splitterWindow.SetMinimumPaneSize(100)
+
             self.datasetlist = wx.ListCtrl(splitterWindow, wx.ID_ANY,style = wx.LC_REPORT|wx.LC_SINGLE_SEL)
             self.datasetlist.InsertColumn(0, 'key', width=100)
             self.datasetlist.InsertColumn(1, 'filename', width=100)
             self.datasetlist.InsertColumn(2, 'path', width=500)
-            splitterWindow.AppendWindow(self.datasetlist)
             
             self.Bind(wx.EVT_LIST_ITEM_SELECTED, self.on_select_dataset,  self.datasetlist) 
             self.Bind(wx.EVT_LIST_KEY_DOWN, self.delete_dataset, self.datasetlist)
 
-        self.datalist = wx.ListCtrl(splitterWindow, wx.ID_ANY,style = wx.LC_REPORT|wx.LC_SINGLE_SEL)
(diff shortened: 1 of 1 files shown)


Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -109,43 +111,54 @@ def build_panel(self):
         sizer0 =  wx.BoxSizer(wx.VERTICAL)
         
         if self.standalone:
-            self.datasetlist = wx.ListCtrl(self.setup, wx.ID_ANY,style = wx.LC_REPORT|wx.LC_SINGLE_SEL)
+            splitterWindow = wx.SplitterWindow(self.setup, style=wx.SP_LIVE_UPDATE)
+            splitterWindow.SetMinimumPaneSize(100)
```
RastislavTuranyi, line 115: I was wondering if a value of 50 would not be more appropriate here. That is just about the height of the header of the table, so it will still prevent the overlapping issue, while at the same time giving users greater freedom with how the two parts of the Data panel are distributed. What do you think?
The current minimum size: 
image
The proposed minimum size:
image

Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -378,6 +391,7 @@ def __init__(self, parent, title="2D/3D Plotter"):
         wx.Frame.__init__(self, parent, wx.ID_ANY, title, style=wx.DEFAULT_DIALOG_STYLE|wx.MINIMIZE_BOX|wx.MAXIMIZE_BOX|wx.RESIZE_BORDER)
         self.__build_dialog()
         self.__build_menu()
+        self.SetMinSize((-1,800))
```
RastislavTuranyi, line 394: Is there a reason why this has to be added? Adding it means that the size of the Plotter window cannot be reduced vertically at all, which can cause issues on smaller screens (like laptops). Further, it changes the current behaviour of the GUI (which allows resizing in all directions), which could be detrimental/confusing for users who expect/rely on the current behaviour.

RastislavTuranyi (commented):
Thank you for your work! This definitely looks much better than it does in develop, but what do you think of these comments?

Src/GUI/Plugins/PlotterPlugin.py
RastislavTuranyi: I see, in that case I would prefer if this were removed. Thanks!

eurydice76:
Hi @sanghamitra-mukhopadhyay , can you please review in order that I can proceed to squash merge ? Thanks

RastislavTuranyi (approved):

RastislavTuranyi:
Wait, am I seeing things or have you reverted all your changes? The changes tab shows blank:
image

RastislavTuranyi (changes requested):
I have pulled the latest changes on this branch, and indeed everything has been reverted; the GUI is back to how it had been before. GitHub also shows that no changes have been made. I am unsure what happened; can you please check? Thank you!

eurydice76:
You right everything has been reverted. What happened ??? I do not have a single clue. I put back the code and keep you in touch

RastislavTuranyi (approved):
Well, regardless of what happened, everything seems to be back, except for the default size of the variable part, which I believe you have changed at the beginning.
image

sanghamitra-mukhopadhyay (approved):
I find it okay. so approve.

PR #60, opened 2022-04-11
Make the names of output files more descriptive

### LINKED ISSUE #41: ENHANCEMENT: Change default name of output files
Currently, the default name of output files of all analysis is `output_<trajectory>` where <trajectory> is the name of the trajectory file. This default name is highly undescriptive, requiring to be changed every time an analysis is performed because if it is not, the file will simply get overwritten. A better name would be `<trajectory>_<analysis>` where <analysis> is an abbreviated name of the analysis performed. Even better, MDANSE could also check if the default file name already exists, and if it does, add (1) etc at the end to prevent overwriting.

RastislavTuranyi:
**Description of work.**

Resolves #41 

The default names of output files in MDANSE are, at the moment, too generic and unhelpful, and so they have to be changed during every analysis and conversion. This work attempts to fix that by making the default names more relevant, such that they can often be usable as is, speeding up the workflow and removing this little annoyance. This has been implemented

- for analyses immediately when the window is opened so that the name is `<NetCDF_trajectory_name>_<analysis_acronym>` (e.g. `traj_pdf`)
- for trajectory converters once the input files have been loaded

Furthermore, an algorithm has been implemented which prevents file overwriting by checking if the default filename already exists, and if it does, it appends (n), where n is the lowest number that doesn't yet exist. This does not work for file names that have been edited by the user, only new ones.

**To test:**

For trajectory converters:
1. Open a converter of choice
2. Click 'Browse' button and select a relevant file
3. Observe that the 'output files' field has changed from `~/output_trajectory_conversion` to the same path as the input file
4. Perform the conversion and wait for it to finish.
5. Click 'Browse' again and select the same file as before.
6. Observe that (1) has been appended to the file name.

For analyses:
1. Load a trajectory into MDANSE and open an analysis (e.g. pair distribution function)
2. Observe that the 'output files' field has a useful name.
3. Perform the analysis and wait for it to finish.
4. Close the analysis window and open a new one, for the same analysis.
5. Observe that (1) has been appended to the file name.

---

#### Reviewer ####

##### Code Review #####

##### Functional Tests #####

- Do changes function as described? Add comments below that describe the tests performed?
- Do the changes handle unexpected situations, e.g. bad input?
- Has the relevant (user and developer) documentation been added/updated?

### DIFF
--- a/Src/GUI/Widgets/OutputFilesWidget.py
+++ b/Src/GUI/Widgets/OutputFilesWidget.py
@@ -56,15 +56,15 @@ class OutputFilesWidget(IWidget):
                                     
         return (filename, formats)
     
-    def set_data(self, datakey):
+    def set_data(self, datakey, type):
 
         basename = "output"
                 
         if datakey is None:
-            basename = "output"
+            basename = "output_%s" % type
             trajectoryDir = os.getcwd()
         else:
-            basename = "output_%s" % os.path.splitext(os.path.basename(datakey))[0]
+            basename = "%s_%s" % (os.path.splitext(os.path.basename(datakey))[0], type)
             trajectoryDir = os.path.dirname(datakey)
             
         self._filename.SetValue(os.path.join(trajectoryDir,basename))
--- a/Src/GUI/Widgets/OutputFilesWidget.py
+++ b/Src/GUI/Widgets/OutputFilesWidget.py
@@ -13,6 +13,7 @@
 # **************************************************************************
 
 import os
+import os.path
 
 import wx.combo
 import wx.lib.filebrowsebutton as wxfile
@@ -56,28 +57,28 @@ class OutputFilesWidget(IWidget):
                                     
         return (filename, formats)
     
-    def set_data(self, datakey, type):
-
-        basename = "output"
+    def set_data(self, datakey, analysis):
                 
         if datakey is None:
-            basename = "output_%s" % type
+            basename = "output_%s" % analysis
             trajectoryDir = os.getcwd()
         else:
-            basename = "%s_%s" % (os.path.splitext(os.path.basename(datakey))[0], type)
-            trajectoryDir = os.path.dirname(datakey)
+            basename = "%s_%s" % (os.path.splitext(os.path.basename(datakey))[0], analysis)
+            trajectoryDir = os.path.dirname(datakey)        
 
-        path = os.path.join(trajectoryDir,basename)
-        for i in range(100):
-            if os.path.exists(''.join([path, '.nc'])) or os.path.exists(''.join([path, '.tar'])):
-                basename = ''.join([basename.split('(')[0], '(', str(i+1), ')'])
-                path = os.path.join(trajectoryDir, basename)
-            else:
-                break
-        else:
-            from random import randint
-            path = os.path.join(trajectoryDir, ''.join([basename.split('(')[0], '(', str(randint(100, 100000)), ')']))
-            
+        filesInDirectory = [os.path.join(trajectoryDir,e) for e in os.listdir(trajectoryDir) 
+                            if os.path.isfile(os.path.join(trajectoryDir,e))]
+        basenames = [os.path.splitext(f)[0] for f in filesInDirectory]
+
+        initialPath = path = os.path.join(trajectoryDir,basename)
+        comp = 1
+        while True:
+            if path in basenames:
+                path = '%s(%d)' % (initialPath,comp)
+                comp += 1
+                continue
+            break           
+
         self._filename.SetValue(path)
         
 REGISTRY["output_files"] = OutputFilesWidget
--- a/Src/GUI/Widgets/OutputFilesWidget.py
+++ b/Src/GUI/Widgets/OutputFilesWidget.py
@@ -66,7 +66,18 @@ class OutputFilesWidget(IWidget):
         else:
             basename = "%s_%s" % (os.path.splitext(os.path.basename(datakey))[0], type)
             trajectoryDir = os.path.dirname(datakey)
+
+        path = os.path.join(trajectoryDir,basename)
+        for i in range(100):
+            if os.path.exists(''.join([path, '.nc'])) or os.path.exists(''.join([path, '.tar'])):
+                basename = ''.join([basename.split('(')[0], '(', str(i+1), ')'])
+                path = os.path.join(trajectoryDir, basename)
+            else:
+                break
+        else:
+            from random import randint
+            path = os.path.join(trajectoryDir, ''.join([basename.split('(')[0], '(', str(randint(100, 100000)), ')']))
             
-        self._filename.SetValue(os.path.join(trajectoryDir,basename))
+        self._filename.SetValue(path)
         
 REGISTRY["output_files"] = OutputFilesWidget
--- a/Src/GUI/Widgets/OutputFilesWidget.py
+++ b/Src/GUI/Widgets/OutputFilesWidget.py
@@ -20,11 +20,33 @@ import wx.lib.filebrowsebutton as wxfile
 
 from MDANSE import REGISTRY
 
+from MDANSE.GUI import PUBLISHER
 from MDANSE.GUI.ComboWidgets.ComboCheckbox import ComboCheckbox
 from MDANSE.GUI.Widgets.IWidget import IWidget
 
+def get_unique_filename(directory,basename):
+
+    filesInDirectory = [os.path.join(directory,e) for e in os.listdir(directory) 
+                        if os.path.isfile(os.path.join(directory,e))]
+    basenames = [os.path.splitext(f)[0] for f in filesInDirectory]
+
+    initialPath = path = os.path.join(directory,basename)
+    comp = 1
+    while True:
+        if path in basenames:
+            path = '%s(%d)' % (initialPath,comp)
+            comp += 1
+            continue
+        return path
+
 class OutputFilesWidget(IWidget):
-            
+
+    def __init__(self, *args, **kwargs):
+
+        IWidget.__init__(self, *args, **kwargs)
+
+        PUBLISHER.subscribe(self.msg_input_file_set, 'input_file_loaded')
+
     def add_widgets(self):
 
         sizer = wx.BoxSizer(wx.VERTICAL)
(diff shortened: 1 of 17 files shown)


Src/GUI/Widgets/OutputFilesWidget.py
```diff
@@ -1,29 +1,52 @@
...
 import wx.combo
 import wx.lib.filebrowsebutton as wxfile
 
 from MDANSE import REGISTRY
 
+from MDANSE.GUI import PUBLISHER
 from MDANSE.GUI.ComboWidgets.ComboCheckbox import ComboCheckbox
 from MDANSE.GUI.Widgets.IWidget import IWidget
 
+def get_unique_filename(directory,basename):
```
RastislavTuranyi, line 27: This is not large, but I was wondering if this function should not be a method of the class instead, given that it is solely related to operations in the class. What do you think?

Src/GUI/Widgets/OutputFilesWidget.py
```diff
@@ -56,17 +79,31 @@ def get_widget_value(self):
...
+
+        configurator, filename = message
+
+        if self._configurator.root is None:
+            return
 
-        basename = "output"
+        if self._configurator.root == configurator.name:
+            filenameWithoutExt = os.path.splitext(filename)[0]
+            outputFileWithoutExt = '%s_converted' % filenameWithoutExt
```
RastislavTuranyi, line 91: I don't like the '_converted' part of the name; I always call my converted trajectories by the same name as the initial ones. Since during trajectory conversion the file extension changes, I think we could drop that part and just use the original name. What do you think?

eurydice76: Yeap let's drop the _converted part

eurydice76: wait a minute. If you try on your side to remove the _converted you will see that it triggers some nasty side effect of the get_unique_filename method. Not a bug because it is an expected behavior yet not desirable. Indeed, as the get_unique_filename method works only with basename without extensions, if you pick for instance the DL_POLY input file FIELD_methanol, you will end up with a proposed output file name of FIELD_methanol(1). The use of _converted suffix or whatever suffix would save us from this.

RastislavTuranyi: Well, in that I case I would prefer to check using the extensions as well rather than keeping the _converted in. Also, now that you have mentioned DL_POLY, it might be a good idea to remove the FIELD_ part from the suggested file name since that's the equivalent to an extension.

eurydice76: The problem with the extensions is that the selected ones are not known at the time the dialog is popped up. That's the exact reason why at this level I work only with the basename.

For The FIELD, I can not remove it because that would make the OutputFilesWidget not generic anymore

RastislavTuranyi: The smallest intervention to the code to achieve the former is probably to generate the basenames like so:
```
basenames = [os.path.splitext(f)[0] for f in filesInDirectory
                 if os.path.splitext(f)[1] == '.nc' or os.path.splitext(f)[1] == 'tar']
```

eurydice76: Moreover coming back to the extensions, if you select more than one extension, it can happen that one of the filename with the extension exists while the others not. What should we do in such case ? We are stuck

RastislavTuranyi: Aah, I see what you mean. My solution is to use both, and select the higher number.

eurydice76: In the modified code that you showed, I dislike the fact that the extension are hardcoded. We are losing the generality of the widget.

RastislavTuranyi: If by unselecting you mean doing this,
image
then I got an error.

About the generality of our code; is it more important than user experience? I don't know much about this aspect of software development, but my instinct is to prioritise users over the code.

RastislavTuranyi: I think we don't need to hardcode the formats; there is the `self._configurator.formats` that we can access.

RastislavTuranyi: I was trying to get the formats through REGISTRY
```
temp = []
        for i in REGISTRY['formats']:
            temp.extend(i.extensions)
```
but it says that `REGISTRY['formats']` is an empty dictionary; do you know what's going on?

eurydice76: I can not reproduce the error you describe on ubuntu focal.

eurydice76: For the generality of the code, let's derive a brand new class dedicated to the converter output. The class can be standalone or a child class of the OutputFileWidget. That should do the job without having too specific code in general class. However, it still does not solve the problem of the formats.

RastislavTuranyi: It's strange that you are not getting an error when you have no formats selected, because, looking at the code, there definitely should be an error. There is no line in `OutputFilesConfigurator` that sets a certain format when none were selected. Instead, there is a line (71) that raises an error. I have confirmed on Windows 10 and Ubuntu 21.

RastislavTuranyi: Also, I got the issue with REGISTRY, I had a typo. It should be like this:
```
temp = []
for i in REGISTRY['format'].values():
    temp.extend(i.extensions)
```
This allows us to access all the output formats. The only issue is that the ASCII format creates a .tar file.

eurydice76: Thanks for the hint. However, if we consider that the output file of a converter is a very special case as it is an output file with a single extension (i.e. nc), deriving a specific widget would make sense. For instance, it would not contain the format combobox as the output extension is by construction nc. What do you think ?

eurydice76: yeap exactly

Src/GUI/Widgets/OutputFileWidget.py
```diff
@@ -0,0 +1,97 @@
...
+    def msg_input_file_set(self, message):
+
+        configurator, inputFilename = message
+
+        if self._configurator.root is None:
+            return
+
+        if self._configurator.root == configurator.name:
+            inputFileNameWithoutExtension = os.path.splitext(inputFilename)[0]
+            filename = '%s%s' % (inputFileNameWithoutExtension,REGISTRY['format'][self._configurator.format].extension)
```
RastislavTuranyi, line 81: I have noticed that the converter widget appends the .nc extension but the analysis widget does not; is there a reason for this? I think it would be best to be consistent.

Src/GUI/Widgets/OutputFileWidget.py
```diff
@@ -0,0 +1,97 @@
...
+            
+            path = OutputFileWidget._get_unique_filename(filename)
+            self._filename.SetValue(path)
+
+    def set_data(self, datakey):
+        
+        if datakey is None:
+            filename = os.path.join(os.getcwd(),'output.nc')
+        else:
```
RastislavTuranyi, line 90: I believe that exists only when an analysis is being performed performed. Therefore, this option should never occur in this widget.

Src/GUI/Widgets/OutputFileWidget.py
```diff
@@ -0,0 +1,97 @@
...
+        configurator, inputFilename = message
+
+        if self._configurator.root is None:
+            return
+
+        if self._configurator.root == configurator.name:
+            inputFileNameWithoutExtension = os.path.splitext(inputFilename)[0]
+            filename = '%s%s' % (inputFileNameWithoutExtension,REGISTRY['format'][self._configurator.format].extension)
+            
+            path = OutputFileWidget._get_unique_filename(filename)
```
RastislavTuranyi, line 83: Is there a reason to refer to the class with `OutputFileWidget` rather than `self` here?

Src/GUI/Widgets/OutputFilesWidget.py
```diff
@@ -56,17 +78,17 @@ def get_widget_value(self):
                                     
         return (filename, formats)
     
-    def set_data(self, datakey):
-
-        basename = "output"
+    def set_data(self, datakey, analysis):
                 
         if datakey is None:
```
RastislavTuranyi, line 83: I believe that `datakey` is only absent when a trajectory conversion is performed. Therefore, this option should never occur with this widget.

Src/GUI/Widgets/OutputFilesWidget.py
```diff
@@ -56,17 +78,17 @@ def get_widget_value(self):
...
+            basename = "output_%s" % analysis
             trajectoryDir = os.getcwd()
         else:
-            basename = "output_%s" % os.path.splitext(os.path.basename(datakey))[0]
+            basename = "%s_%s" % (os.path.splitext(os.path.basename(datakey))[0], analysis)
             trajectoryDir = os.path.dirname(datakey)
-            
-        self._filename.SetValue(os.path.join(trajectoryDir,basename))
+
+        path = OutputFilesWidget._get_unique_filename(trajectoryDir,basename)
```
RastislavTuranyi, line 90: Similarly, is there a reason to reference this class via `OutputFilesWidget` rather than `self` here?

RastislavTuranyi (commented):
Firstly, thank you for all this work! However, I can't help but think that `OutputFileConfigurator` and `OutputFilesConfigurator`, as well as `OutputFilesWidget` and `OutputFileWidget` have names that are too similar. Further, it is not clear what the difference between them is. I will try to alleviate the latter by adding docstrings to the classes, but I think it is worth considering renaming the classes so as to make them more distinct. What do you think?

Src/GUI/Widgets/OutputFilesWidget.py
eurydice76: You can use both. It is a question of styling. `_get_unique_name` being a staticmethod, I found more descriptive to use it through the class object.

Src/GUI/Widgets/OutputFilesWidget.py
eurydice76: About both configurators having names that are too close, what about renaming the `OutputFile` configure to `SingleOutputFile` ?

eurydice76:
About both configurators having names that are too close, what about renaming the OutputFile configure to SingleOutputFile ?

Src/GUI/Widgets/OutputFilesWidget.py
eurydice76: You right. As the converter now use the `OutputFileWidget` instead of the `OutputFilesWidget`, we can safely remove the the `datakey  None` case

RastislavTuranyi:
Sounds good!

eurydice76:
@sanghamitra-mukhopadhyay can we proceed to Merge ?

eurydice76:
@sanghamitra-mukhopadhyay can we proceed to merge ?

sanghamitra-mukhopadhyay (approved):

Issue #30, opened 2022-01-19
BUG: Scalar field plotter causes MDANSE to crash on Windows


RastislavTuranyi:
In 2D/3D Plotter, when 'Plot in new window' is clicked with Scalar field plotter selected, MDANSE freezes and after several seconds closes.

eurydice76:
The problem also happens with the Elevation plotter selected. I suspect that all the Plotter3D based plotters are impacted. The crash occurs when the page is added to the tab.

RastislavTuranyi:
I have traced the crash to l[ine 797](https://github.com/ISISNeutronMuon/MDANSE/blob/ffe60e4f78e78da198cb58b1091f777d9a611bac/Src/GUI/Plugins/Plotter3D.py#L797) of the Plotter3D.py file:
```
def center_on_actor(self,actor, out=False):#TODO out is useless : remove param
        Xmin,Xmax,Ymin,Ymax,Zmin,Zmax = actor.GetBounds()
        Xavg = (Xmin+Xmax)/2. 
        Yavg = (Ymin+Ymax)/2.
        Zavg = (Zmin+Zmax)/2.
        self.camera.SetFocalPoint(Xavg, Yavg, Zavg)
        self.camera.SetPosition(Xavg, Yavg, 3*Zmax)
```
This function is indeed used by all the 3D plotters, but I have not experienced any issues with any of them other than the Scalar Field one.

Given that we know that the newer versions of VTK (7.x and higher) introduce issues in MDANSE, perhaps this is one of these?

eurydice76:
Yeap. I can give a try with an older version of vtk and see. However, using the python debugger, I find the error at line 287 of the PlotterPlugin.py file. Very strange

eurydice76:
I tried on a machine installed through conda (see [wiki](https://github.com/ISISNeutronMuon/MDANSE/wiki/Installing-MDANSE-from-source-on-ubuntu@focal-using-conda-pip)) and version 6.3 of vtk and 3.0.2 of wxPython and surprisingly I have no crash. I can plot with all 3D plots available in MDANSE (elevation, isosurface, scalar field).

RastislavTuranyi:
[...]

PR #61, opened 2022-04-12
MD compliant plotter

### LINKED ISSUE #58: ENHANCEMENT: Create more granular plotting options for trajectory variables
Currently, the plotting of variables in MMTK NetCDF trajectories is far from useful, as documented in #34. All the variables in these trajectories do not have the 'Axis' parameter, which determines what the data is plotted against. Many of these variable also do not provide any meaningful information (e.g. the plot of time against time etc.). Yet others are, however, so packed with information as to be indecipherable. This latter category includes configuration, velocities, and gradients. The 3D plot which plots all the x,y,z components for every atom at every time-step into one graph is too dense to be useful. However, parts of this dataset contain very valuable information.

Therefore, the plotting options for MMTK NetCDF trajectories should be greatly expanded. This should include the following options and possibly more:

- plot the components of coordinates/velocities/gradients of particular atoms against time
- plot coordinates/velocities/gradients of particular atoms against time

All of this has arisen from the discussions in #34, but if anyone has further ideas for what would be good to be able to be plotted, please share!

eurydice76:
Resolves #58 

Hi @sanghamitra-mukhopadhyay  and @RastislavTuranyi , I implemented two ways for viewing the trajectory data (e.g. coords, velocities, gradients). One using the good old netcdf plotter and one through a dedicated plugin.

For the one using the current plotter, I added a new mode for plotting 3D data called 2D slice. For any 3D data, its plots a 2D image for a selected dim and a slice along that dimension. From there, we can use the already implemented slice mode which allows to have 1D slice plots which would correspond to what @sanghamitra-mukhopadhyay would like to have. For example, say that we want to plot the velocities against time for the first atom then we would select dim = 1 and slice = 0 (the velocities array has dimension (nFrames,nAtoms,3)) and from there do a slice plot.
This solution yet not perfect has some advantages: 1) it does not break the generality of our plotter 2) it uses already coded material. 

For the dedicated plugin, I implemented a brand new `ComponentPlugin` specifically dedicated to the viewing of Trajectory variables (the ones of dim `nframes x natoms x 3`). It is accessible under the Miscellaneous section under the Molecular Viewer plugin. As such, once a variable and a dimension selected, you can pick up an atom from the viewer and it corresponding variable will be plotted. It should fulfill directly the need of @sanghamitra-mukhopadhyay.


RastislavTuranyi:
Thank you! I have tried it out, and I would say the main issue is that the resultant graphs (for both implementations) do not have x- and y-axis labels, and either do not show units or the units shown are incorrect (always au on both axes). This makes the data interpretation more difficult, and ensures that the data has to be saved and then replotted using other means if the graph is to be used for any purpose.

I have also found the following bug in the 2D/3D Plotter:
image
It can be obtained by following these steps:

1. Open a trajectory file in 2D/3D Plotter
2. Open the result of an analysis in 2D/3D Plotter
3. Plot a slice of the trajectory
4. Click on the loaded analysis result in the 'key' part of the Data panel

eurydice76:
For the plugin, it is hard to define unit other than hardcoding them. Keep in mind that those variables are coming directly from MMTK trajectory conversion which does not implement the unit mechanism.

eurydice76:
I added support for unit in the trajectory viewer plugin. As stated before, these variables are automatically set during trajectory conversion. As such, a MMTK trajectory is not like a NetCDF output file coming from an analysis which implements the unit mechanism. So I had to hard code in some way the units for the configuration, the velocities and the gradients. For any other other vector field that may be associated to the trajectory the unit will be au (arbitrary unit).

eurydice76:
Hi @RastislavTuranyi @sanghamitra-mukhopadhyay can we proceed to merge ?

RastislavTuranyi:
I have confirmed that the units you have added work, but if we merge #72, perhaps it would be better to remove the hardcoded parts from the trajectory viewer plugin?

Further, the bug I have described still exists. I have discovered more accurate description of how to reproduced it, which I have edited into my previous post. If you think this bug is not an issue, I guess we could merge, but I think it would be best if it got fixed.

eurydice76:
I am sorry but I can't reproduce your bug because I do not understand the step 4. What does it mean ?

eurydice76:
OK, I could reproduce it finally. Lets fix that

eurydice76:
I could solve the bug you were describing in my last commit. I also fixed another bug introduced by the adding of the units to trajectory. I also took the decision to remove the forbidden_variables stuff. Indeed, while for a generic NetCDF viewer filtering the data from non-numeric values is completely normal, I do not want to introduce trajectory-specific treatment here. This will somehow break the generality of the code for a very little user comfort. Moreover, it does not hurt that much to have those variables showed especially since we can resize the variables list.

RastislavTuranyi (approved):
I am OK with reverting to showing all variables (except non-numeric). @sanghamitra-mukhopadhyay what about you?

I can also confirm that the bug I had previously observed has been removed. The things with units should be resolved in #72, so I would say that this can be merged.

sanghamitra-mukhopadhyay:
I have reviewed the implementation. The 2D plot of 1D data, i.e.,  MSD, VACF are great including the unit on axes. However, there is no legend when plotting in new window, so it is difficult to identify different plots, such as MSD_C and MSD_H when over plotting in current figure. Can it be implemented? 
Is it possible to remove a plot after doing a overplot ? That is after plotting MSD_C, MSD_H, MSD_total  etc. if one wants to remove MSD_C ? 
In case of loading a trajectory the 3D configuration is available for 2D slice plotting. However, it is not any meaning full plot. 
MDANSE crashes if I choose Scaler field plot, by mistake, to plot the trajectory configuration. 
It will be good if can be debugged. :)

sanghamitra-mukhopadhyay:
Just to emphasize that I checked the plot in MDANSE gui on Windows 10.

eurydice76:
First of all, do you really use the version corresponding to that branch ?
1) to get the legend, right-click on the plot, open the general settings, and check the Show legend checkbox
2) to remove a plot pick that plot with your left-mouse button and click on Suppr key, a dialog will ask to confirm that you want to delete that line
3) I can not reproduce your crash on linux/ubuntu20.04. @RastislavTuranyi are you on windows ? If so, can you reproduce the crash ? Moreover, I derived a dedicated plugin for viewing trajectory variables. To use it, 1] Open the trajectory with mdanse_gui 2) Open the molecular viewer plugin 3) Open the Miscellaneous/Trajectory viewer plugin

RastislavTuranyi:
Yes, the Scalar-Field plotter crashing on Windows is well known. It is detailed in #30.

eurydice76:
How/where can I get a windows installer so that I can reproduce the bug on my windows laptop ?

RastislavTuranyi:
An older version of each installer is on the [ISIS pages for MDANSE](https://www.isis.stfc.ac.uk/Pages/MDANSEproject.aspx). The latest builds are in one of the runs in the Actions tab of this repo, currently [this one](https://github.com/ISISNeutronMuon/MDANSE/actions/runs/2079303853).

sanghamitra-mukhopadhyay:
I have checked the latest CI/CD one "MDANSE-1.5.0-beta-ffc45bab-win-amd64". That has the same bug.

sanghamitra-mukhopadhyay:
@eurydice76  I like your new plugin for trajectory viewer :)
I think it is the good way to view trajectory and so we can remove that option from 2D/3D viewer 
 all together. In that case the bug in Scalar-plotting will be removed. We can develop the new plugin further to view other parameters available in the trajectory.  
I cannot remove selected plots either from the new plugin viewer or from the 2D/3D viewer on Windows. The Super key is not working the same way as you have mentioned. No new dialog is opened!
Also can we keep the legend option 'true' as default? Users can change it by doing the right click etc?

RastislavTuranyi:
Unfortunately, removing the Slicing plotter will not fix the Scalar-Field plotter bug; that bug has been present since the first time I build MDANSE on my laptop.

Concerning removing plots, this seems to be available only inside the 2D/3D Plotter. There, for me on a Windows laptop, using a 'Delete' key when a line is selected works. Lines can also be removed by right-clicking the plot and selecting the 'Lines settings', then selecting the line and clicking the delete button.
image
The Trajectory Viewer does not seem to have an option to remove plotted lines.

RastislavTuranyi:
Wow, I can confirm that the Scalar-Field plotter now works on Windows. Have you checked if the changes haven't caused any issues on Ubuntu?

eurydice76:
It seems to be OK on Linux. I also just committed the possibility to remove lines for the TrajectoryViewerPlugin. To remove a line select it by left-clicking on it, the press the Suppr key and the line will be removed upon confirmation. Please check on your side and if everything is OK, we are finally ready to merge.

eurydice76:
There is still something buggy with the legend. probably for a long time. Let's inspect this.

eurydice76:
OK, I fixed the buggy behavior with the legend. For info, when all lines were removed, there was still a small rectangle without any legend showing up on the plot. There was also a bug with setting or not the legend in the General settings dialog. @sanghamitra-mukhopadhyay the legend is set by default now. We should be ready for merge now.

PR #59, opened 2022-04-08
Fix PDF totals having zero values

RastislavTuranyi:
**Description of work.**

Fixes #56 

Adding RDF and TCF calculations in #40 introduced a bug where the total were calculated incorrectly. This was because the calculation of pdf_total for atom pairs was forgotten to be modified, and so it used the `self._outputData` object which was empty. This has been fixed by replacing the empty references with the appropriate, calculated arrays.

**To test:**

1. Run PDF analysis using the included trajectory waterbox_in_periodic_universe.nc
2. Observe that `pdf_total_OO` and `pdf_total` have non-zero values.

---

#### Reviewer ####

Please comment on the following ([full description](http://developer.mantidproject.org/ReviewingAPullRequest.html)):

##### Code Review #####

- Is the code of an acceptable quality?
- Does the code conform to the [coding standards](http://developer.mantidproject.org/Standards/)?
- Are the unit tests small and test the class in isolation?
- If there is GUI work does it follow the [GUI standards](http://developer.mantidproject.org/Standards/GUIStandards.html)?
- If there are changes in the release notes then do they describe the changes appropriately?
- Are the release notes saved in a separate file, using Issue or PR number for file name and in the correct location?

##### Functional Tests #####

- Do changes function as described? Add comments below that describe the tests performed?
- Do the changes handle unexpected situations, e.g. bad input?
- Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**.

PR #40, opened 2022-03-04
Implement rdf and tcf inside pdf

RastislavTuranyi:
**Description of work.**

Resolves #25 

Radial distribution function and total correlation function were calculated alongside pair distribution function in nMOLDYN, the predecessor to MDANSE, but are not implemented in MDANSE. This has now been changed. RDF and TCF are calculated as described in nMOLDYN Users' Guide, ie. the equations below are applied to calculate intermolecular, intramolecular, and total RDF and TCF for each pair of atoms as well as all atoms.

> image
> image
> image

**To test:**

1. Run a PDF analysis.
2. Open the result with 2D/3D Plotter.
3. Observe the calculated RDFs and TCFs

---

#### Reviewer ####

Please comment on the following ([full description](http://developer.mantidproject.org/ReviewingAPullRequest.html)):

##### Code Review #####

- Is the code of an acceptable quality?
- Does the code conform to the [coding standards](http://developer.mantidproject.org/Standards/)?
- Are the unit tests small and test the class in isolation?
- If there is GUI work does it follow the [GUI standards](http://developer.mantidproject.org/Standards/GUIStandards.html)?
- If there are changes in the release notes then do they describe the changes appropriately?
- Are the release notes saved in a separate file, using Issue or PR number for file name and in the correct location?

##### Functional Tests #####

- Do changes function as described? Add comments below that describe the tests performed?
- Do the changes handle unexpected situations, e.g. bad input?
- Has the relevant (user and developer) documentation been added/updated?

Does everything look good? Mark the review as **Approve**.

PR #20, opened 2021-10-05
Add continuous integration using GitHub actions

RastislavTuranyi:
**Description of work.**
Resolves #19 and resolves #17.

The CI/CD currently implemented in `.gitlab-ci.yml` was created for GitLab and is not compatible with GitHub actions. Therefore, the pipeline presented there has been rewritten using the syntax of GitHub actions to make the CI work in this repository. GitHub-hosted runners are used to run the CI/CD.

The Ubuntu pipelines have been moved from a docker container directly to the VM running the pipeline, mostly due to ease of implementation. Wx and VTK are obtained through anaconda since that is simpler and works on all versions of Ubuntu. A postinst script has been added to the .deb so that necessary environment variables are automatically set, thus not requiring the user to do anything other than install the package.

The MacOS pipelines similarly use anaconda wx and vtk. The deployment has been changed to properly generate a bundle using py2app, which solved #17. However, many dynamic libraries have to be relinked using `install_name_tool`. Still, this does not impact the speed of the CD, unlike the `hdiutil create` command necessary to generate a .dmg, which occasionally gets stuck and takes up to an hour to complete. Another issue that arose from using py2app is that, although the bundled python now has no signing issues, it does not have `PYTHONHOME` environment variable set. Rather than setting this for bash and risking disrupting other python installations, a script was created that will work exactly like a python executable.

The windows pipeline now installs wx and vtk from wheels included in the repo. The installer has been changed to only create a desktop shortcut when a user chooses to do so. Simultaneously, all link creations by the installer have been repaired and now work as intended. This pipeline has been significantly sped up by caching already installed Visual Studio 2008 (which takes upwards of 15 minutes to install) and then installing VC++ For Python 2.7 from a cached .msi. However, since this works only for MDANSE and not Scientific, the whole python with all packages is cached, and then MDANSE is reinstalled. This sped up the workflow further.  If this pipeline seems confusing, [here](https://github.com/ISISNeutronMuon/MDANSE/files/7528656/How.Windows.CI-CD.works.txt)  is a doc describing how it works.

As a result, the automatically generated installation files have been confirmed to work on the following platforms:
- Ubuntu 21.04
- Ubuntu 20.04
- Ubuntu 18.04
- MacOS 11 (with M1 chip)
- Windows 10

---

**To test:**

1. Download an executable for a platform of your choice from the [latest successful run](https://github.com/ISISNeutronMuon/MDANSE/actions) (at the very bottom of the summary page)
2. Install it
[...]

PR #66, opened 2022-04-20
Fix broken tests

RastislavTuranyi:
In recent development, on multiple occasions changes to the code have been made without changing the tests, resulting in a number of broken tests that currently exist on `develop`. Namely,

- #39  changed the behaviour of CASTEP converter. The testing behaviour was fixed on this branch.
- #40 changed the behaviour of PDF without reflecting the changes in the tests. However, since the resulting code has a bug, the tests for this will be corrected in #59 together with the bugfix.
- #50 changed the behaviour of a multitude of trajectory converters, breaking the tests for DFTB, Forcite, CASTEP, DMol, and Discover converters. This has been partially solved by ignoring the 'temperature' and 'kinetic_energy' variables present in the reference files, but DFTB and Forcite tests remain broken because the new code is capable of reading velocities while the reference files do not contain velocities. Can you please advise on this, @eurydice76? Is it better to hardcode inside `BuildJobTests.py` that in the dftb and forcite tests the velocities would not be compared, or is it better to generate new reference files in some manner?

Edit: Thanks to Miguel, new trajectories have been added for testing DFTB and Forcite converters. Reference files have been generated from these using current version of MDANSE for use in testing.

Edit 2: The tests broken due to the change of the `output_files` parameter in trajectory converters to `output_file` in #60 has also been fixed by handling the KeyError arising in the tests.


eurydice76:
First of all, please apologize me, I may be the culprit for some of the failing tests because I forgot to update the Tests. To come on your question, the best would be obviously to have DFTB and Forcite trajectories which contains velocities and gradients but to do so we have to generate ones. @gonzalezma could you generate new (short) trajectory for us in order that we can update our tests ? In the meantime we can hardcode in BuildJobTests in order to have a successful test suite.

eurydice76:
For me everything is OK, you can merge

eurydice76 (commented):
For me everything is OK. Just one question, why did you change the example trajectory for Forcite converter ? The old one used to work correctly so far.

RastislavTuranyi:
I didn't give it much thought. Since Miguel generated a new trajectory for Forcite, I made that the default trajectory. I trusted that the trajectory was created for a reason (not just so that I would then ignore it), but indeed it seems that both trajectories have velocities but not forces, so both are equally useable for testing purposes. Should I revert to the old trajectory?

eurydice76:
No keep the one you use. No problem. You can merge.

RastislavTuranyi:
This should have fixed all the tests (except PDF), so it should be ready to be merged.

eurydice76:
Great ! Thanks. By the way, would there be a way to speed up the build for windows (e.g. not installing Visual Studio each time). That would be great.

RastislavTuranyi:
I have already implemented a caching mechanism, as described in the attached file. It succeeded in making the windows pipeline faster than the MacOS one when I implemented it in #20.  So far, it seems that the failed CI runs have not triggered the mechanism, so I will observe whether it still works correctly now that multiple branches are involved, and make adjustments if necessary.
[How Windows CI-CD works.txt](https://github.com/ISISNeutronMuon/MDANSE/files/8589706/How.Windows.CI-CD.works.txt)

Issue #71, opened 2022-04-22
ENHANCEMENT: Add units to trajectory files


eurydice76:
The trajectory files, being NetCDFFile, should also have units in their inner variables (configuration, velocities ...).

PR #72, opened 2022-04-22
Write units in trajectory files

eurydice76:
See #71 for info.

### DIFF
--- a/Src/Framework/Jobs/Converter.py
+++ b/Src/Framework/Jobs/Converter.py
@@ -13,10 +13,54 @@
 #
 # **************************************************************************
 
+import abc
+
+from Scientific.IO.NetCDF import NetCDFFile
+
 from MDANSE.Framework.Jobs.IJob import IJob
 
 class Converter(IJob):
+
+    __metaclass__ = abc.ABCMeta
     
     category = ('Converters',)
     
     ancestor = ['empty_data']
+
+    @abc.abstractmethod
+    def run_step(self,index):
+        pass
+
+    def finalize(self):
+
+        if not hasattr(self,'_trajectory'):
+            return
+
+        f = NetCDFFile(self._trajectory.filename,'a')
+
+        if 'time' in f.variables:
+            f.variables['time'].units = 'ps'
+            f.variables['time'].axis = 'time'
+            f.variables['time'].name = 'time'
+
+        if 'box_size' in f.variables:
+            f.variables['box_size'].units = 'ps'
+            f.variables['box_size'].axis = 'time'
+            f.variables['box_size'].name = 'time'
+
+        if 'configuration' in f.variables:
+            f.variables['configuration'].units = 'nm'
+            f.variables['configuration'].axis = 'time'
+            f.variables['configuration'].name = 'coordinates'
+
+        if 'velocities' in f.variables:
+            f.variables['velocities'].units = 'nm/ps'
+            f.variables['velocities'].axis = 'time'
+            f.variables['velocities'].name = 'coordinates'
+
+        if 'gradients' in f.variables:
+            f.variables['gradients'].units = 'amu*nm/ps'
+            f.variables['gradients'].axis = 'time'
+            f.variables['gradients'].name = 'coordinates'
+
+        f.close()
--- a/Src/Framework/Jobs/Converter.py
+++ b/Src/Framework/Jobs/Converter.py
@@ -36,31 +36,33 @@ class Converter(IJob):
         if not hasattr(self,'_trajectory'):
             return
 
-        f = NetCDFFile(self._trajectory.filename,'a')
+        try:
+            f = NetCDFFile(self._trajectory.filename,'a')
+            if 'time' in f.variables:
+                f.variables['time'].units = 'ps'
+                f.variables['time'].axis = 'time'
+                f.variables['time'].name = 'time'
 
-        if 'time' in f.variables:
-            f.variables['time'].units = 'ps'
-            f.variables['time'].axis = 'time'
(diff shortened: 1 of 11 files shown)


Src/Framework/Jobs/Converter.py
```diff
@@ -13,10 +13,54 @@
...
+        pass
+
+    def finalize(self):
+
+        if not hasattr(self,'_trajectory'):
+            return
+
+        f = NetCDFFile(self._trajectory.filename,'a')
+
+        if 'time' in f.variables:
```
RastislavTuranyi, line 41: I think it would be a good idea to wrap these if statements inside a  try finally block to ensure that the file is always closed, even if an error occurs.

RastislavTuranyi (commented):
I have noticed that you didn't add the super() call to PDB converter; is there a reason for that?

Otherwise, everything seems to be working as intended.

Src/Framework/Jobs/Converter.py
```diff
@@ -13,10 +13,54 @@
...
+            f.variables['time'].name = 'time'
+
+        if 'box_size' in f.variables:
+            f.variables['box_size'].units = 'ps'
+            f.variables['box_size'].axis = 'time'
+            f.variables['box_size'].name = 'box_size'
+
+        if 'configuration' in f.variables:
+            f.variables['configuration'].units = 'nm'
+            f.variables['configuration'].axis = 'time'
```
RastislavTuranyi, line 53: Shouldn't the axis here consist of three variables, given that the data is 3D (i.e. similar to how F(Q, t) has axis of 'q, time')? As it stands, when I take a trajectory that was converted to MMTK NetCDF format using this branch and plot it using the 2D Slice plotter from #61, I get the following, which does have a variable on the x-axis, but does not have one on the y-axis.
image

eurydice76: Yeap but the time dimension (the first one) is the only dimension that has a sense. The other (number of atoms, and dimension) does not exist. They have no meaning.

Src/Framework/Jobs/Converter.py
eurydice76: Why not

eurydice76: On the other hand the if statements insure that the variables exists in the NetCDF file and so no error can occur from there.
If you prefer, according to the principle 'it is better to ask forgiveness than permission', I can replace the if by a try/except KeyError.

RastislavTuranyi: I personally don't mind the if statements. If you are sure there can never be an error inside the if statements, then I guess we can leave it be. It's just that leaving a file open tends to be a really bad idea, so I think it is good to take precautions just in case.

Src/Framework/Jobs/Converter.py
RastislavTuranyi: If the other two dimensions have no meaning, does this 2D slice have any meaning either?

eurydice76: That's a good philosophical question. For inspecting trajectories, I would say that the dedicated plugin I developed in branch 58 is much better. However, the 2D Slice, as such, is another way of inspecting 3D data and I would keep it.

Src/Framework/Jobs/Converter.py
eurydice76: you right. let's add some safety guards out there

eurydice76:
I just pushed a securized version of the code around the if statement. If you are OK, I can proceed to Merge.

Src/Framework/Jobs/Converter.py
```diff
@@ -13,10 +13,56 @@
...
+                f.variables['velocities'].axis = 'time'
+                f.variables['velocities'].name = 'velocities'
+
+            if 'gradients' in f.variables:
+                f.variables['gradients'].units = 'amu*nm/ps'
+                f.variables['gradients'].axis = 'time'
+                f.variables['gradients'].name = 'gradients'
+        except:
+            pass
+        else:
```
RastislavTuranyi, line 67: I think this does not do the job of closing the file regardless of whether there was or wasn't an error. The empty `except` block means that if an error is thrown, the file will be closed, since the else block is only executed when there were no errors. What I was referring to was this pattern:
```
f = NetCDFFile(self._trajectory.filename, 'a')
try:
    ...
finally:
    f.close()
```

eurydice76: Even better would be:

```
        try:
            f = NetCDFFile(self._trajectory.filename,'a')
        except:
            return
        
        try:
            if 'time' in f.variables:
                f.variables['time'].units = 'ps'
                f.variables['time'].axis = 'time'
                f.variables['time'].name = 'time'

            if 'box_size' in f.variables:
                f.variables['box_size'].units = 'ps'
                f.variables['box_size'].axis = 'time'
                f.variables['box_size'].name = 'box_size'

            if 'configuration' in f.variables:
                f.variables['configuration'].units = 'nm'
                f.variables['configuration'].axis = 'time'
                f.variables['configuration'].name = 'configuration'

            if 'velocities' in f.variables:
                f.variables['velocities'].units = 'nm/ps'
                f.variables['velocities'].axis = 'time'
                f.variables['velocities'].name = 'velocities'

            if 'gradients' in f.variables:
                f.variables['gradients'].units = 'amu*nm/ps'
                f.variables['gradients'].axis = 'time'
                f.variables['gradients'].name = 'gradients'
        finally:
            f.close()

```

eurydice76:
My bad I forgot to commit my last change. I will do in another PR. Sorry.

Issue #75, opened 2022-04-29
ENHANCEMENT: Create a way to transform S(Q, ω) into S(Q, E)


RastislavTuranyi:
In my discussions at NMSUM, one of the issues that was expressed to me is that often the scattering function S is of greater interest as a function of energy rather than of frequency. This is not a major problem, but it does add a layer of difficulty to data interpretation, and so removing this would be a nice quality-of-life change. 

There are two ways in which I think this could be achieved:

- Change the DSF analyses to create new outputs, or
- Add a feature to the 2D/3D Plotter that allows for not only unit conversion, but also a limited physical quantity conversion.

gonzalezma:
There is already a unit converter in the 2D/3D plotter. Personally, I don't find it too useful, but I don't understand what you mean by physical quantity conversion. I imagine that you mean e.g. rad/ps vs meV vs cm-1 vs THz, although I would call this simply unit conversion. I thought that the unit converter was able to do this, but I could not manage to make it work myself (I get an error message the the axis unit is inconsistent with the current unit).

The unit converter affects only to the plot and not to the netCDF file, so this is not very useful if one wants to compare the MD result to experiment, which I assume is done outside MDANSE. The alternative is to add an option to each relevant analysis allowing to choose the output units, which is something that was already possible in some nMoldyn versions. Personally, I prefer the new way, but I admit that this is debatable. My reasons are:

1. This removes one option from the analysis menu. And simplifies the analysis, which otherwise needs to include a unit conversion at the end.

2. It is clearer that after converting a trajectory, all internal handling and analysis outputs are in MDANSE units (basically nm, 1/nm, ps, rad/ps). Then converting to any other unit is straightforward and should not be a problem for any user.

So I prefer to keep the code and input menu simpler at the cost of making life perhaps a little bit more uncomfortable to the users. I would even add that there is some educational value in understanding the relationship between the different units and how to pass from one to another, and that anyone trying to compare MD and experiment should be able to handle this and not use MDANSE blindly.

But if we decide to make user's life easier, I would go for the option of selecting the units in the input of the analysis. We should decide then to which analysis and quantities to apply it (only for the frequency/energy or also for time, length, wavevector, etc.) and how generous we should be with the units (e.g. only angstrom and nm for length, or also cm, mm, m, etc.).

eurydice76:
[...]

PR #78, opened 2022-05-05
Add advanced units manipulation and definition options

eurydice76:
See #75 for more info. The goal of this PR is to improve the units conversion. So far, we used an external dependency called magnitude which does some part of the job. However, this module was missing a key component for us: the ability to convert between equivalent units (e.g. THz/eV, THz/K ...). To do so, based on this module, I rewrote a new Unit module which is now used by the plotter. To change between units, just do like before, open plot settings and change X/Y unit at the difference that you can now switch seamlessly e.g. from THz to eV. I also developed a units editor which allows to manipulate units. This editor can be launched through a standalone app (mdanse_units_editor) or through the manse_gui.


RastislavTuranyi:
Firstly, thank you for this work; it adds a lot of options!

Now, some observations/suggestions based on using the new features:

- The `Add unit` button in the Unit editor allows for silent overwriting of default units. I think this should be either completely forbidden, but if not, at least a warning asking for confirmation should be shown.
- Similarly, when overwriting a user-defined unit through the `Add unit` window, a warning should be shown (I don't think this should be forbidden).
- If a non-numeric value is inputted into one of the 'kg' etc. fields, the value is always changed to -100 regardless of the inputted character. I don't know if this is intended or not.
- If a non-numeric value is inputted into the `Factor` field of the Units editor (not Add unit) an error is thrown, which I like. However, afterwards the `Factor` field cannot be changed, even when I close and reopen the units editor. I have to close and reopen MDANSE for the bug to go away. Everything works OK in the Add unit window.
- When I tried to convert rad/ps into eV , I got an error.
- I was able to convert T into Hz, T into eV, and eV into Hz.

eurydice76:
Thanks Rastislav for the careful review. So let's answer to your remark/question one after the other:
1. I push a new commit where the user is informed that he/she is overwriting existing unit
2. I do not see the difference with the point 1. Maybe it is OK now
3. I am very surprised that you can input something different than number in the kg,m,s ... widgets because they are coded as `SpinCtrl` widgets which accept only integers. Maybe I misunderstand something in your question.
4. I can not reproduce this bug on my machine
5. I just pushed a new commit with the rad/s convertible to eV. It should be OK now.
6. OK
7. I could make it work by doing:
    UNITS_MANAGER.add_unit('KK',2.0,*[0,0,0,1,0,0,0,0,0])
    print(measure(2.0, 'KK').toval('K'))
    Maybe you forgot the '*' before the dimension list.
8. OK

Could you please check that the changes I did solve some or all your problem ?

RastislavTuranyi:
The difference between 1 and 2 is that some units, such as Angstrom and Kelvin, come pre-defined in MDANSE because they are fundamental/important, either physically or for MDANSE, while other units, such as the KK and HZ testing units in my previous post, are defined later by a user because they want to use a unit that is not pre-defined. I was trying to point out that it might be worth it to prevent users from editing the former category since they are the correct units corresponding to essentially immutable physical quantities, and so there should never be a reason to change them, yet someone may accidentally change them/forget they changed them, and then be confused that the default units are not working as expected.

1. I can confirm; there is now a warning.
2. I can confirm; there is now a warning.
3. It is possible to type in any character into the field:
image
but when I click outside the box, the value automatically changes to -100:
image
If you think this behaviour is acceptable, I don't mind leaving it at that.
4. Hmm, might be another issue with Windows wxpython.
5. Can confirm.
6.
7. I only used the GUI; this is how my KK looks:
image
And then I tried to change the units of a temperature graph in the 2D/3D Plotter through the Axes menu.

eurydice76:
@RastislavTuranyi I just pushed some updates in the `UnitsEditor` dialog which address some of the problems you mentioned. Now, in the main dialog the user can not change the factor and the dimension of a selected unit. This is normal because, as you mentioned, once defined a unit has no reason to be changed. However, a unit still can be removed by clicking on delete key. However, if you do so you will be asked whether you really want to delete this unit. When  you crate a new unit, I added more control on the unit to be added. Indeed, you are prevented from adding a unit that already exists. The unit must also have a name and a valid factor. With that, the unit database should be safe.  Please send me your file with the temperature in order that I can reproduce the bug that you mentioned about the KK unit.

eurydice76:
I just pushed a commit where the problem with KK unit should be solved. If everything is OK for you we are ready to merge.

RastislavTuranyi:
Thank you! I can confirm the changes you have described, though I have noticed that I cannot convert from kJ_per_mole to cal_per_mole, probably because kJ_per_mole is not pre-defined.

eurydice76:
Yeap. It was not defied. I just did it in my last commit.

RastislavTuranyi:
I have also noticed that I can't convert the default units of Q, inv_nm, into anything else. Should we add a conversion of inv_nm into 1/nm? Other than this, I think all other units should be able to be converted into some other units.

eurydice76:
Yeap that's because this unit is not defined anymore as a unit. Instead for each job based on Q vectors, the unit for Q vector should be "1 / nm" instead of inv_nm. I will modify the corresponding jobs. For working with old analysis file I will also add this unit in the Units.yml file.

eurydice76:
There is still something buggy with the conversion from nm to ang. I will look at this.

Src/Framework/Units.py
```diff
@@ -0,0 +1,816 @@
...
+        """
+
+        return self._equivalent
+
+    @equivalent.setter
+    def equivalent(self, equivalent):
+        """Setter for _equivalent attribute.
+        """
+
+        self._equivalent = equivalent
```
RastislavTuranyi, line 505: Is there any point to having a `@property` wrapper and a `setter` if the setter simply acts as a wrapper, without calling any function in between the assignment? Why not just have a `self.equivalent` attribute without the `@property` instead of `self._equivalent` attribute and wrappers? 

This also goes for the other wrappers below, `format` and `units`. On the other hand, `dimension` and `factor` are ok since they don't have setters, so it's obvious that they have to be coded the way they are.

Tests/UnitTests/TestUnits.py
```diff
@@ -0,0 +1,196 @@
...
+class TestUnits(unittest.TestCase):
+    '''
+    Unittest for the geometry-related functions
+    '''
+        
+    def test_basic_units(self):
+
+        m = measure(1.0,'kg')
+        self.assertAlmostEqual(m.toval('g'),1.0e+03,10)
+        self.assertAlmostEqual(m.toval('mg'),1.0e+06,10)
```
RastislavTuranyi, line 61: This is not good unit test design, since if the first assertion fails, then all the following ones within this massive method will be skipped. Ideally, each assertion statement should be in a separate method, though in this case I think having these two assertions related to kg within one method would be acceptable. Nevertheless, each `measure()` call afterwards should be in a separate method.

RastislavTuranyi (approved):
I took a look at the code it looks good, though I found a couple minor things, the largest of which is with unit tests.

eurydice76:
It should be OK now. The units for scattering related analysis are now in 1/nm instead of inv_nm. I tried with an old file and the conversion from inv_nm to 1/pm is OK for example.

Tests/UnitTests/TestUnits.py
eurydice76: sorry for not being a good unit test designer !

eurydice76: I am in partial agreement with you. Indeed, that unit test is massive but its goal is to test the 9 basic units. For me, breaking into 9 unit tests is not that good design too.  Anyway, I will take your remark into consideration.

eurydice76: https://stackoverflow.com/questions/762512/is-it-bad-practice-to-have-more-than-one-assertion-in-a-unit-test

eurydice76: I think that the truth is in-between. For instance if I test the prefixes. I think it is reasonable to build one unit test for all prefix and not a unit test per prefix.

RastislavTuranyi: Firstly, apologies if I worded that initial comment incorrectly; I was trying to point out a place where the code can be improved and nothing else. 

If testing all prefixes is the goal, [parametrisation](https://stackoverflow.com/questions/32899/how-do-you-generate-dynamic-parameterized-unit-tests-in-python) is the way to go, I would say. But, as I have belatedly noticed, this method tests the units and prefixes, so now that I think about it, it would be good to separate these aspects.

Further, I used this method as an example, but in the initial comment, I was commenting on all the test cases, especially the later ones. For example, `test_product_units()` uses three assertions, where each assertion relies on the previous one.

Would it be ok if I rewrote these tests?

RastislavTuranyi:
I can confirm that newly generated calculation are in 1/nm instead of inv_nm, but I am unable to convert units on older files from inv_nm into 1/nm using the Plotter.

eurydice76:
I added unit test for prefixes. As mentioned in the reply of your comment, I decided to build a single test for all prefixes. I understand your concern about avoiding to have several assertions in a single test, but I do think that having a unit test pr prefix is not good design too. Unit test should be grouped by theme when possible. In the test_basic_units, I grouped for example the concept of testing the basic units and the concept of testing prefixes which was indeed not good design. So that is the reason why, ccording to your recommendation, I splitted that test into two unit tests: one for the basic unit and one for the prefixes.

eurydice76:
About the fact that you still can not convert inv_nm to 1/nnm, this is because you may already have a copy of the unis database in your app data path. Remove it and it should be OK. Keep me in touch.

Src/Framework/Units.py
```diff
@@ -0,0 +1,816 @@
...
+             'Y': 1e24,   # yotta
+        }
+
+class UnitError(Exception):
+    pass
+
+def get_trailing_digits(s):
+
+    for i in range(len(s)):
+        if s[i:] in string.digits:
```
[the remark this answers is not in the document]
eurydice76: you right

Src/Framework/Units.py
eurydice76: I set properties for `_equivalent` because I always declared my variable as _hidden_ when designing a class. In the case of `_equivalent`, I agree that it is a bit overdesign but it allows to be consistent with other variables such as `dimension`.

RastislavTuranyi:
You were right, I deleted the yaml from AppData, and the inv_nm to 1/nm conversion now works.

sanghamitra-mukhopadhyay:
@eurydice76  Can you please resolve this conflict.

eurydice76:
Hi @sanghamitra-mukhopadhyay , the conflict is fixed. You can go on with reviewing the code.

sanghamitra-mukhopadhyay (approved):

PR #81, opened 2022-05-05
Use netCDF4 as the NetCDF dependency instead of Scientific.IO/NetCDF

eurydice76:
In the process of getting rid of ScientificPython and MMTK, I use in this PR a dependency to NetCDF4 library which is standard (can you please install it in the build machine). I had to change the unit test and now some are breaking again. Could you please investigate on this Rastislav as you were on that topic a few days ago ? Many thanks in advances.

### DIFF
--- a/Src/GUI/Plugins/PlotterPlugin.py
+++ b/Src/GUI/Plugins/PlotterPlugin.py
@@ -17,13 +17,13 @@ import collections
 
 import numpy
 
+import netCDF4
+
 from matplotlib.figure import Figure
 from matplotlib.backends.backend_wxagg import FigureCanvasWxAgg
 
 import wx
 import wx.aui as wxaui
-
-from Scientific.IO.NetCDF import NetCDFFile
 
 from MDANSE import REGISTRY
 from MDANSE.Core.Error import Error
@@ -472,11 +472,11 @@ class PlotterFrame(wx.Frame):
             basename = baselist[i]    
             filename = filelist[i]
             
-            f = NetCDFFile(filename,"r")
+            f = netCDF4.Dataset(filename,"r")
             _vars = f.variables
             data = collections.OrderedDict()
             for k in _vars:
-                dtype = _vars[k].getValue().dtype
+                dtype = _vars[k][:].dtype
                 if not numpy.issubdtype(dtype,numpy.number):
                     continue
                 data[k]={}
@@ -487,7 +487,7 @@ class PlotterFrame(wx.Frame):
                         data[k]['axis'] = []
                 else:
                     data[k]['axis'] = []
-                data[k]['data'] = _vars[k].getValue()
+                data[k]['data'] = _vars[k][:]
                 data[k]['units'] = getattr(_vars[k],"units","au")
             
             unique_name = self.unique(basename, self.plugin._dataDict)
--- a/Src/GUI/Plugins/PlotterPlugin.py
+++ b/Src/GUI/Plugins/PlotterPlugin.py
@@ -468,32 +468,29 @@ class PlotterFrame(wx.Frame):
         baselist = dialog.GetFilenames()
         filelist = dialog.GetPaths()
         
-        for i in range(len(filelist)):
-            basename = baselist[i]    
-            filename = filelist[i]
-            
-            f = netCDF4.Dataset(filename,"r")
-            _vars = f.variables
-            data = collections.OrderedDict()
-            for k in _vars:
-                dtype = _vars[k][:].dtype
-                if not numpy.issubdtype(dtype,numpy.number):
-                    continue
-                data[k]={}
-                if hasattr(_vars[k], 'axis'):
-                    if _vars[k].axis:
-                        data[k]['axis'] =  _vars[k].axis.split('|')
+        for basename, filename in zip(baselist, filelist):
+            with netCDF4.Dataset(filename, "r") as f:
+                _vars = f.variables
+                data = collections.OrderedDict()
+                for k in _vars:
+                    dtype = _vars[k][:].dtype
+                    if not numpy.issubdtype(dtype, numpy.number):
+                        continue
+                    data[k]={}
+                    if hasattr(_vars[k], 'axis'):
+                        if _vars[k].axis:
+                            data[k]['axis'] = _vars[k].axis.split('|')
+                        else:
+                            data[k]['axis'] = []
                     else:
                         data[k]['axis'] = []
-                else:
-                    data[k]['axis'] = []
-                data[k]['data'] = _vars[k][:]
-                data[k]['units'] = getattr(_vars[k],"units","au")
+                    data[k]['data'] = _vars[k][:]
+                    data[k]['units'] = getattr(_vars[k], "units", "au")
             
-            unique_name = self.unique(basename, self.plugin._dataDict)
+                unique_name = self.unique(basename, self.plugin._dataDict)
             
-            self.plugin._dataDict[unique_name]={'data':data,'path':filename,'basename':basename}
-            self.plugin._dataPanel.show_dataset(-1)
+                self.plugin._dataDict[unique_name]={'data': data, 'path': filename, 'basename': basename}
+                self.plugin._dataPanel.show_dataset(-1)
     
     def unique(self, key, dic):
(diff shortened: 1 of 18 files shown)


eurydice76:
I feel sorry for having asked you to solve the failing build. I did not think it would be such a fight. Could you tell me what is going on ? Is that the adding of NetCDF4 library which makes the things so hard ?

RastislavTuranyi:
That's just how things are with the pipeline, since I have to push every time I want to test the changes I made. I have solved the issue with Ubuntu CI/CD, where it turned out that I had to alter the `PATH` and `LD_LIBRARY_PATH` environmental variables before running the tests so that the location where apt installs netcdf and libnetcdf is searched (for some reason it was not necessary before).

The only thing that remains is this bizarre error generated by `BuildJobTests.py` and `AllTests.py` respectively on Windows:
```
  File "FunctionalTests\Jobs\BuildJobTests.py", line 258, in <module>
    job_file_generator = JobFileGenerator(job, job_id=job_id)
  File "FunctionalTests\Jobs\BuildJobTests.py", line 34, in __init__
    self.job.get_default_parameters()
  File "C:\hostedtoolcache\windows\Python\2.7.18\x64\lib\site-packages\mdanse-1.5.0-py2.7-win-amd64.egg\MDANSE\Framework\Configurable.py", line 252, in get_default_parameters
    cfg=REGISTRY["configurator"][typ](name, **kwds)        
KeyError: 'netcdf_input_file'
  File "FunctionalTests\Jobs\AllTests.py", line 37, in <module>
    run_test()
  File "FunctionalTests\Jobs\AllTests.py", line 32, in run_test
    if not unittest.TextTestRunner(verbosity=2).run(suite()).wasSuccessful():
  File "FunctionalTests\Jobs\AllTests.py", line 24, in suite
    modules = [__import__(os.path.splitext(f)[0],globals(),locals(),[],-1) for f in files]
  File "D:\a\MDANSE\MDANSE\Tests\FunctionalTests\Jobs\test_gromacs_trr.py", line 6, in <module>
    import netCDF4
  File "C:\hostedtoolcache\windows\Python\2.7.18\x64\lib\site-packages\netCDF4\__init__.py", line 3, in <module>
    from ._netCDF4 import *
ImportError: DLL load failed: The specified module could not be found.
```
What's puzzling about it is that I have no idea where the key `netcdf_input_file` comes from, and that `_netCDF4` file clearly exists:
```
Directory of C:\hostedtoolcache\windows\Python\2.7.18\x64\Lib\site-packages\netCDF4

05/09/2022  09:17 AM    <DIR>          .
05/09/2022  09:17 AM    <DIR>          ..
12/[21](https://github.com/ISISNeutronMuon/MDANSE/runs/6349010748?check_suite_focus=true#step:8:21)/2018  03:51 PM            35,547 utils.py
12/21/2018  03:51 PM            25,997 utils.pyc
12/21/2018  03:51 PM           973,8[24](https://github.com/ISISNeutronMuon/MDANSE/runs/6349010748?check_suite_focus=true#step:8:24) _netCDF4.pyd
12/21/2018  03:51 PM               686 __init__.py
12/21/2018  03:51 PM               849 __init__.pyc
               5 File(s)      1,036,903 bytes
               2 Dir(s)  94,164,[25](https://github.com/ISISNeutronMuon/MDANSE/runs/6349010748?check_suite_focus=true#step:8:25)0,624 bytes free
```

RastislavTuranyi:
Ok, update;

- It seems the issue with Ubuntu is actually that the netCDF4 library can't really use the netcdf apt package; instead it needs the libraries it itself installs.
- The KeyError happens when it tries to create the test for `StructureFactorFromScatteringFunction`; I have no idea at the moment why it can't find the `NetCDFInputFileConfigurator`

eurydice76:
It can't find the `NetCDFInputFileConfigurator` because that means that there is an error inside that module which prevents the `import` from working well.

eurydice76:
What is strange is that on my machine, nothing wrong happens when loading that configurator.

eurydice76:
However, I still have failing tests for CASTEP for example. Does this test fails also on your machine ?

RastislavTuranyi:
Yes, those tests are failing for me as well, and they are failing on the working MacOS pipeline too. I think it may be something we fixed after this branch was created.

And, yes, it is very weird that the configurator is not initialised, as is the fact that the Ubuntu pipeline is now failing because MMTK can't find `_netcdf`, even though during its installation it says `NetCDF header found --> /opt/hostedtoolcache/Python/2.7.18/x64/include/python2.7/Scientific/netcdf.h`.

RastislavTuranyi:
I see what's going on; the `NetCDFInputFileConfigurator` is failing because it itself import netCDF4, which the pipeline has an issue with; therefore, what remains is to find out why it can't import the `_netCDF4` module which clearly exists.

eurydice76:
Good, you're on the way.

RastislavTuranyi:
Almost everything works now:

- the problem on Ubuntu was that `Scientific.IO.NetCDF` was using the `netcdf.so` library from the netCDF4 installation instead of the apt installation, so all the relevant libraries had to be copied.
- on Windows, netCDF4 couldn't find the netcdf DLLs, so they had to be copied to the netCDF4 directory in site packages.
- when you edited `BuildJobTests.py`, you accidentally overwrote some of the changes implemented in #66; I have essentially merged together the changes from both branches + some extra (netCDF4 implements a context manager!)

The only issue is that the Ubuntu 20 pipeline has a permission problem in one of the tests.

eurydice76:
Well done ! Sorry for the mistake.

eurydice76:
This is very strange that you do not have the right to write on /tmp on the ubuntu 20.04 server. This is exactly the aim of this directory: allowing anybody to write on it. You should set the permission on this directory by  doing a `chmod -R 777 /tmp`

eurydice76:
Can you login on the ubuntu 20.04 machine and send me the return value of `ls -lrth /` ?

RastislavTuranyi:
I have made changes to `BuildJobTests.py` so that the output file path of the MDANSE jobs is directly `../../../Data/Jobs_reference_data` rather than outputting it to temp and then copying it from temp to the reference data folder. This has solved the permission issue.

Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -472,11 +472,11 @@ def on_load_data(self, event=None):
             basename = baselist[i]    
             filename = filelist[i]
             
-            f = NetCDFFile(filename,"r")
+            f = netCDF4.Dataset(filename,"r")
```
RastislavTuranyi, line 475: When/where does this file get closed?

eurydice76: This file should be closed at the end of the method.

Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -487,7 +487,7 @@ def on_load_data(self, event=None):
                         data[k]['axis'] = []
                 else:
                     data[k]['axis'] = []
-                data[k]['data'] = _vars[k].getValue()
+                data[k]['data'] = _vars[k][:]
                 data[k]['units'] = getattr(_vars[k],"units","au")
             
             unique_name = self.unique(basename, self.plugin._dataDict)
```
RastislavTuranyi, line 493: Also, why is `basename` defined inside the for loop, but referenced only outside of it?

eurydice76: No, basename is defined at the `for i in range(len(filelist))` level and use at the end of that block. I do not see any problem here.

Src/GUI/Plugins/PlotterPlugin.py
eurydice76: Yeap exactly

eurydice76:
Hi @sanghamitra-mukhopadhyay, could you please review asap this important PR so that we can develop based on this ?

sanghamitra-mukhopadhyay (approved):

sanghamitra-mukhopadhyay:
I am merging this pull request. @RastislavTuranyi can you please update the compilation documentation to reflect this change including the current draft version of the Users Guide ? Thanks.

Issue #80, opened 2022-05-05
BUG: Q vector definitions not working


RastislavTuranyi:
When trying to create a new Q vector definition by clicking the 'new definition' button inside an analysis, the following error is thrown:
```
C:\Users\dni83241\Documents\Virtualenvs\quickfix\lib\site-packages\MDANSE\GUI\Widgets\UserDefinitionWidget.py -- 97 -- on_new_definition -- dlg = UserDefinitionDialog(None,self._trajectory,self._type)
C:\Users\dni83241\Documents\Virtualenvs\quickfix\lib\site-packages\MDANSE\GUI\Widgets\UserDefinitionWidget.py -- 45 -- __init__ -- self._plugin.set_trajectory(trajectory)
                                 
AttributeError: 'ElasticIncoherentStructureFactorPlugin' object has no attribute 'set_trajectory'
```

Further, already existing definitions cannot be accessed; the drop-down menu is simply empty.

eurydice76:
Waw, this is a nasty bug. It is related with the mechanism used to auto-generates the job dialogs. I will have a look at this asap.

PR #84, opened 2022-05-09
Fixed regression in ConfigurationPanel

eurydice76:
I think that I fixed the bug describes in #80. It was a regression introduced by the recent changes we did concerning the output files widget. Given the fact that it was kind of kernel-related bug, could you please carefully check that everything is OK from the GUI point of view. On my machine, everything seems OK but who knows ...


RastislavTuranyi (approved):
Thank you! Not only does this solve the initial bug, it also solves the same bug which occurs when opening the PartialChargesPlugin. Everything I have tried doing works for me too.

eurydice76:
Great ! So let's merge.

sanghamitra-mukhopadhyay (approved):

Issue #83, opened 2022-05-09
BUG: Partial charges user definition dialog crashes


eurydice76:
When closing the partial charges user definition dialog, this one crashes unexpectedly with a C++ assertion.

PR #85, opened 2022-05-09
Fixed bug when closing partial charges plugin

eurydice76:
See #83 for info. The bug was in the `OnSize` method which sets the width of the dialog to 0 when it was closed.


RastislavTuranyi (approved):
I cannot reproduce the bug described in #83, but these changes seem reasonable and have not caused any issues. Everything seems to work as intended.

sanghamitra-mukhopadhyay (approved):

Issue #68, opened 2022-04-21
ENHANCEMENT: Add abstraction for plotter data


eurydice76:
Quite soon, we will have to deal with hdf5 file in MDANSE. However, it would be good if the plotter could handle both NetCDF and HDF files transparently. To do so we have to add a layer of abstraction around the data passed to the plotter.

eurydice76:
Hi @RastislavTuranyi , I made some advances in the plotter data abstraction. It is nearly finished. As the reminder, the goal is to provide an abstraction layer to give to the plotter the possibility to accept both NetCDF and HDF (and why not other formats in the future) file in a nice way. Everything seems OK and we are in position to have a plotter for any NetCDF file (even ones not coming from MDANSE) now but at one cost. Indeed, thanks to the new NetCDF4 dependency, we can now have nested variables in the NetCDF file hierarchy. This was not possible with the implementation in Scientific.IO.NetCDFFile. This is a cool feature. So that means that we have to recursively find all the variables present in the hierarchy starting from '/'. So every variable found will have a name like '/msd', '/time' ... This is how it will be displayed in the plotter. I am OK with this because this name reflects where the variables is located in the file hierarchy and as such it provides some information. But on the other hand I can understand that you or MDANSE user may find this less lice than before where the variable was displayed without '/' at the beginning. Could you please give a try and tell me what you think of this ?

RastislavTuranyi:
I think that having the / show in the variable box is not a huge issue, but when it appears in a plot, the user will have to manually edit it out every time they want to use the plot. Further, I think it has a potential to cause confusion as to what is going on in the plot, since not everyone would be able to associate the / with file structure (I wouldn't realise if you hadn't told me), as opposed to e.g. division or something like that. Overall, I think what a user wants from an MDANSE plot, is a standard plot that can be used in other context, and a / before a variable name is not part of a standard plot.

Regardless, isn't it easy to remove the / from the plot? E.g. instead of `self.Xaxis_label = self.dataproxy[varname]['axis'][0]` do `self.Xaxis_label = self.dataproxy[varname]['axis'][0].replace('/', '')`.

eurydice76:
[...]

PR #86, opened 2022-05-13
Add abstraction for plotter data

eurydice76:
See #68 for info.


Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -469,28 +472,40 @@ def on_load_data(self, event=None):
...
+                else:
+                    data[vname]['axis'] = []
+                data[vname]['path'] = vpath
+                data[vname]['data'] = arr
+                data[vname]['units'] = getattr(variable, "units", "au")
+
+            unique_name = self.unique(basename, self.plugin._dataDict)
+        
+            self.plugin._dataDict[unique_name]={'data': data, 'path': filename, 'basename': basename}
+            self.plugin._dataPanel.show_dataset(-1)
```
RastislavTuranyi, line 508: I can't see the netcdf file being closed; am I wrong? Ideally, I would love if we could implement a context manager to handle the file, to ensure that it is always closed.

Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -128,9 +130,10 @@ def build_panel(self):
             self.datalist = wx.ListCtrl(self.setup, wx.ID_ANY,style = wx.LC_REPORT|wx.LC_SINGLE_SEL)
 
         self.datalist.InsertColumn(0, 'Variable', width=100)
-        self.datalist.InsertColumn(1, 'Axis', width=50)
-        self.datalist.InsertColumn(2, 'Dimension')
-        self.datalist.InsertColumn(3, 'Size')
+        self.datalist.InsertColumn(1, 'Path', width=100)
```
RastislavTuranyi, line 133: I would like to propose a different order for these variables; I think that keeping the order as it has been and just adding the path at the end (last column) is better, given that I don't know  what knowing the path could be useful for. What do you think?

RastislavTuranyi (commented):
Hi, thank you for your work! I think the new naming works better, though I have a couple extra comments. The main issue is the file not being closed. Theoretically,  a context manager would be best, but I'm not sure if it is practical in this particular case;, so I'll take a look at this, if you don't mind.

Src/GUI/Plugins/PlotterPlugin.py
```diff
@@ -469,28 +472,40 @@ def on_load_data(self, event=None):
...
+                    break
+            else:
+                continue
+
+            data = collections.OrderedDict()
+            for vname, vinfo in f.variables.items():
+                vpath, variable = vinfo
+                arr = variable.get_array()
+
+                if not numpy.issubdtype(arr.dtype, numpy.number):
```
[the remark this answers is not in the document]
eurydice76: yeap you right the extra `numpy.issubdtype(arr.dtype, numpy.number)` is superfluous as the non numeric variables are already filtered out.

Src/GUI/Plugins/PlotterPlugin.py
eurydice76: yeap a context manager would the best thing to do

eurydice76:
thanks for the quick review Rastislav. As you will investigate for the context manager, I let you also proceed to the small changes you recommended in your review.

Src/GUI/Plugins/PlotterData.py
```diff
@@ -0,0 +1,129 @@
...
+                var_key = '{:s}_{:d}'.format(var_key, comp)
+                comp += 1
+
+            var_dict[var_key] = (path, NetCDFPlotterVariable(var))
+
+        for _, sub_group in group.groups.items():
+            NetCDFPlotterData.find_numeric_variables(var_dict, sub_group)
+
+
+PLOTTER_DATA_TYPES = {'.nc': NetCDFPlotterData, '.cdf': NetCDFPlotterData, '.netcdf': NetCDFPlotterData}
```
RastislavTuranyi, line 129: I changed this dict to make accessing the class easier/more readable at the cost of repetition in this definition. I think this should be ok as long as no class has too many extensions, but feel free to change it if you prefer a different solution.

RastislavTuranyi:
Also, I keep getting this error:
```
Exception RuntimeError: RuntimeError(u'NetCDF: Not a valid ID',) in <bound method NetCDFPlotterData.__del__ of <MDANSE.GUI.Plugins.PlotterData.NetCDFPlotterData object at 0x000000003B9C5DC8>> ignored
```
which causes no issues (since it's ignored), but I have not idea why this is happening.

Src/GUI/Plugins/PlotterData.py
eurydice76: no that's OK for me

eurydice76:
Hi @RastislavTuranyi , I removed the __del__ method of _IPlotterData class and it seems to solve the error/warning that you get. I think that we are ready to merge.

RastislavTuranyi (approved):
I can certainly confirm that the warning is gone.

eurydice76:
Hi @sanghamitra-mukhopadhyay , could you please review and proceed to merge if you are OK ?

sanghamitra-mukhopadhyay (approved):

Issue #8, opened 2021-07-05
create and test compilation/installation instructions


franzlang:
create and test instructions to compile the software and run the existing tests. a decent starting point might be these instructions sent by Remi Perenon, one of the developers at the ILL:

1. Install dependencies first. Cython pyro and numpy are required for non-GUI use. Wx and vtk must be added if you want to use GUI.
2. Then, you must install modified versions of Scientific-python (https://code.ill.fr/scientific-software/scientific-python) and MMTK (https://code.ill.fr/scientific-software/mmtk). Installation should be fine with git clone - python setup.py build - sudo python setup.py install. Be careful, no netcdf error should appear in your MMTK installation log.
3. Install mdanse from source (https://code.ill.fr/scientific-software/mdanse) as previously
4. Unit tests and dependencies tests can be launched from <mdanse>/Tests/UnitTests or <mdanse>/Tests/DependenciesTests, 

and/or these specific instructions which have been used to install MDANSE on the ISIS Data Analysis as a Service platform (https://isis.analysis.stfc.ac.uk/), which runs CentOS 7 machines:
```
#install python2-pip, wxPython and netcdf headers and dependencies
yum install -y python2-pip wxPython wxBASE wxGTK wxGTK-gl wxGTK-media python2-matplotlib-wx netcdf-devel
#upgrade pip
pip install pip --upgrade
#install virtualenv for python2
pip install virtualenv
#create virtualenv for mdanse at /opt/mdanse
python -m virtualenv /opt/mdanse/
#install python dependencies for mdanse
/opt/mdanse/bin/python -m pip install cython pyro numpy vtk matplotlib
#copy over python2 packages for wxPython and matplotlib-wx backend into virtualenv
#the needed files can be identified via
#repoquery -l wxPython; and; repoquery -l python2-matplotlib-wx
cp -R /usr/lib64/python2.7/site-packages/wx* /opt/mdanse/lib64/python2.7/site-packages/
cp -R /usr/lib/python2.7/site-packages/wxPython_common-2.8.12.0-py2.7.egg-info /opt/mdanse/lib/python2.7/site-packages/
#note the '\cp' is not a type in the following
\cp -R /usr/lib64/python2.7/site-packages/matplotlib/backends/* /opt/mdanse/lib64/python2.7/site-packages/matplotlib/backends/
#export path to netcdf.h header file so the following modules to be installed know where to look for it
export NETCDF_HEADER_FILE_PATH=/usr/include/
[...]

PR #94, opened 2022-05-23
Added support for HDF5 format for output files

eurydice76:
Hi @RastislavTuranyi and @sanghamitra-mukhopadhyay , I implemented the HDF file format support for the output of the analysis. Rastislav, to make it work, could you will please set h5py as a new dependency on the build server. The way I implemented it allows to use both NetCDF and HDF file in the MDANSE plotter. Please review carefully and give me some feedback/suggestions/corrections.


RastislavTuranyi:
I keep getting this error `Warning! ***HDF5 library version mismatched error***`. Have you experienced that?

eurydice76:
No, I did not have this warning on ubuntu20.04. Apart from that, does it work ?

RastislavTuranyi:
Well, the code seems ok, but I can't test it until I resolve this issue, not unless I want to risk `Data corruption or segmentation faults may occur if the application continues.`.

eurydice76:
Unfortunately I can't help you

eurydice76:
Maybe it can help https://github.com/h5py/h5py/issues/853

RastislavTuranyi:
Thanks for looking for solutions, but no I thread I have seen so far has been helpful/relevant. For now, I added installation of h5py in the pipeline, and also enabled the deploy for windows so that I can check if the GitHub Windows runner also has issues like me.

eurydice76:
OK, let's cross the fingers

eurydice76:
Hi @RastislavTuranyi , could you solve the warning with h5py on your machine ?

RastislavTuranyi:
I can get it to work in my own environment by using some files from conda, but I'm having issues with making it work with the executable generated by CD.

RastislavTuranyi:
I had to rewrite the Windows CI/CD to use conda python instead of the GitHub runner python, but I succeeded in making it produce an executable that works. However, I have discovered that when `UnitTests/AllTests.py` is run from a conda environment (and not just on the windows CI, but also on my laptop; on the other hand, this does not happen when the virtual environment is from `venv`), it gets stuck on opening the trajectory inside the `TestConfigurator.setUP()` method, never succeeding but never failing either. Can you please check if this happens for you too?

eurydice76:
I just tried on my mahcine with a conda environment and everything was OK

eurydice76:
Yeap, if that solves the problem, go ahead. However, it would be interesting to find why this trajectory triggers such infinite loop.

RastislavTuranyi:
I have found the issue! As can be seen from this log, [alog.log](https://github.com/ISISNeutronMuon/MDANSE/files/8799415/alog.log), the while loop as below never ends because the END line has a wrong amount of whitespace. Changing the if statement on line 73 into `if l[0].strip() == 'END': break` has fixed the issue. Should I directly push that into the ILL MMTK repo, open a PR on GitLab, or just deal with it inside the pipeline?

```
        while True:
            l = FortranLine(file.readline()[:-1], 'A4,6X,A2')
            if l[0] == 'END ': break
            set_name = _normalizeName(l[0])
            ljpar_set = AmberLJParameterSet(set_name, l[1])
            self.ljpar_sets[set_name] = ljpar_set
            self._readLJParameters(file, ljpar_set)
```

eurydice76:
The best would be to correct this in the ILL MMTK repo. Do you have the right to push ?

RastislavTuranyi:
Now that you mention it, I probably don't, given that I don't work on it, don't have an ILL user account, and don't have a gtilab account. It would probably be easier if you handled that, if that is ok?

eurydice76:
I commited the changes on MMTK repo. Everything should be OK now.

RastislavTuranyi:
It seemed that the all the tests were stuck in an infinite loop, so I cancelled the run. After checking the new commit in the ILL MMTK repo, I noticed that the new line, `if l[0].strip() == 'END ': break`, has a space in the string, `'END '` instead of `'END'`, which is breaking the trajectories since after calling the strip() method on the l[0] string, there are no spaces left.

eurydice76:
I just corrected the bug on MMTK. Sorry for not having been more careful.

eurydice76:
However, they are still problems with MMTK tests. I do not know what happened to these days which works perfectly so far

RastislavTuranyi:
OK, the CI/CD is now functional and creates installers that work. I had to remove the caching of python, which has slowed the windows CI even further, and the created installer is also larger than it used to be, but at least it works.

@sanghamitra-mukhopadhyay, for testing, I have updated the building instructions for windows in [#8](https://github.com/ISISNeutronMuon/MDANSE/issues/8#issuecomment-916223070). The functioning Windows installer is in [this run](https://github.com/ISISNeutronMuon/MDANSE/actions/runs/2423547472). MacOS and Ubuntu installers are in [this run](https://github.com/ISISNeutronMuon/MDANSE/actions/runs/2386973849).

eurydice76:
I think that we are now ready to merge.

RastislavTuranyi (approved):

sanghamitra-mukhopadhyay (approved):

RastislavTuranyi:
Just for the record, I would like to mention here that I have now found the root cause of this issue; it is due to the simultaneous use of h5py and netCDF4. It seems that when these two packages are installed with pip, each of them is built with a different version of HDF5 C binaries, and python, for some reason, cannot keep the packages linked separately. Therefore, when one is imported, its HDF5 binaries are then used for it as well as for the one that is imported later.

PR #96, opened 2022-06-06
Fix unit definitions

RastislavTuranyi:
While updating the user guide, I noticed that the lumen and lux units didn't have steradian in their definition, so I have remedied that. Now `1 lm = 1 cd sr` and `1 lux = 1 cd sr m^-2`.


eurydice76:
OK, good

sanghamitra-mukhopadhyay (approved):

PR #100, opened 2022-06-09
Fix AtomsListPlugin

RastislavTuranyi:
**Description of work.**

*There is no associated issue.*

Currently, when `AtomsListPlugin` is opened through a 'New definition' button in 'Axis selection' or 'Reference basis' section, the nAtoms value, which controls the number of atoms that must be selected for the definition to be valid for use in the analysis from which the plugin was opened, is always set to 1. The reason for this was that the method handling the opening of this plugin in `AtomsListWidget` had a name that did not overwrite the name in the base class (`UserDefinitionWidget`), and therefore wasn't called anywhere to the best of my knowledge. By changing the name, the `AtomsListPlugin` is now created with `nAtoms` set to the value that is expected by the analysis.

However, further change was made to disable the input field for `nAtoms` instead of hiding it completely, as it achieves similar purpose while also showing the user how many atoms they have to select for the analysis to work.

**To test:**

1. Open Angular correlation analysis
2. In the Axis selection window, click on 'New definition'
3. Observe that the 'Number of atoms' field is greyed out and is set to 2

---

#### Reviewer ####

##### Code Review #####

### DIFF
--- a/Src/GUI/Plugins/AtomsListPlugin.py
+++ b/Src/GUI/Plugins/AtomsListPlugin.py
@@ -136,7 +136,7 @@ class AtomsListPlugin(UserDefinitionPlugin):
         
     def enable_natoms_selection(self,state):
         
-        self._nAtomsSelectionSizer.ShowItems(state)
+        self._nAtomsSpinCtrl.Enable(state)
 
     def set_natoms(self,nAtoms):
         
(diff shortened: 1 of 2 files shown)


eurydice76 (approved):
Everything seems ok to me

sanghamitra-mukhopadhyay (approved):

PR #116, opened 2022-06-23
Set .nc and .h5 file browsing by default in plotter

RastislavTuranyi:
A small quality of life improvement that sets the file browser in 2D/3D Plotter (when loading data into plotter) to show both NetCDF and HDF5 files by default. I noticed that it's quite annoying to plot stuff at the moment since the default output file format of analyses is HDF5, but the default format of the plotter is NetCDF. Therefore, every time I load a file into the 2D/3D Plotter, I have to switch the file formats displayed from .nc to .h5. Now, both are visible, while the options to only see NetCDF or only HDF5 still exist. What do you think, @sanghamitra-mukhopadhyay?

Before:
image
Now:
image

### DIFF
--- a/Src/GUI/Plugins/PlotterPlugin.py
+++ b/Src/GUI/Plugins/PlotterPlugin.py
@@ -466,7 +466,8 @@ class PlotterFrame(wx.Frame):
                 
     def on_load_data(self, event=None):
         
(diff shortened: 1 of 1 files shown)


sanghamitra-mukhopadhyay:
Thank you for this. I also have noticed this and had to change it manually. I have reviewed it . All Okay.

sanghamitra-mukhopadhyay (approved):

eurydice76 (approved):
OK

PR #134, opened 2022-07-12
Expand README

RastislavTuranyi:
I have been thinking that the README could be improved since right now it's very short and only has a description. I added some information for new people, links to guides on installing from source, and organised the information with headings and some other formatting. I also added a section that provides a bit more depth on what MDANSE can do, both so that new people get a better idea of what it's for, and so that hopefully the keywords in it help it appear in search engines. What do you think?


eurydice76 (commented):
Excellent !

sanghamitra-mukhopadhyay (approved):

Issue #64, opened 2022-04-20
ENHANCEMENT: Prototype for porting MDANSE to python3


eurydice76:
@sanghamitra-mukhopadhyay @RastislavTuranyi @gonzalezma After weeks of thinking and working I could finally write a prototype for porting MDANSE to python 3. To do so, I wrote from scratch the class of `ChemicalSystem`  which is made of either `Atom`, `AtomCluster`, `Molecule`, `PeptideChain` or `NucleotideChain`. Those classes uses definitions of Atom, Residue; Nucleotide and Molecule stored in YAML files that I built on that occasion. The `ChemicalSystem` class is the light equivalent of the so-called `Universe` in MMTK. As such the `ChemicalSystem` keeps the full molecular hierarchy that you find in real system such as peptidic or nucleotidic chains. Based on this and using some parts of the MMTK PDB reader, I could derive a new PDB reader which seems OK when tried with a few PDB files. It is perhaps less universal than the one of MMTK but it has the advantage of fulfilling exactly our needs. The tricky part was to serialize a `ChemicalSystem` to an HDF5 file. To do so, I wrote some helper classes `H5Atom`, `H5Molecule`, `H5Residue` ... which are used through a recursive algorithm. This allows to store the chemical contents to HDF5. I also wrote from scratch a `Trajectory` class which represents a trajectory in memory. It reads the trajectory in HDF5 and builds from there a `ChemicalSystem` alongside with its configuration (coordinates, unit_cell, velocities ...). All the code is in Protos directory. Many work remains to do and that design can be improved but we finally have a good starting point for porting mdanse to python 3 by getting rid of MMTK and ScientificPython. To summarize:

    - no more dependency on ScientificPython and MMTK
    - no more dependency on python2
    - trajectories stored in HDF5 format
    - uses of YAML files to define the building blocks of a `ChemicalSystem`

eurydice76:
Hi @RastislavTuranyi @sanghamitra-mukhopadhyay , I made a lof of changes in the protos branch. This branch allows now to:
    - convert trajectories toHDF5 format (only works for DL_POLY converter right now)
    - run MSD analysis
    - produce a HDF file as output format
    - view the molecules with the molecular viewer
[...]

PR #137, opened 2022-07-26
Add unit tests for ChemicalEntity.py

RastislavTuranyi:
Hi @eurydice76, I have been looking over the new code and thought that the best way to understand it would be to write tests for it, so I did that. It's not finished yet, but I wanted to get this to you before my leave so you can look over it. I will be gone from today until 5.7., but please leave any comments here or push any changes yourself. We can discuss it when I return.

I have only gone over ChemicalEntity.py in detail so far. Overall, I think it works pretty well. The cyclic references are a bit concerning, but ultimately it should probably OK. There were some bugs and so on, but I should have fixed all of them. There are only two things that I was not able to resolve and I want your input on:

1. I mentioned this #64, but I am not sure how to deal with managing the atoms in the `AtomCluster` class. For now, I kept the atoms in a list (the way it was) and just fixed the bug in the `reorder_atoms()` method, taking into account the possibility of multiple atoms with the same name. However, this might not be the best solution. What is, though, depends on what purpose the `AtomCluster` class is supposed to play in MDANSE, and I am not sure what that is. Therefore, I wanted to ask you what you intend the purpose of  `AtomCluster` to be and how you intend it to be used. If you settle on an implementation that you think works best, please feel free to push that here.
2. The use of `eval()` in `ChemicalSystem.load()` is a security vulnerability in this case since it is used on data stored on disk. It would be trivial to create a malicious HDF5 file and send it to someone who would then use MDANSE to load that file, executing the malicious code within. I reduced this vulnerability, by restricting the eval's access to `__builtins__`, but [that is still not secure](https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html). I propose changing how the chemical entities are stored on disk; instead of storing strings containing data on instantiating H5ChemicalEntity objects, I think we should only store the arguments, and then instantiating the objects manually (in the code) using the stored data. We can sort out the details later, but I think that the security of users is the most important thing. Please let me know your opinions on this.

Besides that, I have done many more changes; here's the list:

1. Additions
    1. Unit tests: I have added unit tests for almost all classes. They should have a very good coverage, so if there is any change, we should be able to notice if anything breaks. Writing them has also allowed me to find a number of bugs.
    2. Documentation: I have added docstrings to all methods, and they should be fine, but I think they can still be improved, so please make any improvements you can see.
    3. Type hints: since type hinting has become a permanent feature of Python3, I tried adding type hints to all methods, and I have to say that it makes writing code much, much easier. I propose taking this opportunity to use type hints everywhere in MDANSE.
    4. `__str__` and `__repr__`: I implemented these methods in every class to make bugfixing, unit testing, and exceptions better.
    5. Exception handling in `NucleotideChain._connect_nucleotides()` and `PeptideChain._connect_residues()`: this should provide helpful errors is the `set_nucleotides()`/`set_residues()` method is called with invalid input.
2. Bugfixes
    1.  Correct bonds setter in `Atom`
    2. Fix `AtomGroup.__setstate__()`
    3. In `AtomCluster.copy()`, set the parent of copied atoms to the copied `AtomCluster` only if `parentless is True`
    4. Correct the exception handling in `Residue.__init__()` and `Nucleotide.__init__()`
    5. Correct method name in `NucleotideChain.copy()` from `self._connect_residues()` to `self._connect_nucleotides()`
    6. Fix total number of atoms count in `ChemicalSystem`
3. Improvements
    1. Improve the usefulness of error messages by providing more information
    2. Add the possibility to use `ChemicalSystem.load()` with an h5py.File object instead of just a string.
    3. Improve PEP8 etc. compliance
4. Changes
    1. Change all `serialize()` methods to conform to the ABC and accept only one argument. This is an h5py.File in `ChemicalSystem` and a dict in every other class. There was no point passing the file down to the other classes anyway since they were not using it.

Lastly, I have some questions and proposals:

1. Why does the `Atom.__init__()` only define kwargs instead of directly defining specific keyword arguments? I.e., why not do this: `def __init__(symbol='H', name=None, bonds=None, groups=None, ghost=False, **kwargs)` and then using if else  statements inside to save desired values?
2. I think that methods `atom_list()`, `number_of_atoms()`, `total_number_of_atoms()`, etc. would make more sense as `@property`, since they don't really do anything and are instead, well, properties of the chemical entities. What do you think?

Hopefully I got everything. Please let me know all the comments you have about this, and thank you for your help!

### DIFF
--- a/Extensions/.gitignore
+++ b/Extensions/.gitignore
@@ -1,6 +1,12 @@
+/atomic_trajectory.c
+/atoms_in_shell.c
+/change_coordinates.c
+/com_trajectory.c
+/contiguous_coordinates.c
 /mic_fast_calc.cpp
 /distance_histogram.c
 /fast_calculation.c
+/fold_coordinates.c
 /mt_fast_calc.c
 /qhull.c
 /sas_fast_calc.c
--- a/Src/Chemistry/ChemicalEntity.py
+++ b/Src/Chemistry/ChemicalEntity.py
@@ -572,6 +572,27 @@ class Atom(_ChemicalEntity):
 
         self._symbol = symbol
 
+    @classmethod
+    def build(cls, h5_contents: Union[None, dict[str, list[list[str]]]], symbol: str, name: str, ghost: bool) -> Atom:
+        """
+        Creates an instance of the Atom class. This method is meant to be used when loading a trajectory from disk and
+        so may be called when the :class: `ChemicalSystem`.load() method is called.
+
+        :param h5_contents: A parameter present solely for compatibility with other classes' build() methods. It is not
+            used.
+        :type h5_contents: None or dict
+
+        :param symbol: The chemical symbol of the Atom. It has to be registered in the ATOMS_DATABASE.
+        :type symbol: str
+
+        :param name: The name of the Atom. If this is not provided, the symbol is used as the name as well
+        :type nameL str
+
+        :param ghost:
+        :type ghost: bool
+        """
+        return cls(symbol=symbol, name=name, ghost=ghost)
+
     def serialize(self, h5_contents: dict[str, list[list[str]]]) -> tuple[str, int]:
         """
         Serializes the Atom object into a string in preparation of the object being stored on disk.
@@ -592,7 +613,8 @@ class AtomGroup(_ChemicalEntity):
     """
     An arbitrary selection of atoms that belong to the same chemical system. Unlike in Molecule and AtomCluster, the
     atoms in an AtomGroup do not have to be related in any way other than that they have to exist in the same system.
-    Further, AtomGroup does not have the serialize() method defined, and so it will not be saved on disk.
+    Further, AtomGroup does not have the serialize() or the build() method defined, and so it will not be saved on disk
+    and cannot be loeaded from disk.
     """
 
     def __init__(self, atoms: list[Atom]):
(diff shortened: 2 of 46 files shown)


eurydice76:
Hi @RastislavTuranyi , first of all a big big thanks for the deep review you did on the protos branch. It was clear to me that there were remaining bugs and other problems given the amount of work I put in there and the pace at which I implemented them. To answer to your question:
1) the goal of atom cluster is to provide an API for storing a set of atoms which are not bonded or for which the bond network is not known at runtime. In the original design, I implemented it as a list of atoms as for me several atoms with the same name could be present in an AtomCluster. Rethinking about his, I think that AtomCluster should have the same constraint than the other chemical entities i.e. not containing duplicate atom name. That's definitely safer. So, a new implementation will certainly use a dict of atoms instead of a list.
2) you are absolutely right about your consideration about eval. That's a risk that I thought about when writing this code. Unfortunately, at the time I had no better design in my pocket ! One way would be to store in the HDF not a string such as 'Atom(name="H")' but instead an array of the form ['Atom','name="H"']. Then in MDANSE we would store a lookup which connect the name of the chemical entities with the actual class behind. Something like d = {"Atom":Atom ...}. In such way we are safe from any malicious attack. I have no strong opinion about the design and I am opened to any new suggestion.

About or last questions 1) and 2), you are right for both questions, it could designed on the way you recommend. Especially for point 2 which proposes the use of property for which you know I am a big fan.

Thanks again

eurydice76 (commented):
Nice work. Thanks

RastislavTuranyi:
I see, thanks for your input! I have implemented the keyword arguments and  properties and will now look into changing the serialisation. For `AtomCluster`, do you think it's enough to simply raise an error if the input contains two atoms with the same name, or would you prefer for there to be some (possibly just for the dictionary inside `AtomCluster`) renaming mechanism?

RastislavTuranyi:
@eurydice76, I have implemented a new way for serialising: the arguments used to reconstruct a class are stored in the HDF5 file as an array of strings (therefore each Dataset is now an array of arrays of strings). The H5 objects are instantiated via a dictionary which is accessed in the code by using the `enitity_type` values from `skeleton`. Please let me know any comments, and after we both agree, I'll regenerate all the HDF5 files in Data/ using this new implementation.

RastislavTuranyi:
I have come across this `DeprecationWarning` when doing `_ChemicalEntity.center_of_mass()` and wanted to ask you which one you had in mind: float or np.float64?

```
<input>:1: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

RastislavTuranyi:
Hi @eurydice76, I have taken a look at H5ChemicalEntity.py since it also has to be changed to match the new serialisation system, but I'm struggling to see why there was a need to put the `build()` functions into separate classes. Thus, I would like to ask you about the reasoning that lead to this design decision as well as propose to handle the function that these classes perform directly in ChemicalEntity.py instead. What do you think?

eurydice76:
Hi Rastislav, thanks for your comments. I am currently on holidays so I can not deeply review your changes. For the DeprecationWarning I had np.float64 in mind. For the build method, I must admit that I forgot why I designed the code that way. Perhaps there was a good reason. Perhaps not. If your design is functional go ahead, I fully trust you.

RastislavTuranyi:
All right, I have finally finished implementing a new loading mechanism  to replace the H5Entity classes. I have settled on adding a class method to each Entity class that returns a new instance of that class. I also attempted to make the error handling during parsing comprehensive and helpful. Eric, can you please take a look and let me know what you think?

It would also be great if you could check out the other parts of I/O and have a think about the `AtomCluster` issue we discussed previously (as quoted below).

> I see, thanks for your input! I have implemented the keyword arguments and properties and will now look into changing the serialisation. For `AtomCluster`, do you think it's enough to simply raise an error if the input contains two atoms with the same name, or would you prefer for there to be some (possibly just for the dictionary inside `AtomCluster`) renaming mechanism?

PR #140, opened 2022-08-12
Add database tests

RastislavTuranyi:
I have rewritten and expanded the tests for AtomsDatabase and added tests for the other database objects. Like before, I have also expanded the documentation and added type hints. I made only two major changes:

1. I have refactored out a base class `_Database` of which all the other databases are a subclass of. I have moved the methods that were common to all or most of the classes there.
2. I have added a new method to `AtomsDatabase` that is able to retrieve the value of a given property for a specified list of atoms efficiently. It is made with the assumption that the list of atoms has a much greater length than the variety (e.g. 5000 H2O molecules = 5000 O atoms + 10000 H atoms) which should hold true for most chemical systems. In those cases, it is much faster than in a loop repeatedly calling `ATOMS_DATABASE[atom][property]`. I have rewritten the `centre_of_mass()` method of `_ChemicalEntity` using this new method and numpy, which has sped it up. I believe that it should work the same as the previous implementation, but please check.

Please take a look and let me know any comments you have.

Lastly, though, I would like to ask you about the singleton. From reading and watching on the internet, I have got the impression that the singleton design pattern is universally considered bad coding. Here are some examples: [why singletons are evil](https://web.archive.org/web/20210715184717/https://docs.microsoft.com/en-us/archive/blogs/scottdensmore/why-singletons-are-evil), [why singletons are controversial](https://web.archive.org/web/20210506162753/https://code.google.com/archive/p/google-singleton-detector/wikis/WhySingletonsAreControversial.wiki), [youtube video on singleton](https://www.youtube.com/watch?v=hUE_j6q0LTQ), etc. Personally, I'm convinced that it is a good idea limit the use of this design pattern as much as possible. However, I would like to ask you what you think about this. Do you think it is feasible to change these databases into not-singletons?


RastislavTuranyi:
Another thing: the docstrings of some of the __getitem__ methods have the following mentioned, but there is no matching implementation; should this be removed from the docstring, or should the described behaviour be implemented?
`If the item is a 2-tuple then the return value will the property of the database whose element and property match respectively the first and second elements of the tuple.`

eurydice76:
Hi @RastislavTuranyi,

yes these docstrings are deprecated. This part of the docstring should be removed.

eurydice76 (commented):
OK. However, I am not sure that the gain in rewriting the center_of_mass code is so clear. OK, you avoid a loop but the code greatly loses in clarity imho.

eurydice76:
If the code is such faster keep your change. That's perfect.

RastislavTuranyi:
Ok, cool! The last thing I wanted to ask here about is the Singleton pattern. Do you think everything would work correctly if we removed it from the databases? (i.e. remove the `metaclass=Singleton`, with everything else the same, for aforementioned reasons)

eurydice76:
Considering `Singleton` as a bad pattern is a long story ... The problem here is to ensure that we have a single instance of a given database (Atoms, Molecules ...) running throughout the code. I do not see a lot of alternative designs to ensure that, excepted the use of a global variable. In that case, you will fall on preachers that claims that global variables should in turn be avoided ... I really dislike those guys who claims this or that to be bad design. Sometimes, you simply have no choice.

RastislavTuranyi:
I am not proposing removing the `ATOMS_DATABASE` etc. and its use throughout the code since, as you said, we want every part of MDANSE to use the same database (for each database). However, unit testing would be easier if this was not enforced through the Singleton metaclass, i.e. all the code in Src/ would continue using the instance of databases created in MDANSE.Chemistry.__init__.py (therefore the same instance) while unit tests would be able to create a new instance. Would this work? Would it be thread-safe?

Also, this is only partly related, but it has just occurred to me that I'm unsure what the current behaviour of MDANSE is if a user alters e.g. the ATOMS_DATABASE while a job is currently running. Does the job continue using the values it had been using when it started, or does it start using the newly changed values, even though it's midway through?

eurydice76:
OK, but using a chemical database instance stored in __Chemistry/__init__.py__ corresponds to a global variable and I can find on the web billions of posts that say that using global variable is considered bad design too ... The problem is that we need a single instance of the chemical database throughout the code. I do think that the Singleton is safer than the global variable approach. Regarding, the risk of altering the db by another job: this can not happen as a job is a different process. As such it will use the db in the state it is in the db file.

RastislavTuranyi:
Oh, I didn't know about that! I have done a bit of studying on this, and indeed I cannot think of a way of removing both. Since you say that removing the singleton is potentially unsafe, that was the last of my comments, so I guess we can merge this?

RastislavTuranyi:
Actually, @eurydice76, since we are keeping the singleton, why don't we make the full use of it? Please correct me if I am wrong, but with the current implementation, when a trajectory is loaded for the first time, _all_ the databases are loaded from the yaml files into memory. However, if we stop creating an instance of each database in Chemistry.__init__.py (i.e.  `ATOMS_DATABASE = AtomsDatabase()` etc.) and then change the files that use the databases to call the database constructor instead of the variable (e.g. `MOLECULES_DATABASE.items()` → `MoleculesDatabase().items()`), only the databases that the trajectory requires would be loaded into memory. That should lower the memory usage a bit as well as speed up the opening of most trajectories. What do you think?

PR #145, opened 2022-08-16
Improve Configuration tests

RastislavTuranyi:
As with the other PRs, here I have also mostly just added type hints, expanded documentation. The tests now cover all classes and all methods;; I tried keeping the values from the previous tests where I was able to see which ones did what. In terms of documentation, I reformatted it to match the style from other parts of MDANSE, particularly the ones I have reviewed already. There were no major changes, only a couple minor ones:

1. I renamed the `atomsInShell()` method to `atoms_in_shell` so as to match both the names of the other methods and PEP8.
2. In the `contiguous_configuration` and `continuous_configuration` methods of `_PeriodicConfiguration` subclasses, I have changed the line `conf = self` to `conf = self.clone()` since I reckoned that the purpose of those methods was to return a new object rather than modify itself and then return the modified self. Please correct me if I am wrong.
3. I had to add `.astype(np.float64)` when passing the coordinates to some of the Cython functions to get them to work
4. I modified the `contiguous_offsets` methods to pass in to the Cython function only the coordinates corresponding to the atoms that were passed in via the chemical_entities argument.
5. I had to add the `casting='unsafe'` argument to np.add in `_Configuration.apply_transformation` method to get it to work with rotations.

Please have a look and let me know all your comments.

That said though, the tests are not passing, and with that, I would like to ask for your help. The tests that are failing I have kept that way because I do not understand what/why exactly the method is doing, so I didn't want to just copy the results over. In fact I'm a bit sceptical about some of the results, so I would like to discuss with you.

1. `_Configuration.apply_transformation()` - in this one, it's, I think, simply an issue me not understanding how the transformations work. I attempted to test this by rotating 180° around the x-axis, but the results I'm getting are different from that, so I think the Rotation object I created simply does something else. Can you please check this? If the result of the method is correct for the transformation, I'll just copy the results as the expected results.
2. `PeriodicBoxConfiguration.atoms_in_shell()` and `PeriodicRealConfiguration.atoms_in_shell()` - here, I genuinely believe that with the given parameters, the output should be [1, 2] instead of [1, 2, 3]. Further, I don't understand what's going on in the Cython code; why are these lines `sdx -= round(sdx)` present? Rounding by itself I could understand, but why is it subtracted?
3. `PeriodicBoxConfiguration.contiguous_offsets()` and `PeriodicRealConfiguration.contiguous_offsets()` - in this case, I have no idea what's going on, but I would assume that, since it's an offset, these methods would return the difference between the coordinates calculated by the `contiguous_configuration()` methods and the normal coordinates, but that's not the case. If the current results are correct, I''ll just copy them as the expected results, but please have a look.

Lastly, I have no idea what the methods `PeriodicBoxConfiguration.contiguous_configuration()`, `PeriodicRealConfiguration.contiguous_configuration()`, and `PeriodicRealConfiguration.continuous_configuration()` are calculating, so I just copied the current output as the expected output, but because of that, I would appreciate if you could check these as well that they are doing what they are supposed to be doing.

Thank you for your help in advance!

### DIFF
--- a/Extensions/atoms_in_shell.pyx
+++ b/Extensions/atoms_in_shell.pyx
@@ -130,8 +130,6 @@ def atoms_in_shell_box(ndarray[np.float64_t, ndim=2]  config not None,
 
     cdef int i
 
-    cdef ndarray[np.float64_t, ndim=2] scaleconfig = np.empty((config.shape[0],3),dtype=np.float)
-
     cdef double mini2 = mini**2
     cdef double maxi2 = maxi**2
 
@@ -150,9 +148,9 @@ def atoms_in_shell_box(ndarray[np.float64_t, ndim=2]  config not None,
         sdy = config[i,1] - refy
         sdz = config[i,2] - refz
 
-        sdx -= round(sdx)
-        sdy -= round(sdy)
-        sdz -= round(sdz)
+        sdx = round(sdx)
+        sdy = round(sdy)
+        sdz = round(sdz)
             
         r2 = sdx*sdx + sdy*sdy + sdz*sdz
 
--- a/Extensions/atoms_in_shell.pyx
+++ b/Extensions/atoms_in_shell.pyx
@@ -38,28 +38,28 @@ def atoms_in_shell_real(ndarray[np.float64_t, ndim=2]  config not None,
 
     cdef int i
 
-    cdef ndarray[np.float64_t, ndim=2] scaleconfig = np.empty((config.shape[0],3),dtype=np.float)
+    cdef ndarray[np.float64_t, ndim=2] scaleconfig = np.empty((config.shape[0], 3), dtype=np.float64)
 
     cdef double mini2 = mini**2
     cdef double maxi2 = maxi**2
 
-    refx = config[refIndex,0]
-    refy = config[refIndex,1]
-    refz = config[refIndex,2]
+    refx = config[refIndex, 0]
+    refy = config[refIndex, 1]
+    refz = config[refIndex, 2]
 
     for 0 <= i < config.shape[0]:
 
-        x = config[i,0]
-        y = config[i,1]
-        z = config[i,2]
+        x = config[i, 0]
+        y = config[i, 1]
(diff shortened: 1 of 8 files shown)


eurydice76 (commented):
For me the code is OK.

eurydice76:
Hi @RastislavTuranyi,

1. the apply transformation code, is an old code that comes from the predecessor of MDANSE, aka nMOLDYN. I amm not the author of this code and I must admit that I copied it as is.
2. the `sdx -= round(sdx)` is used as the final step of the periodic boundary condition algorithm. It allows to fold the coordinates into the simulation box.
3. As far as I remember when I wrote it, `contiguous_configuration` allows to get a configuration where the coordinates are folded into the simulation box. `continuous_configuration` does more or less the same job but is an improved version. As such `contiguous_configuration` should be considered as deprecated. However, for the moment, I prefer to keep it.

RastislavTuranyi:
I see, thank you for the clarification. I will look into quaternions to get a better idea of 1.. After playing with the atoms_in_shell method and with PBC in mind, I think I understand what it does, so I have updated the tests. However, I got these questions:

1. Since the `atoms_in_shell()` methods of `PeriodicBoxConfiguration` and `PeriodicRealConfiguration` deal with PBC, should we raise an error if the specified radius is larger than the unit cell?
2. Since the `atoms_in_shell()` method in `PeriodicBoxConfiguration` uses box coordinates for the calculations, should we convert the inputted `mini` and `maxi` arguments into box coordinate units as well? Or, would it be better to just change the documentation to say that the expected values are supposed to be in box units and not nm?

Lastly, your description of `contiguous_configuration` sounds very similar to the `fold_coordinates()` method, but the Cython code is markedly different for each; what's the difference between them?

eurydice76 (commented):
Changes are OK. Nice work for the documentation of the Configuration related classes.

eurydice76 (commented):
OK. Nice work for the documentation of Configuration related classes.

PR #156, opened 2022-09-12
Add acknowledgements

franzlang:
Adds an acknowledgements section to the README. They were taken from the current MDANSE user guide: https://www.isis.stfc.ac.uk/Gallery/RAL-TR-2022-003.pdf


RastislavTuranyi:
Should we not also add a citation to the nMOLDYN paper? 

> T. Róg, K. Murzyn, K. Hinsen, and G. R. Kneller, “nMoldyn: A program package for a neutron scattering oriented analysis of molecular dynamics simulations,” J. Comput. Chem., vol. 24, no. 5, pp. 657–667, Apr. 2003

franzlang:
good point. I have added a citation to the publication for the latest version of nMoldyn (v3). that paper contains citations of the previous versions' publications.

RastislavTuranyi:
Two more things:

1. I think we should keep the citation style consistent among the two citations. Not sure which style each one is in, but choosing one (or a completely different) and using it for both citations should make things clearer.
2. Isn't the Acknowledgements section a bit too high up? I would expect anyone who comes to our GitHub to be way more interested in Quick Start than in the history of MDANSE. Personally, I see Acknowledgements as extra information meant for those interested, so I would put it either below the Citing MDANSE section of below License section. What do you think?

RastislavTuranyi (approved):

franzlang:
I made the citation style the same as for the other citations and moved the acknowledgements further down in the README.

sanghamitra-mukhopadhyay (approved):

eurydice76 (commented):
Sorry for being late, the acknowledgements are fine with me

PR #161, opened 2022-09-14
Add tests for Analysis and TrajectoryUtils

RastislavTuranyi:
As in the other PRs, I have mostly just expanded documentation, added type hints, and written unit tests. The only code change is that in `cpt_cluster_connectivity_nopbc` no longer takes an empty list as an argument and instead creates an empty list itself. I can't remember exactly why I made that change, but I think it was causing a bug. However, I have one question:

1. In `build_connectivity()`, shouldn't we call the `cpt_cluster_connectivity_nopbc` function only if the configuration is `RealConfiguration`, and then use `cpt_cluster_connectivity_pbc` for when the configuration is `PeriodicBoxConfigutaion`? Further, shouldn't there be separate functions for `PeriodicBoxConfigutaion` and `PeriodicRealConfiguration`?

### DIFF
--- a/Extensions/fast_calculation.pyx
+++ b/Extensions/fast_calculation.pyx
@@ -17,14 +17,17 @@ import cython
 import numpy as np
 cimport numpy as np 
 
+
 cdef extern from "math.h":
 
     double floor(double x)
     double ceil(double x)
     double sqrt(double x)
 
+
 cdef inline double round(double r):
     return floor(r + 0.5) if (r > 0.0) else ceil(r - 0.5)
+
 
 @cython.cdivision(True)
 @cython.boundscheck(False)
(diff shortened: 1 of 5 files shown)


eurydice76 (approved):
OK

eurydice76:
Hi Rastislav,
I used the `cpt_cluster_connectivity_nopbc` function in the `build_connectivity` method because, if I remember right, otherwise it triggers nasty representation of the molecule in the molecular viewer. Indeed, if it found bonds between two atoms that are on say opposite side of the box then the moleculat viewer ends up with straight lines crossing the simulation box.

RastislavTuranyi:
I see, if the Molecular Viewer is the only place that this function is used, then that is fine. I have done a small change so that the `to_real_coordinates()` method is used to retrieve coordinates from the configuration to ensure that the coordinates passed in to the `fast_calculation.cpt_cluster_connectivity_nopbc` function are always real and never box. If that's fine, I guess we can merge, right?

Issue #159, opened 2022-09-14
Animation crashes on Windows 11


RastislavTuranyi:
When the Animation plugin is opened on Windows 11 obtained with the installer, the whole MDANSE freezes and then crashes without any error. Everything else works well enough, including the 3D plots and Density Superposition. There are no issues on Windows 10.

ChiCheng45:
I've tested MDANSE on Windows 11 and the GUI loads up fine now.

Issue #157, opened 2022-09-12
MDANSE does not launch on macOS 11.6 M1 chip


perenon:
Hello,
With Eric, we tested MDANSE on a macOS M1 machine, with system BigSur (11.6).

MDANSE crashed on startup with no error message.
However, using the provided python and lauching manually MDANSE gives:
```
$ /Applications/MDANSE.app/Contents/MacOS/python2
Python 2.7.18 (default, Oct 29 2020, 07:55:29) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import MDANSE
>>> from MDANSE.GUI.Apps import MainApplication
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Applications/MDANSE.app/Contents/Resources/lib/python2.7/MDANSE/GUI/Apps.py", line 19, in <module>
    from MDANSE.GUI.Plugins.PlotterPlugin import PlotterFrame
  File "/Applications/MDANSE.app/Contents/Resources/lib/python2.7/MDANSE/GUI/Plugins/PlotterPlugin.py", line 21, in <module>
    import netCDF4
  File "/Applications/MDANSE.app/Contents/Resources/lib/python2.7/site-packages/netCDF4/__init__.py", line 3, in <module>
    from ._netCDF4 import *
ImportError: dlopen(/Applications/MDANSE.app/Contents/Resources/lib/python2.7/site-packages/netCDF4/_netCDF4.so, 2): Library not loaded: @rpath/libnetcdf.13.dylib
  Referenced from: /Applications/MDANSE.app/Contents/Resources/lib/python2.7/site-packages/netCDF4/_netCDF4.so
  Reason: image not found
```

So we relocated netcdf lib
`$ install_name_tool -change @rpath/libnetcdf.13.dylib @executable_path/../Frameworks/libnetcdf.13.dylib /Applications/MDANSE.app/Contents/Resources/lib/python2.7/site-packages/netCDF4/_netCDF4.so`

`$ install_name_tool -change @rpath/libhdf5.103.dylib @executable_path/../Frameworks/libhdf5.103.dylib /Applications/MDANSE.app/Contents/Resources/lib/python2.7/site-packages/netCDF4/_netCDF4.so`

And we are facing now an issue with matplotlib
```
$ /Applications/MDANSE.app/Contents/MacOS/python2
[...]

Issue #147, opened 2022-08-22
MDANSE cannot be started on MacOS 12


RastislavTuranyi:
After installation, the MDANSE for MacOS cannot be started on MacOS 12. No error message is returned, but it appears to be caused by linking of the dylib libraries.

sanghamitra-mukhopadhyay:
Okay, I see. I did not check on Mac 12.

RastislavTuranyi:
Does everything work on the older version of Mac then?

PR #163, opened 2022-09-20
Update CI/CD on Protos

RastislavTuranyi:
**Description of work**

The CI/CD has been updated for Protos (Python 3) such that it works in (nearly) its entirety, generating valid installers for all supported platforms. Besides bugs with the GUI (such as some parts like Animation and 2D/3D Plotter either not working or crashing), the generated installers seem to work. Most importantly, the issues described in #147 and #157 do not appear. Issue #159 is difficult to ascertain since Animation does not work on other platforms too.

I have tested on Windows 10, Windows 11, Ubuntu 20, Ubuntu 21, and Ubuntu 22, and MDANSE there installs smoothly except for some issues with Ubuntu 20 and Ubuntu 21, which I am still fixing.

**To test**

1. Download an installer for your platform from the latest push on this branch and install it.
2. Open GUI and see if things generally work (besides bugs like described above)
3. Start the enclosed Python interpreter and check that it works.


MBartkowiakSTFC (commented):
On MacOS 12.6, ARM64 architecture, the GUI starts without errors, and it is possible to load a trajectory. After that, the input windows for analysis are unusable: all the widgets appear in the top left corner, on top of each other.
So, this build is not usable on my computer.

RastislavTuranyi:
Isn't that just another GUI bug that needs to be fixed? I have checked this on my Windows 11 laptop, and the analysis windows don't open there either. I think we can still merge this since it seems the installer works correctly on you Mac, but we can also wait for all the GUI bugs to be fixed on protos to check and only then merge this PR.

MBartkowiakSTFC:
> Isn't that just another GUI bug that needs to be fixed? I have checked this on my Windows 11 laptop, and the analysis windows don't open there either. I think we can still merge this since it seems the installer works correctly on you Mac, but we can also wait for all the GUI bugs to be fixed on protos to check and only then merge this PR.

If this is as good as you expected, then we can merge it, sure.

perenon:
Tested on macOS 12.6 M1 chip:
- Installation is OK
- HDF trajectories can be loaded
- Molecular viewer seems functional
- I could even launch an analysis (Thermodynamics > Density)
- `import MDANSE` in the console is fine
- Yes, there are still many bugs apart from these points ;-)

I think it would be useful to merge, so it will ease the future testing phase.

Cheers,
Remi

Issue #167, opened 2022-09-22
Create a converter for CP2K package trajectories


eurydice76:
I have been asked at the ILL to create a converter to CP2K ab-initio package. Right now, the development has been old on the python2-MDANSE.

PR #168, opened 2022-09-22
Create a converter for CP2K package trajectories

eurydice76:
See #167 for info


RastislavTuranyi (changes requested):
When I run the tests, I get this with BuildJobTests.py `/!\ Reference data file is not present for job <class 'MDANSE.Framework.Jobs.CP2K.CP2KConverter'>` which means that the test fails. Can you please generate, somehow, a reference trajectory? I have re-enabled the CI (accidentally disabled previously) so that you can see the test results for yourself.

eurydice76:
That's ok, I put a reference trajectory. The tests are passing now. Thanks for the feedback.

MBartkowiakSTFC (approved):
Thanks for adding the reference file! It looks good, and the tests are passing.

Issue #170, opened 2022-09-30
Add the possibility of some jobs parameters to be optional


eurydice76:
I have been asked in the context of the CP2K converter to make the input file for velocites optional. On a general point of view the optional state is related to the input file (whether it is set or not ) as for the other configurator. This requires a small refactoring of the `Configurable` class and related `InputFileWidget` class.

PR #171, opened 2022-10-04
170 optional parameters

eurydice76:
See #170 for info.


MBartkowiakSTFC (approved):
I did my testing on Ubuntu in VirtualBox. CP2K trajectory conversion works both with and without the velocity file specified as input. I have not found any problems with the changes.

Issue #172, opened 2022-10-07
GUI for trajectory conversion crashes on MacOS 11 (x86 CPU)


MBartkowiakSTFC:
The problem has been described here:
https://github.com/ISISNeutronMuon/MDANSE/pull/164#issuecomment-1261308229

It seems that the wxPython libraries are trying to allocate excessive resources for no obvious reason. I originally thought that the issue may be related to the problem described here:
https://forum.kicad.info/t/kipython-scripting-console-greyed-out-extensions-not-working/36249/13

MBartkowiakSTFC:
@sanghamitra-mukhopadhyay When you have the time, please check if this build works for you on your Mac. Several changes have been merged into this branch.
https://github.com/ISISNeutronMuon/MDANSE/suites/8667301331/artifacts/390236540

eurydice76:
Hi @MBartkowiakSTF, do you have any news for this issue ?

eurydice76:
Does the crash occur on a particular converter ? I can not reproduce it on my (ubuntu) machine so far.

MBartkowiakSTFC:
I think that the crash occurs on dialog creation, before a converter is even started. If I understand it correctly, it happens for all of them.
I could not reproduce it either. On my Mac it works fine, and on my Ubuntu it is fine too.
A colleague from the Mantid team is testing it now on a Mac with an x86 CPU. We will see if he gets it to crash or not.

MBartkowiakSTFC:
Martin checked the same build of MDANSE on a Mac with an x86 CPU, and he could not reproduce the problem either.

sanghamitra-mukhopadhyay:
Unfortunately I am still facing the same problem to open the trajectory converter. My Mac is with 3.3 GHZ intel i7 processor.

eurydice76:
Hi @sanghamitra-mukhopadhyay Can you give me the result of the command:

ls _crashing-directory_ | wc -l

where _crashing directory_ is the directory that is mentionned in the error message that you get.

eurydice76:
[...]

PR #173, opened 2022-10-17
172 crashing trajectory converter dialogs

eurydice76:
See #172 for info.


MBartkowiakSTFC (approved):
Thanks for the changes! It looks like we have a build of MDANSE that works on all our target platforms again.

PR #175, opened 2022-11-02
Disable MolecularViewer on MacOS

MBartkowiakSTFC:
The MolecularViewerPlugin now checks what platform it is on. If it is on MacOS ("Darwin"), it sets the ancestor of the plugin to a different value.


MBartkowiakSTFC:
Also, with the fixed numpy==1.16.6, the Windows build works again.

eurydice76 (commented):
OK. Nothing to add.

Issue #177, opened 2022-11-21
CP2K trajectory conversion: time steps not matching


MBartkowiakSTFC:
Unintentionally, I created an example of input which seems correct, but does not work in MDANSE. When I try to convert my CP2K trajectory, I get the error message:

```
/var/folders/jx/__bc7gns12g9b7f_v09x8mzm0000gq/T/MDANSE_maci_54769_2a14cg.py0WvNjY -- 24 -- <module> -- cp2k.run(parameters,status=True)
/Applications/MDANSE.app/Contents/Resources/lib/python2.7/MDANSE/Framework/Jobs/IJob.py -- 326 -- run -- raise JobError(self,tb)
/Applications/MDANSE.app/Contents/Resources/lib/python2.7/MDANSE/Framework/Jobs/IJob.py -- 47 -- __init__ -- tback = traceback.extract_stack()

Traceback (most recent call last):
  File "/Applications/MDANSE.app/Contents/Resources/lib/python2.7/MDANSE/Framework/Jobs/IJob.py", line 304, in run
    self.initialize()
  File "/Applications/MDANSE.app/Contents/Resources/lib/python2.7/MDANSE/Framework/Jobs/CP2K.py", line 187, in initialize
    raise CP2KConverterError("Inconsistent time step between pos and cell files")
CP2KConverterError: 'Inconsistent time step between pos and cell files'
```

I attached a trimmed trajectory containing just the first three steps, which generates the same error. As far as I can tell, the time values are exactly the same in the three files. What did I miss?

[sample_cp2k_trajectory.zip](https://github.com/ISISNeutronMuon/MDANSE/files/10058470/sample_cp2k_trajectory.zip)

Side note:
I had to create the cell file manually, since I had not included it as output in my simulation. The unit cell size was fixed in the simulation, so I did not see any reason to write it out when I was running the simulation - but later I discovered that it is needed in MDANSE.

eurydice76:
OK, I got the bug. Usually, in most of the simulations the first step has a time step of 0.0. So to get the time step I read the second step of the simulation and get the time step from there. In your case, the situation is different: the first step is not 0.0 but already one time step away from zero (i.e. 0.4). I hope that I did not confuse you ! I will provide a fix.

MBartkowiakSTFC:
Thanks for solving the problem. I tried it out on two different trajectories, and it worked for both of them.

PR #179, opened 2022-11-24
Fixed bug when computing the time step from CP2K XYZ file

eurydice76:
See #177 for info


MBartkowiakSTFC (approved):
Thanks for the changes! I loaded my _full_ trajectory now, and did not have any problems.

Issue #182, opened 2023-01-13
Units Editor blocks the GUI on MacOS


MBartkowiakSTFC:
On MacOS the Units Editor dialog opens as modal dialog, blocking all the other parts of the GUI. However, after closing the Units Editor other elements of the GUI remain blocked, and MDANSE needs to be shut down.

PR #183, opened 2023-01-13
Add EndModal before closing the dialog

MBartkowiakSTFC:
This is a simple fix to make sure that the GUI remains responsive after the Units Editor has been closed.

This should fix the problem described here:
https://github.com/ISISNeutronMuon/MDANSE/issues/182


eurydice76 (approved):
Everything seems OK to me. You can proceed to merge.

PR #192, opened 2023-03-10
Improved the contents of the output file in case of ASCII selection

eurydice76:
@MBartkowiakSTFC @sanghamitra-mukhopadhyay @gonzalezma After our discussion, I could finally find a way to implement a better output when ASCII is selected as the output file format for the analysis. Now, for 1D variables, the output file contains two columns one for the Y axis and one for the X axis and for 2D variables the output contains the matrix with the X axis as the 1st column and the Y axis as the 1st row. While doing this, the support for data whose dimensionality is > 2 has been stopped because 1) so far there has never been such data produced by MDANSE 2) it makes nearly impossible the handling of the dependant axis when producing the output.  I also fixed some bugs in "time_window" output variable which was set as depending upon "time" variable which is not true.

### DIFF
--- a/Src/Framework/Formats/ASCIIFormat.py
+++ b/Src/Framework/Formats/ASCIIFormat.py
@@ -53,11 +53,10 @@ class ASCIIFormat(IFormat):
         tf = tarfile.open(filename,'w')
         
         for var in data.values():
-
             tempStr = StringIO.StringIO()
             tempStr.write(var.info())
             tempStr.write('\n\n')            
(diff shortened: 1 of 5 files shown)


eurydice76:
No problem, it is a pleasure to help you. I had free time on Friday afternoon after the meeting so I thought it was the right time for doing this. I did a few tests about my implementation but it needs perhaps to be dug a bit further. Thank you for doing that.

PR #193, opened 2023-03-22
Update documentation before 1.5.2 release

MBartkowiakSTFC:
I tried to create a concise description of the analysis workflow, which is now located in its own section of the documentation, just before the description of the specific analysis types.

Also, I extended the description of different input parameters, especially the instrument resolution.

Where suitable in the Input Parameters section, a mathematical formula was added alongside the text.

Please check if you think that the changes I made correspond to the way MDANSE really works.


Doc/pages/scattering.rst
```diff
@@ -947,13 +951,33 @@ one is, etc.
...
 
-.. _neutron-dynamic-total-structure-factor:
+.. _analysis-ndtsf:
 
 Neutron Dynamic Total Structure Factor
 ''''''''''''''''''''''''''''''''''''''
 
 -  available for trajectories only
 
+This is a specific combination of the Dynamic Coherent and the
```
gonzalezma, line 961: I think that this is not very clear. I will try to prepare an alternative text and then I let you decide what to write.

Doc/pages/workflow.rst
```diff
@@ -0,0 +1,90 @@
...
+your trajectory is in whatever format was output by your preferred Molecular Dynamics
+simulation software, and you need to convert it first. Once you have converted your
+trajectory to the MDANSE NetCDF format, you can use it as input for all kinds of
+analysis. See also :ref:`trajectory-converters`)
+
+Frames
+^^^^^^
+
+It is not necessary to use all the time frames of your MD simulation. You can decide
+to limit the range of simluation time, and reduce the number of the frames taken in that
```
gonzalezma, line 27: correct simluation

Doc/pages/workflow.rst
```diff
@@ -0,0 +1,90 @@
...
+The partial properties calculated so far will now be combined using the weights
+chosen by the user, as described in the section :ref:`param-weights`. Please remember
+that the MDANSE GUI normally recommends the weighting scheme appropriate to the
+type of analysis performed.
+
+Writing output
+--------------
+
+All the output arrays created in the analysis are written to the filesystem in the
+format chosen by the user. (If you intend to continue visualsing the results within
```
gonzalezma, line 87: correct visualsing

gonzalezma (commented):
This is fine with me, but I would like to propose an alternative text for the explanation of the ndtsf analysis. See comments in the code for the few typos that I found.

MBartkowiakSTFC:
@gonzalezma Thanks for checking the text! Yes, it would be great if you could prepare a better description of the NDTSF.

MBartkowiakSTFC:
@gonzalezma Based on the MDANSE code, the c_i, N_i and their square roots really are used the way you described. I have not found discrepancies between your text and the code.

Doc/pages/scattering.rst
sanghamitra-mukhopadhyay, line 1: Should we put the equations ntdsf-eq6 and ntdsf-eq7 in the _analysis-dcsf and _analysis-disf sections? Also can we comment about the units of outputs obtained from _analysis-dcsf and _analysis-disf  sections as we have discussed about units in the _analysis-ntdsf section?

sanghamitra-mukhopadhyay (commented):
I have left a little comment in the scattering section. Thank you @MBartkowiakSTFC and @gonzalezma

Doc/pages/scattering.rst
gonzalezma: I am not sure. I put those equations there just to clarify the differences, but they show only the usual weighting normalization that is applied in all the analysis and that I suppose is explained in the weights section? Note also that here I wrote the weights as b_coh or b_inc^2 because those are imposed in NDTSF, but in principle in the DISF and DCSF analysis one could use other weights. In most cases, they will not have much physical sense, but at least for IXS it will be more reasonable to choose the atomic numbers in the DCSF analysis. As for the units, the units in the numerator and denominator will cancel, so the output will be unitless (strictly speaking it will have the units of the partial I(Q,t) or S(Q,w), i.e. unitless or energy^-1).

sanghamitra-mukhopadhyay: I understand. In that case can we refer these equations at the end of DCSF and DISF documentation to encourage users to visit this NTDSF section for more information? Also I think we should mention the default unit of each output. Although there is a Unit section in the documentation, it is not clear what is the unit of the output users getting from their calculations.

MBartkowiakSTFC:
@sanghamitra-mukhopadhyay @gonzalezma Thank you both for the comments. I linked the DISF/DCSF sections to the equations in the NTDSF description.

At the same time I changed the description of the Q-shells input significantly, and I think that now it is much closer to what really happens. I also made a small change to the code at the same time, since I was not happy that, when I specified a Q-shell range from 1 to 3 in steps of 0.3, the largest Q value ended up being 3.1.

Src/Framework/Configurators/RangeConfigurator.py
sanghamitra-mukhopadhyay, line 1: Should we put a logical check on upper limit, so that when any value falls outside the limit, that will be discarded?

PR #195, opened 2023-04-05
Merging develop into release-next in preparation to release 1.5.2

sanghamitra-mukhopadhyay:


MBartkowiakSTFC:
Apart from the Ubuntu 18.04 build job (which is not supported by GitHub anymore and will not run), all the other workflows are failing on
`git clone https://code.ill.fr/scientific-software/scientific-python.git`
which ends with error code 510.

@eurydice76 Has anything changed in the ScientifcPython repository? It still worked in March.

eurydice76:
There were troubles with the ILL code repo this morning. This may explain the failing clone. Otherwise I did push something on scientific and mmtk two days ago but on separate branches so there should not be any effect. Please tell me if the situation is OK now.

MBartkowiakSTFC:
@eurydice76 So far it is still not working. The full output we get is:
```
Cloning into 'scientific-python'...
error: RPC failed; HTTP 510 curl 22 The requested URL returned error: 510
fatal: expected flush after ref listing
Error: Process completed with exit code 128.
```
What do you think could the problem be?

PR #222, opened 2023-09-06
Protos cleanup

MBartkowiakSTFC:
All the changes in this PR fall into one of the two categories:
1. We remove references to a library which we do not use anymore (netCDF4)
2. We remove code that apparently is not used anywhere within MDANSE.


MBartkowiakSTFC:
While this was a good idea in general, where this PR went too far was removing the Logger. This cleanup will have to be prepared again, starting from the current protos branch, and a bit more carefully this time.

PR #229, opened 2023-10-20
Protos alternative inputs

MBartkowiakSTFC:
Remaining references to netCDF4 are removed from the code.
Also, the DistributedComputing section is removed, since it relies on Pyro throughout.


MBartkowiakSTFC:
This PR will need re-working. Closed for now.

PR #246, opened 2023-11-16
Epic 1 changes made to Ijob and commandline

ashmeigh:
User Story 1: Portability of Trajectory and Job Files
Obj: Enable users to transfer their molecular dynamics (MD) trajectory and job files to another computer without needing to adjust the settings.
Imp: The format of the trajectory and job files should be universally readable and should not depend on system-specific paths or configurations.ranslated into the JSON file format


MBartkowiakSTFC:
This will have to be implemented differently at a later stage instead.

Issue #254, opened 2023-12-08
[BUG] Saving a script fails for Scattering Factor dialogs


MBartkowiakSTFC:
**Description of the error**
The changes to the input widgets and the configurators made it impossible to save an analysis run as a script.

**Describe the expected result**
The "save" button in an analysis dialog should produce a working Python script with the same inputs as those in the GUI

**Describe the actual result**
The GUI crashed with a TypeError

**Suggested fix**
Either the "configurators" or the save method itself have to be modified to get the correct parameters for the script.

**Additional details**
N/A

PR #260, opened 2023-12-12
Add missing valueType in QVector settings

MBartkowiakSTFC:
**Description of work**
This PR will fix issue https://github.com/ISISNeutronMuon/MDANSE/issues/254
For some combinations of inputs it is currently impossible to save the analysis Python script from the GUI.

**Fixes**
For "range" and "vector" variables in the QVector settings it is now necessary to have the 'valueType' entry in their dictionary. This was missing in 3 files, and has been added now.

**To test**
Run the GUI and try to save the script for different types of neutron scattering factors and different types of Q vector generators.


ChiCheng45 (approved):
Looks good.

PR #273, opened 2024-01-06
Protos remove registry

MBartkowiakSTFC:
**Description of work**
This PR removes the REGISTRY from the MDANSE code. Main arguments for the change:
- improved clarity of the code, as the import statements in the beginning of each file now convey information about what other classes each file depends on,
- standard Python approach of using a metaclass and the `__init_subclass__` method should be easier to follow for new team members,
- a number of inconsistencies in file and class naming could be identified and removed in the process. 

**Fixes**
Instead of a central registry, a `SubclassFactory` metaclass has been added to the code. Each interface from the Framework uses this metaclass and becomes a factory object which can create instances of a requested class using the 'create()` method.

Additionally, the text string needed to create a class is the actual class name.

Many parts of the code had to be modified to account for the change, and there is still some work needed to ensure that the GUI is also up to date with the changes.

**To test**
The Analysis unit tests have been shown to work after the changes. However, the changes to the code in this PR are deep-reaching to the extent that all the functionalities of MDANSE need to be tested.


MBartkowiakSTFC:
@eurydice76 While some work is still needed before all parts of the code work correctly, you are welcome to have a look at the Src/Core/SubclassFactory.py file. So far my plan is to replace the central registry with several individual registries, each created using the SubclassFactory metaclass.

eurydice76 (approved):
OK

PR #278, opened 2024-01-18
MDANSE Trajectory and analysis extension update

### LINKED ISSUE #221: HDF5 files: difference between trajectories and analysis results
At the moment both the converted trajectories and the analysis results are HDF5 files with the .h5 extension. In the long run we should make it easier to tell the difference between the two file types.

ChiCheng45:
**Description of work**
Updated MDANSE so the trajectory HDF files now have the extension .mdt and analysis hdf files now have the extension .mda. Updated the GUI so that the default filter is set to *.mdt. Added sandbox to gitignore.

**To test**
Run MDANSE unit test to ensure passing. Run conversion and analysis scripts to check the correct extensions are written. Test with the GUI.

image

image

closes #221


MBartkowiakSTFC (commented):
There are 6 analysis types (in Jobs) which output a trajectory and not an analysis result. I have changed the output format to MDTFormat for those. Please check if there are any tests where those jobs run, and if the output format there needs to be changed too.
(they are the files from commit: https://github.com/ISISNeutronMuon/MDANSE/pull/278/commits/60678d2d973d0309e80368d34033cab4837c5fe2 )

Other than that it passes tests and looks good. Thanks!

ChiCheng45:
> There are 6 analysis types (in Jobs) which output a trajectory and not an analysis result. I have changed the output format to MDTFormat for those. Please check if there are any tests where those jobs run, and if the output format there needs to be changed too. (they are the files from commit: [60678d2](https://github.com/ISISNeutronMuon/MDANSE/commit/60678d2d973d0309e80368d34033cab4837c5fe2) )
> 
> Other than that it passes tests and looks good. Thanks!

I've checked the tests and it looks good. I've tested with the GUI and its works fine now too.

PR #284, opened 2024-01-23
Fix ase converter

MBartkowiakSTFC:
**Description of work**
The ASE converter is now able to take an ASE .traj file and convert it to the .MDT format.

**Fixes**
- changed the ASE converter code
- added a sample trajectory to the tests Data directory
- added a corresponding test to the test_converters.py file

**To test**
Please confirm that the tests pass. Optionally try to convert and load the sample trajectory in the GUI.


MBartkowiakSTFC:
I added extra functionality to the ASE converter. Now it also works for the .xyz files.

It seems that the GitHub Actions that failed did it quite randomly. We can always re-run them later to be sure.

ChiCheng45 (approved):
Look good. I've run the ASE conversions with both the GUI and the scripts and its working. Unit test pass for me as well.

PR #288, opened 2024-01-26
Artificial trajectory generator

MBartkowiakSTFC:
**Description of work**
Extends the code to allow Analysis runs to use other input than an HDF5 file.

**Fixes**
- added MockTrajectory
- implemented basic tests of MockTrajectory
- added an MDMCTrajectoryConfigurator, which allows MDMC input using a wrapper on MDMC side

**To test**
Unit tests should all pass.
It is impossible to use MockTrajectory from the GUI at this stage, but it can be used in a script.


ChiCheng45 (approved):
Looks good this should be really useful and all tests pass on my end.

PR #289, opened 2024-01-26
Code cleanup

MBartkowiakSTFC:
**Description of work**
Mainly removing unnecessary fragments of code.

**Fixes**
- deleted the old GUI code (wxWidgets part),
- deleted the SVG file format,
- deleted ComplexNumber and the matching configurator,
- moved the GUI code from MDANSE_GUI/PyQtGUI to MDANSE_GUI

**To test**
If the deleted code was correctly identified as unused, there should be no change to the way MDANSE works.
Unit tests should all pass.


ChiCheng45 (approved):
Looks good, all unittest pass on my end and GUI loads up without issues.

PR #290, opened 2024-01-30
Improve molecule finder

MBartkowiakSTFC:
**Description of work**
MoleculeFinder has been modified to include a tolerance parameter. Also, output trajectories are now handled by OutputTrajectoryConfigurator.

**Fixes**
- MoleculeFinder has a "tolerance" input,
- MoleculeFinder uses Trajectory the same way as other Analysis runs,
- OutputTrajectoryConfigurator can determine floating point number format of the trajectory, and type of compression used.
- test_converters.py has been modified to use the new parameters of OutputTrajectoryConfigurator.
- There is no input field for output trajectory format, since we have only one format.

**To test**
All tests should pass. Also, run a converter in the GUI, and run MoleculeFinder in the GUI to make sure the widgets are operating correctly.
In the GUI only the gzip compression is enabled at the moment. lzf crashed on my computer with "Bus error: 10", even though it works in the tests. I am guessing that lzf can fail for some sizes of np.float other than np.float64.


ChiCheng45 (approved):
Looks good, tests all pass and converter in the GUI works.

PR #291, opened 2024-01-30
atom selection/filtering backend refactor

ChiCheng45:
**Description of work**
The way that atom selection are done has now been changed. Previously a complicated parsing system was implemented which selected molecules based on a string which included set operations in it. I have simplified the approach so that it simply applies specific filters to remove specific atoms. This also makes the setting of the atom filtering more human json string e.g. `'{"water": true, "elements": {"S": true}}'` which will remove water molecules and sulfur atoms. 

FilterSelection settings can be updated with the `update_settings` method e.g.

```
filter = FilterSelection(protein_chemical_system)
filter.update_settings({"elements": {"S": True}, "water": True})
atm_idxs = filter.get_idxs()
```

setting can be dumped to a json string with `settings_to_json`

```
filter = FilterSelection(protein_chemical_system)
filter.update_settings({"elements": {"S": True}, "water": True})
json_dump = filter.settings_to_json()  # '{"water": true, "elements": {"S": true}}'
```

and updated with `update_from_json`.

For molecule or molecular group filtering it now uses smarts pattern matching. This will make it much easier if we want to add matches for larger molecular groups or molecules. See the `select_water` function.

```
def select_water(system: ChemicalSystem) -> set[int]:
    ...
    patterns = ["[#8X2;H2](~[H])~[H]"]
    return system.get_substructure_matches(patterns)
```

**To test**
Run units tests to ensure passing, test GUI loads up without issues.


MBartkowiakSTFC (approved):
Thank you for the changes to the code. A definite improvement to the way we handle trajectory information.

Issue #210, opened 2023-06-29
Unable to import AtomSelectionParser


MBartkowiakSTFC:
If I try to import AtomSelectionParser
`from MDANSE.Framework.AtomSelectionParser import AtomSelectionParser`
I immediately get an error:
```
Traceback (most recent call last):
  File "/Users/maciej.bartkowiak/science/fake-data-for-MDMC/new_mdanse/analysis_script_python3.py", line 12, in <module>
    from MDANSE.Framework.AtomSelectionParser import AtomSelectionParser
  File "/Users/maciej.bartkowiak/python/builder/lib/python3.11/site-packages/MDANSE/Framework/AtomSelectionParser.py", line 20, in <module>
    from MDANSE.Externals.pyparsing.pyparsing import delimitedList, oneOf, opAssoc, operatorPrecedence, printables, Forward, OneOrMore, Optional, Word
  File "/Users/maciej.bartkowiak/python/builder/lib/python3.11/site-packages/MDANSE/Externals/pyparsing/pyparsing.py", line 110, in <module>
    range = xrange
            ^^^^^^
NameError: name 'xrange' is not defined. Did you mean: 'range'?
```

The code included in Externals/pyparsing is most likely seriously outdated. The main question is whether to try to fix it, or just to get rid of it altogether. I am hoping that pyparsing was included in the code because some feature was missing in Python 2 at the time, and we can replace it with something from the standard library now.

MBartkowiakSTFC:
It looks like the `Externals.pyparsing` can be replaced by the up-to-date `pyparsing` module from pypi. The structure of the modules has changed a little, but it will not require too complex changes.

MBartkowiakSTFC:
While it was possible to get pyparsing to work in the protos branch, it would be the best to come up with a more stable method of selecting atoms.

The criteria are:
1. Inputs should be general enough to be transferred from one trajectory file to another (e.g. "Pick all oxygen atoms", "Pick all water molecules"),
2. There should be operators combining the selections (e.g. "Pick all oxygen atoms that are NOT in water molecules"),
3. The result should be a list of atom indices.
4. The default selection should be all the atoms in the trajectory.

PR #293, opened 2024-02-06
Updated the atom selector and added the front end atom selection helper

ChiCheng45:
**Description of work**
Changed the atom selection mechanism so that it is now a selector and added a setting to invert the selection so a filter can still be done. Added a atom selection front end so that the setting can be created using the GUI.

image

**To test**
Unit tests should pass.

In the GUI load a trajectory and open a analysis window with atom selection such as the position autocorrelation function and check that:
- the atom selection helper button opens the helper window
- clicking the atom selection helper button should close the window
- helper window should open next to the analysis window
- options in the atom selection helper should be greyed out if those group/molecules/elements are not seen in the chosen trajectory
- select all in the combobox should behave in the expected way e.g. unchecking select all deselects all options.
- the textbox should update correctly after selecting different selections
- clicking apply should update the atom_selection field in the analysis window
- editing the atom_selection value in the analysis window should change black/red on a correct/incorrect setting

After applying the atom selection setting run an analysis and check that the atom selection works e.g. remove a specific oxygen and there are no oxygen atoms in you results.


MBartkowiakSTFC:
One observation so far: on my computer, I am unable to leave the combo boxes in the atom selection widget. I can open them and make a selection, but the rest of the widget becomes inactive and the only option is to close the whole atom selection.

ChiCheng45:
> One observation so far: on my computer, I am unable to leave the combo boxes in the atom selection widget. I can open them and make a selection, but the rest of the widget becomes inactive and the only option is to close the whole atom selection.

Not sure what the issue could be. It works fine on my windows machine. Could it be down to the version of QT that is being used? I'm on pyqt6 6.6.1 with qtpy 2.4.1. I've added a gif of the expected behaviour.


Animation

MBartkowiakSTFC:
It turns out that I am able to close the drop-down part of the combo box after all. It only closes when I click on the **parent** dialog, and not on the helper dialog. So, to make the GUI active again, I have to click on the PositionAutoCorrelationFunction dialog, and not on the Atom Selection Helper dialog.

An issue with the parent-child hierarchy?

MBartkowiakSTFC:
My changes to the order of children and parents in the GUI did not help.

A brief investigation shows that Qt has a history of problems with combo boxes in dialogs:
https://forum.qt.io/topic/123101/quick-controls-1-combobox-items-unselectable-on-macos
The link above shows a problem occurring in Qt 5, using QtQuick, which should be quite different from our setup - and yet a combo box in a dialog still would not work correctly on a Mac (and just fine on other platforms)

MBartkowiakSTFC (approved):
While the GUI may still be optimised, it is already usable in its current form. Overall, the new atom selector is a major improvement over the previous mechanism.

ChiCheng45:
Fixed #210

PR #298, opened 2024-02-13
atom selection performance fix

ChiCheng45:
**Description of work**
With a large number of atoms the it can take a while to load up the actions dialog. This is because of the atom selections checkable combobox. I've made some updates to improve its performance.

**Fixes**
Long action dialog loading times (1-2 mins).

**To test**
Load up a trajectory with a lot of atoms. (convert the Cu_10steps_ASEformat.traj) and load up and open an action dialog which has an atom selection section e.g. positionautocorrelationfunction. This should only take a few seconds now. Check that the atom selection helper behaves correctly.


MBartkowiakSTFC (approved):
I tried the Cu trajectory with 4000 atoms. The dialog got created within a second. Everything worked as expected.

PR #305, opened 2024-02-14
Fix gui bugs

### LINKED ISSUE #300: [BUG] MDANSE_GUI crashes when changing options in the trajectory converter
**Description of the error**
MDANSE crashes when changing specific options e.g. change to empty box or a dot (.). Most of these a related to some type conversion issue in the floatwidget.

```
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\InputWidgets\FloatWidget.py", line 71, in newText
    self._value = float(text)
                  ^^^^^^^^^^^
ValueError: could not convert string to float: 'e'

```

The issue affects the following options in the trajectory converters:

- ASE time step 
- ASE number of time step 
- DCD time step
- LAMMPS mass 
- LAMMPs time step
- LAMMPS number of time step
- VASP time step
- CHARMM time step
- NAMD time step
- XPLOR time step

MBartkowiakSTFC:
**Description of work**
Corrects the code of FloatWidget, IntegerWidget and FramesWidget to improve their error handling.
closes #300 

**Fixes**
FloatWidget and IntegerWidget now use the standard value passing mechanism of WidgetBase.
FramesWidget uses the number of frames as the starting value, and becomes invalid if step=0.
It is impossible to run an analysis if one of the inputs is invalid.

**To test**
Try putting all kinds of bad input into number fields.


ChiCheng45:
This doesn't crash anymore but its still possible to use an empty box for e.g. the time step option which will then cause a crash when running. I think we should also fix this before merging.

Animation2

```
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\Widgets\ActionDialog.py", line 251, in execute_converter
    pardict = self.set_parameters()
              ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\Widgets\ActionDialog.py", line 246, in set_parameters
    results[key] = self._widgets[widnum].get_widget_value()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\InputWidgets\FloatWidget.py", line 67, in get_widget_value
    return float(self._field.text())
           ^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: ''
```

MBartkowiakSTFC:
I have gone through the input widgets and added code to make them show placeholder text whenever they are empty. The current situation is:
- The initial value of each widget with QLineEdit is take to be the default value,
- if the widget is empty, it still shows the placeholder text of the default value,
- if you run the analysis with an empty field, it will take the placeholder text as input.
This way there should be no way to run a converter or analysis with an empty input field.

ChiCheng45:
I found some more issues. I can place an "e" into the time step widget in the converter which causes the conversion to fail. 

image

Some other examples include ".e", "e1", "1.e", "1.1e" and "e1".

MBartkowiakSTFC:
Yes, this should have been marked as invalid. I will let you know once this is fixed.

ChiCheng45 (approved):
Great #300 seems to be all fixed now. I made a small change to the colour of the text for the atom selection widget since for when it uses the default with an empty input box. Also the adjustable left and right layout works great!

Issue #316, opened 2024-02-14
[BUG] Changing the atom size scaling causes crash with no trajectory loaded.


ChiCheng45:
**Description of the error**
Load the GUI, move to the 3D view tab and change the atom size scaling. MDANSE will then crash with the following error in the logs.

```
Traceback (most recent call last):
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\MolecularViewer\Controls.py", line 263, in setAtomSize
    self._viewer._new_scaling(new_value)
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\MolecularViewer\MolecularViewer.py", line 152, in _new_scaling
    self.update_renderer()
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\MolecularViewer\MolecularViewer.py", line 637, in update_renderer
    self._actors = self.build_scene()
                   ^^^^^^^^^^^^^^^^^^
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\MolecularViewer\MolecularViewer.py", line 282, in build_scene
    sphere_mapper.SetScalarRange(self._polydata.GetScalarRange())
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'GetScalarRange'
```

**Describe the expected result**
Should be able to change the value. Then when a trajectory is loaded MDANSE should used that value.

ChiCheng45:
Looks like this got fixed in #347.

PR #322, opened 2024-02-19
3d view tab update

### LINKED ISSUE #303: [BUG] MDANSE_GUI 3D View does not always switch to another trajectory
**Description of the error**
The trajectory view gets updated on loading a new trajectory, but it does not update when switching to another trajectory already on the list.

The view should change every time the user selects a different trajectory.

**Suggested fix**
A Qt Signal is probably not connected.

**Additional details**
N/A

### LINKED ISSUE #307: [BUG] Loading a trajectory from the 3D viewer tab causes crashes later on when loading analysis
**Description of the error**
When loading up a trajectory from the 3D viewer tab and then opening an analysis window, MDANSE crashes.

**Additional details**
Below I show it working when you load a MDT file from the trajectory tab and then a crash when you load it from the 3D View tab.
Animation3

The following error logs are seen.

```
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\Tabs\JobTab.py", line 86, in show_action_dialog
    dialog_instance = dialog(
                      ^^^^^^^
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\Widgets\ActionDialog.py", line 93, in __init__
    path, filename = os.path.split(self._input_trajectory)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen ntpath>", line 213, in split
TypeError: expected str, bytes or os.PathLike object, not HDFTrajectoryInputData
```

ChiCheng45:
**Description of work**
Combined the trajectory and the 3d view tabs in one triple panel tab.

Animation6

**Fixes**
Fixes #307, Fixes #303
Fixed so that the atom properties table updates correctly on trajectory change.

**To test**
1. Load a trajectory and select the trajectory. Check that the atom properties and information box updates correctly.
2. Load another trajectory and select it. Check that everything updates correctly.
3. Check that the play animations works.
4. Check the actions tab and the trajectories are there.
5. Run a analysis job and check that it worked properly.
6. Delete the trajectories in the trajectories tab and check everything updates correctly.


MBartkowiakSTFC (approved):
I have tested the new layout and tried to make the GUI crash. It works well, and I think that it is an improvement from the point of view of overall functionality of the GUI.
As far as I can tell, both issues #303 and #307 will be closed when we merge this. Thank you!

For the record, the following problems still exist:
- GUI crashes if I change the 3D View plotting parameters **before** loading a trajectory (issue #316)
- GUI hangs if I repeatedly try to delete a nonexistent object from the trajectory list.
Those two problems will have to be addressed sooner or later.

PR #325, opened 2024-02-20
Add the enabled property to configurable and test for it

### LINKED ISSUE #324: [ENHANCEMENT] Disable/enable jobs in the GUI
We need an easy mechanism of deciding which analysis and converter types should be available to the users. This will make it easier to control the software when we release a new version.

Either flags in the class definition (disabled = True), or an external list of things to disable.

N/A

**Additional context**
ImprovedASE converter may not be ready to be used, so we could start by disabling that one.

MBartkowiakSTFC:
**Description of work**
Every class derived from Configurable now has the property `enabled = True`. JobTree model only shows the classes for which enabled is True.
closes #324

**Fixes**
ImprovedASE converter is now disabled.

**To test**
Start the `mdanse_gui` and confirm that ImprovedASE does not appear in the Converter tab, but all the other options are still there.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurable.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurable.py
@@ -40,6 +40,8 @@ class Configurable(object):
             #.. 2-value is the dictionary of the keywords used when initializing the configurator.  
     """
 
+    enabled = True
+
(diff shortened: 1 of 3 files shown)


ChiCheng45 (approved):
Looks good.

PR #326, opened 2024-02-21
Trajectory and job tab right click fix

### LINKED ISSUE #308: [BUG] Right click delete can cause unexpected behaviour or a crash
**Description of the error**
When you right click and delete things when there is nothing to delete some unexpected behaviour occurs.

When there is nothing to delete it should either do nothing or the delete should not be available

**Additional details**

Animation4

ChiCheng45:
**Description of work**
Changed so that the right click is blocked when the mouse is not over an item. Adjusted the textinfo so that the some initial information is shown. The jobs table clears when there are when they are all deleted.

**Fixes**
Fixes #308 

**To test**
- The right click on the trajectory and running jobs tab should be blocked.
- After loading a trajectory or running a job the right click on the trajectory or job items should allow for a delete.
- Delete should work as expected.
- When there is nothing else to delete the right click should be blocked again.
- When all jobs are deleted in the jobs tab the table should be cleared.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/RunTab.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/RunTab.py
@@ -49,7 +49,7 @@ class RunTab(GeneralTab):
             model=JobHolder(),
             view=RunTable(),
             visualiser=TextInfo(
-                header="MDANSE Analysis Output",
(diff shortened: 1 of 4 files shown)


MBartkowiakSTFC (approved):
The context menu appears correctly when clicking on an item, and does not appear otherwise. Everything works as expected.

PR #327, opened 2024-02-23
DL_POLY convertor fix

### LINKED ISSUE #310: [BUG] DL_POLY trajectory conversion failure with files from MDANSE-Examples
**Description of the error**
MDANSE fails to convert the DL_POLY trajectories from MDANSE-Examples with all file pairs. The following errors is seen.

```
ic| err: ('<class '
          "'MDANSE.Framework.Jobs.IJob.JobError'>(<MDANSE.Framework.Converters.DL_POLY.DL_POLY "
          "object at 0x0000027A54C49D90>, 'Traceback (most recent call last):\
           File "
          '"C:\\\\Users\\\\xcb63893\\\\PycharmProjects\\\\MDANSE\\\\MDANSE\\\\Src\\\\MDANSE\\\\Framework\\\\Jobs\\\\IJob.py", '
          'line 302, in run\
             self.initialize()\
           File '
          '"C:\\\\Users\\\\xcb63893\\\\PycharmProjects\\\\MDANSE\\\\MDANSE\\\\Src\\\\MDANSE\\\\Framework\\\\Converters\\\\DL_POLY.py", '
          'line 424, in initialize\
             '
          'self.configuration["output_files"]["files"][0],\
             '
          '~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^\
         KeyError: '
          "\\'output_files\\'\
         ')Traceback (most recent call last):
         "
          '  File '
          '"C:\\Users\\xcb63893\\PycharmProjects\\MDANSE\\MDANSE\\Src\\MDANSE\\Framework\\Jobs\\IJob.py", '
          'line 302, in run
         '
          '    self.initialize()
         '
          '  File '
          '"C:\\Users\\xcb63893\\PycharmProjects\\MDANSE\\MDANSE\\Src\\MDANSE\\Framework\\Converters\\DL_POLY.py", '
          'line 424, in initialize
         '
          '    self.configuration["output_files"]["files"][0],
         '
          '    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
         '
          "KeyError: 'output_files'
         ")
```

ChiCheng45:
**Description of work**
DL_POLY conversion fix and added automatic determination of DL POLY version. Fixed DLPOLY file file dialog.

**Fixes**
Fixes #310 

**To test**
- Convert DL_POLY files for different DL POLY versions. They should convert successfully and MDT files should load up without issues.


MBartkowiakSTFC (commented):
Thank you for the changes. The DL_POLY converter works for all the trajectories from MDANSE-Examples.

However, when I try it out on other DL_POLY trajectories which we received recently, there are trajectories for which it still fails. This is not always a case of the file being read wrong - there is a TIP4P water trajectory in which water is composed of 4 particles, H, H, O and Q. MDANSE does not know what to do with the last one, since it is not a chemical element, but a localised charge in space.

Overall, I think we will have to separate the converters into the frontend and backend parts. DL_POLY conversion should follow the steps:
- try to determine the file version
- try version 2
- try version 3
- try version 4
- try ASE
- fail

gonzalezma:
This is not new. Names given by users in the FIELD file are arbitrary. This is why the atom aliases box was introduced , so the user can help the converter by indicating the atom element, e.g.: {"Q":"Du"} for the TIP4P case to indicate that Q refers to a dummy atom.

ChiCheng45:
> Overall, I think we will have to separate the converters into the frontend and backend parts. DL_POLY conversion should follow the steps:
> 
> * try to determine the file version
> * try version 2
> * try version 3
> * try version 4
> * try ASE
> * fail

I think we can just determine the version from the history file. I don't see any reason against doing this as it looks like there enough differences in record 1 and 2 in the history file between version 2, 3 and 4 to do this. See the DLPOLY 2, 3 and 4 manuals.

MBartkowiakSTFC (approved):
Excellent! Personally, I would have to look into the files myself to find out which file format they are, so it is better to have the code check it for us.

Except for the TIP4P trajectories, everything worked for me.

PR #338, opened 2024-02-28
Atom colour update

ChiCheng45:
**Description of work**
The atom colours in MDANSE were not standard e.g. carbon is green when most people with a chemistry background would expect it to be black.

I've used similar colours to what is used in the vtk molecule map which appears to use more standard colours. This also mean that more colours are specified now e.g. gold.

**Before**
image
**After**
image


MBartkowiakSTFC (approved):
Looks good. Happy to have green strontium again.

PR #345, opened 2024-03-07
Improved labels, corrected text output

MBartkowiakSTFC:
**Description of work**
This PR mainly deals with labels, docstrings and naming in MDANSE. Additionally, problems with analysis output in text format have been corrected so that text output works again.

**Fixes**
1. Text output works again,
2. Files in text output have correct creation time set,
3. ASCIIFormat is now TextFormat,
4. Renamed 'monoprocessor' and 'multiprocessor' to (more accurate) 'single-core' and 'multicore'.
5. Added a docstring to RigidBodyTrajectory,
6. Activated tooltips in the GUI for several widgets which were missing them before.

**To test**
1. Load a trajectory,
2. Check if all the action dialogs are created correctly. Look at the widget tooltips in some of the dialogs,
3. Run an analysis using TextFormat as output,
4. Check if the files in the resulting .tar archive are created correctly and with today's timestamp.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/AseInputFileConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/AseInputFileConfigurator.py
@@ -102,4 +102,4 @@ class AseInputFileConfigurator(InputFileConfigurator):
         except KeyError:
             print(f"No VALUE in {self._name}")
 
-        return "Input file: %r" % self["value"]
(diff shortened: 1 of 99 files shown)


ChiCheng45 (approved):
Looks good and it works on my end. I added a commit so that there are new lines in some of the get information methods of the configurators so that the jobinfo.txt file is formatted nicely.

PR #346, opened 2024-03-08
Trajectory tab and 3D view updates.

### LINKED ISSUE #323: [BUG] Trajectories with many atoms take too long to load
**Description of the error**
While there are no problems loading trajectories with many frames, having many atoms in a frame is problematic. On a modern computer is takes 10 minutes to load a trajectory with 256000 atoms.

This should have taken a few seconds.

I took 10 minutes to load the trajectory, then another 10 to put it into the 3D viewer. After that, moving to the next frame in the 3D viewer would take 2-3 seconds per frame.

**Suggested fix**
I expect that the problem is caused by the ChemicalSystem instance which creates an instance of Atom for every atom in the trajectory. This will have to be removed.

**Additional details**
There may be some overhead of creating a Configuration out of atom coordinates. This should be investigated once the Atom class has been removed.

ChiCheng45:
**Description of work**
- Improved the trajectory loading performance. 42,592 atoms can be loaded in a few seconds.
- Added bonding to 3D view.
- Added unit cell to 3D view.
- Added method for faster access to atom database properties.
- Updated trajectory panel so that they are not reloaded on trajectory selection change.
- Optimized the build connectivity in MDANSE graph.

Animation12

**To test**
Load a few trajectories up in the trajectory tab and check 3D view is working correctly.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalEntity.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalEntity.py
@@ -182,7 +182,10 @@ class _ChemicalEntity(metaclass=abc.ABCMeta):
         :rtype: numpy.ndarray
         """
         coords = configuration["coordinates"]
-        masses = [ATOMS_DATABASE.get_atom_property(at.symbol, "atomic_weight") for at in self.atom_list]
+        masses = [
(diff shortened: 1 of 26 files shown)


MBartkowiakSTFC (approved):
Everything I tried worked fine. It definitely works faster than the previous version.

PR #347, opened 2024-03-08
Atom Properties speed-up in 3D view

MBartkowiakSTFC:
**Description of work**
The atom property editor in the 3D view tab has been modified, and now it groups atoms of the same type into single entries. This allows the user to change the colour or size of all the atoms of the same type at once.

At the moment it is NOT possible to edit the properties of individual atoms anymore.

Also, at the same time check boxes have been added which enable or disable different parts of the 3D display (atoms, bonds, axes and cell).

**Fixes**
- vtk arrays are now set using numpy_to_vtk helper functions, and not in a `for` loop.
- there is only one entry per chemical element in the atom property editor.
- bonds can be switched off (e.g. if detecting them takes too long).

**To test**
1. Try toggling the checkboxes in 3D view both before and after selecting a trajectory.
2. Edit the atom properties for different atom types and check if the display gets updated each time.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/AtomProperties.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/AtomProperties.py
@@ -12,17 +12,20 @@
 #
 # **************************************************************************
 
+from typing import List
+
 import numpy as np
+from vtk.util.numpy_support import numpy_to_vtk
 import vtk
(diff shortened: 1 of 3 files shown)


ChiCheng45:
All work on my end but can we block the editing of the element name in the atom properties?

image

MBartkowiakSTFC:
The atom name editing is now disabled.

ChiCheng45 (approved):

ChiCheng45:
> The atom name editing is now disabled.

Thanks looks good!

PR #349, opened 2024-03-12
updated so that bonding to dummy atoms are removed in the 3D viewer

ChiCheng45:
**Description of work**
Although the covalent radii of dummy atoms are zero the bonding determination in the 3d viewer may still bond them if they are close enough to other atoms. I have updated the bonding determination so that bond to dummy atoms are never made.

**To test**
Convert a trajectory with dummy atoms e.g. the water_with_velocities from MDANE-Examples DL_POLY. Load the trajectory and uncheck the atoms in the visible objects group. There should not be a bond between the oxygen and dummy atoms.

Before
image

After
image

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -594,12 +594,6 @@ class MolecularViewer(QtWidgets.QWidget):
 
         # update the atoms
         coords = self._reader.read_frame(self._current_frame)
-        cov_radii = np.array(
-            [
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC:
Just a question: do you think that there would be any advantage to checking the atom properties only once (that is, getting the covalent radii and NOT the atoms types) and then using only those with a non-zero covalent radius? It may be that, since you improved the performance of accessing the atoms properties, it does not really matter from the point of view of execution time. It _might_ matter for people who like to define their own atom types and would like to add new dummy types - assuming that such users exist.

MBartkowiakSTFC (approved):
Thank you for the changes! Looks good.

PR #354, opened 2024-03-14
Action tab analysis loading performance update

ChiCheng45:
**Description of work**
Some performance updates to the action tab analysis job loading. Made optimisations to the atom selection object and its widget.  

Previously when you selected an analysis job it would reload the trajectory four times! This has been updated so that it only reload it once. We have to reload it once so that it configures the HDF configurators which other configurators depend on.

**To test**
Load up a few trajectories and run a analysis job.


MBartkowiakSTFC (approved):
Looks good, there is a visible improvement in performance.

PR #356, opened 2024-03-14
Atom selection dummy update

### LINKED ISSUE #352: [BUG] Remove dummy atoms during conversion?
**Description of the error**
During the trajectory conversion process dummy atoms are carried on though the MDANSE workflow. So that we get some odd cases where the total DOS includes the DOS of the dummy atoms. 

H2O DOS 
image

H2O DOS (H change to dummy)
image

**Suggested fix**
Remove the dummy atoms on trajectory conversion. The user can map the dummy atoms to other atoms with the atom mapping. Alternatively make sure that the dummy atoms are not included in the analysis jobs.

EDIT: I just redid the VACF and it dummy atoms are included in the total VACF so it is consistent with the above.

ChiCheng45:
**Description of work**
Changed the all selector so that it does not include dummy atoms. Added a selector for dummy atoms. The default is now changed so that all atoms except dummy atoms are selected.

image

**Fixes**
Closes #352.

**To test**
1. Load a trajectory with dummy atoms. 
2. Using the atom selection run a analysis with and without dummy atoms. 
3. Load up the MDA file and check that dummy are or are not included.
4. Load a trajectory without dummy atoms.
5. The option to select dummy atoms in the atom selection helper should be greyed out.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/AtomSelector/all_selector.py
+++ b/MDANSE/Src/MDANSE/Framework/AtomSelector/all_selector.py
@@ -5,7 +5,8 @@ from MDANSE.Chemistry.ChemicalEntity import ChemicalSystem
 def select_all(
     system: ChemicalSystem, check_exists: bool = False
 ) -> Union[set[int], bool]:
-    """Selects all atoms in the chemical system.
+    """Selects all atoms in the chemical system except for the dummy
+    atoms.
 
     Parameters
     ----------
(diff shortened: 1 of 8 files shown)


MBartkowiakSTFC (approved):
I tried this version out on trajectories with and without dummy atoms, and ran an analysis for each case (dummy atoms included, dummy atoms excluded). I did not run into any problems.

PR #357, opened 2024-03-14
Python 3.12 fix and removed NetCDF dependency

ChiCheng45:
**Description of work**
imp has been removed in python 3.12. This was used in the class registry. As the class registry is not really used I think we can just remove it. I've also removed the NetCDF dependency and some related code.


MBartkowiakSTFC (approved):
Thanks for the changes. I have not found any problems with it.

PR #366, opened 2024-03-18
Fix ase converter box

### LINKED ISSUE #344: [BUG] ASE converter fails to read the box size from extended XYZ files
**Description of the error**
We converted a trajectory in the extended XYZ format using the ASE converter. It was a system with a unit cell and periodic boundary conditions both defined in the trajectory file. However, the MDANSE trajectory after the conversion does not contain this information. 

The trajectory information should indicate that the system is periodic.

The value for periodic boundary conditions was shown to be False.

**Suggested fix**
1. Manual test of the ASE loader to verify that ASE correctly reads the unit cell information.
2. Changes to the MDANSE code to make sure that the unit cell information is stored.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
The ASE converter has been assigning wrong dimensions to the unit cell of the simulation. This is now fixed.

closes #344 

**Fixes**
1. The unit cell dimensions are saved correctly.
2. Systems without periodic boundary conditions are not given a unit cell.
3. Reading performance has been improved.

**To test**
Apart from the normal unit tests, it should be tested on one ASE trajectory (.traj) like the copper trajectory, and on a trajectory without periodic boundary conditions (probably hexane).

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Converters/ASE.py
+++ b/MDANSE/Src/MDANSE/Framework/Converters/ASE.py
@@ -136,6 +136,9 @@ class ASE(Converter):
         try:
             frame = self._input[index]
         except TypeError:
+            frame = next(self._input)
+        else:
+            print("ASE using the slower way")
             frame = read(self.configuration["trajectory_file"]["value"], index=index)
         time = self._timeaxis[index]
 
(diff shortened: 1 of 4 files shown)


ChiCheng45 (approved):
Looks good I've tested it on a few systems and loads and converts them correctly.

PR #367, opened 2024-03-18
Updated so that the HDFInputFileConfigurator map to InputFileWidget

ChiCheng45:
**Description of work**
Updated so that the HDFInputFileConfigurator map to InputFileWidget.

image

**To test**
Load up the GUI and check that structure factor inputs are file input widgets.


MBartkowiakSTFC (commented):
I added extra code to make sure that the HDFInputFileConfigurator tells the InputFileWidget to look for .mda files.

PR #371, opened 2024-03-19
Add unit cell updating for each frame in 3D Viewer

MBartkowiakSTFC:
**Description of work**
The unit cell in 3D viewer was not changing with the frame number.

**Fixes**
Unit cell is now frame-dependent in the hdf5wrapper.

**To test**
Please animate a trajectory with variable cell dimensions.


ChiCheng45 (approved):
Looks good, I've done a few tests and the unit cell is updated in the 3d view.

PR #379, opened 2024-03-20
Added unitless unit and fixed data not plottable for eccentricity jobs

### LINKED ISSUE #363: [BUG] Eccentricity data not plottable
**Description of the error**
The eccentricity job completes successfully but when the MDA file is loaded up certain results are not plottable. 

**Additional details**
Job was run with the DFTB H2O trajectory from MDANSE-Examples.

MDANSE-2
image

MDANSE-1
image

ChiCheng45:
**Description of work**
Added unitless unit and fixed data not plottable for eccentricity jobs

**Fixes**
Fixes #363

**To test**
Run the eccentricity job and load up plot. All results should be plottable.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
@@ -123,7 +123,7 @@ class Eccentricity(IJob):
                 "LineOutputVariable",
                 npoints,
                 axis="time",
-                units="uma*nm2",
+                units="uma nm2",
             )
         for axis in ["a", "b", "c"]:
             self._outputData.add(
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
Screenshot 2024-03-20 at 14 05 05

I tried it with weights "equal" and "atomic_weight", but it was not quite right. I will try for other trajectories and see if there is a difference, but there will definitely be some cases where it does not work.

MBartkowiakSTFC:
OK, I know what happened: while you defined a 'unitless' unit in the json file, on my computer I already have my own _modified_ unit database. MDANSE checks that the user-level unit database exists, and uses it. It does not check if new units appeared in the _main_ database, and if they should be copied into the user database.

New users will not be affected by this problem, but if we add new units in the future, then the existing users may miss out on it, unless we add a mechanism for synchronising the databases.

MBartkowiakSTFC (approved):
It works, thanks for the changes.

PR #385, opened 2024-03-21
Trajectory conversion updates and unit tests

ChiCheng45:
**Description of work**
Added more trajectory converter unit tests and updated job registration so that all converter jobs are registered.

**Fixes**
Fixed the mdanse command job script generation should work with all trajectory converter jobs. This can stop working with some versions depending on whether the conversion job was registered during the package/module importing.

Updated the DFTB and lammps file wildcards in the GUI.

**To test**
All unit tests should pass. `mdanse.exe --js castep` should work with all trajectory converters, please test a few. Check correct file wildcards for DFTB and lammps converters in the GUI.


MBartkowiakSTFC (approved):
The changes look good, and everything works fine for me.

PR #387, opened 2024-03-21
ROG remove weight setting

### LINKED ISSUE #377: [BUG] Radius of Gyration weight setting
The radius of gyration has the weight setting available but it is not used.

**Suggested fix**
Remove the setting.

ChiCheng45:
**Description of work**
Removed the weight setting from the ROG analysis job since is not used and also does not makes sense for it to be included.

**Fixes**
Fixes #377

**To test**
Open the GUI and check that the weight setting has been removed in the ROG analysis job.


MBartkowiakSTFC (approved):
Good change, everything still works the same.

PR #388, opened 2024-03-21
Coordinate folding change

### LINKED ISSUE #372: [BUG] Unexpected behaviour with fold coordinates into box option
**Description of the error**
When I use fold coordinates into box I expect the atomic coordinates to be located inside the unit cell so that in fractional coordinates it will have a position between 0 to 1 for all three axes. However with the fold coordinates into box option MDANSE moves the atomic coordinates so that it will have a position in fractional coordinates of -0.5 to 0.5 for all three corordinates.

without folding
image

with folding
image

**Suggested fix**
With the fold coordinates into box option selected all atoms should have a position between 0 to 1 in fractional coordinates for all three axes.

ChiCheng45:
**Description of work**
- Change the default fold setting to false for all converters.
- Changed the folding so that the atom coordinates are moved to have a value between 0 and 1 in fractional coordinates.
- Added the fold setting to the ASE and CP2K jobs so that all trajectory converters have this option.

DFTB H2O trajectory converted with the fold setting on.
image

**Fixes**
Fixes #372

**To test**
Check that the fold option is available for the ASE and CP2K conversion jobs in the GUI. Convert a trajectory that does not have all atom position inside the unit cell (DFTB H2O for example) with the fold setting on and check that they get folded into the unit cell in the 3D view.


MBartkowiakSTFC (approved):
I tried a few trajectories using the ASE converter. They all worked correctly:
1. The extended XYZ trajectory ends up in the box, with all the atoms folded as expected,
2. The Gaussian trajectory _without_ a cell works the same if the folding is turned on or off.
Overall, it works just as expected.

PR #389, opened 2024-03-22
Group level configurator fix

### LINKED ISSUE #304: [BUG] MDANSE_GUI crashes when the grouping_level option is changed
**Description of the error**
Load up a trajectory and open on GaussianDynamicIncoherentStructureFactor then try change the grouping level option, MDANSE will crash with the following error.

```
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\InputWidgets\ComboWidget.py", line 57, in updateValue
    self._configurator.configure(self._field.currentText())
  File "C:\Users\xcb63893\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Configurators\GroupingLevelConfigurator.py", line 113, in configure
    allAtoms = sorted_atoms(trajConfig["chemical_system"].atom_list)
                            ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
KeyError: 'chemical_system'

```

ChiCheng45:
**Description of work**
Fixed the group level configurator.

**Fixes**
Fixes #304

**To test**
Run an analysis type with the group level configurator setting. Try this for different setting - the jobs should complete without crashing.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
@@ -110,7 +110,7 @@ class GroupingLevelConfigurator(SingleChoiceConfigurator):
         trajConfig = self._configurable[self._dependencies["trajectory"]]
         atomSelectionConfig = self._configurable[self._dependencies["atom_selection"]]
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
It is true that the jobs do not crash anymore. The results are still difficult to understand, but this will have to be fixed later.

PR #396, opened 2024-03-26
fix parameter and labelling bug

ChiCheng45:
**Description of work**
Fixed the analysis job crash due to script label update.

**Fixes**
Fixed analysis run bug.

**To test**
Load a trajectory and run an analysis job. Save job as script. Should work without issues.


MBartkowiakSTFC (approved):
It works again. Thanks for the changes!

PR #397, opened 2024-03-27
Make headers GPL compliant and removed __pkginfo__

MBartkowiakSTFC:
**Description of work**
The MDANSE and MDANSE_GUI codes are both licensed under GPLv3. This requires us to use the recommended license notice throughout the code, as per https://www.gnu.org/licenses/gpl-howto.html. Also removed __pkginfo__ files.

**Fixes**
The headers have been replaced with the GPL-compliant headers. The contents of the copyright notice and authors list are not affected by the change.

**To test**
There should be no change to the code operation. You are welcome to check if there are still some files that need updating.


ChiCheng45:
I've removed copyright and authors from the MDANSE header since they are given in other places and are difficult to maintain.

Also removed \_\_pkginfo\_\_ since this information is all specified in the pyproject.yaml.

ChiCheng45 (approved):

PR #398, opened 2024-03-28
Remove qhull

MBartkowiakSTFC:
**Description of work**
MDANSE has been relying on an old implementation of Qhull copied from scipy. This has now been replaced with the Qhull implemented in scipy.spatial, which gives us the advantage of using an up-to-date version of the code. Also, it reduces the level of uncertainty about the licensing of MDANSE, as the code base now contains less code copied from other projects, where the correct authorship contribution was becoming unclear.

**Fixes**
1. Removed the qhull code from MDANSE/Extensions,
2. Added the scipy.spatial import in Voronoi analysis.
3. Removed the setup.py from MDANSE/Extensions, as it is not being used directly.

**To test**
Run Voronoi analysis on a trajectory and check that it produces plottable results.


ChiCheng45 (approved):
Looks good and my tests show that the old and new version give the same results.

PR #400, opened 2024-03-28
move images files and fixed manifest

ChiCheng45:
**Description of work**
Moved the image files in the MDANSE_GUI and fixed the manifest.in.

**Fixes**
Fixed the manifest.in so that the ico and splash screen are in the GUI package.

**To test**
Build the dist and check that the wheel and tarballs contain the splash and icons. Install MDANSE_GUI from the wheel or tarball and check that the splash screen appears and the icon is seen.


MBartkowiakSTFC (approved):
It works, and the wheels/source packages contain the splash screen image and the icon now.

PR #403, opened 2024-04-02
Velocity interpolation order 0 fix

### LINKED ISSUE #360: [BUG] Velocity interpolation order 0 is not used
**Description of the error**
When there is velocity data and interpolation order is set 0 in the VACF for example MDANSE should use velocities to calculate the VACF. As far as I can tell if I set the interpolation order to 0 in the GUI and run VACF, MDANSE instead uses a interpolation order of 1. Also we should change the the interpolation order to a combobox since we only go up to order 5. The GUI allow us to go to higher numbers.

image

When the interpolation order is set to zero MDANSE should use the velocity data if available. 

MDANSE uses a interpolation order 1 instead of zero so doesn't use velocity data if available.

**Suggested fix**
Replace the interpolation order selection to a combobox. Update MDANSE so that it uses velocity data when available, if it is not available then interpolation order zero should not be selectable in the GUI.

ChiCheng45:
**Description of work**
 Fixed temperature, DOS and vacf jobs so that they use velocities with order=0. Updated GUI so that the interpolation order choices go from 0-5. Order=0 is only available when velocity data is available. Order=0 is the default if available.

**Fixes**
Fixes #360

**To test**
- Load up a trajectory without velocity data in the GUI and check that the possible values of the interpolations order is 1-5.
- Check again but with a trajectory with velocity data. The interpolations order values choices should be 0-5. Defaults should be 0.
- Run a job e.g. temperature with order=0. This should work with issues.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
@@ -52,7 +52,7 @@ class InterpolationOrderConfigurator(IntegerConfigurator):
         :type value: str one of *'no interpolation'*,*'1st order'*,*'2nd order'*,*'3rd order'*,*'4th order'* or *'5th order'*.
         """
 
-        if value is None:
+        if value is None or value == "":
             value = self._default
 
         IntegerConfigurator.configure(self, value)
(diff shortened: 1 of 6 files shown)


MBartkowiakSTFC (approved):
I tried it out on a few trajectories. Everything worked as expected.

PR #406, opened 2024-04-03
Updated the ROG docs

### LINKED ISSUE #404: [BUG] Documentation of Radius Of Gyration does not agree with the code
**Description of the error**
The documentation of Radius Of Gyration does not agree with the code. The formulas used in the code and documentation are not the same.

**Suggested fix**
Change the documentation so that it is the same as what is in the code, since the code follows with the formula by IUPAC https://doi.org/10.1351/goldbook.R05121 and https://charmm-gui.org/?doc=lecture&module=scientific&lesson=10.

**Additional details**
Documentation:
image

Code:
```python
...
    if masses is None:
        masses = np.ones((coords.shape[0]), dtype=np.float64)

    com = center_of_mass(coords, masses)

    rog = np.sum(np.sum((coords - com) ** 2, axis=1) * masses) / np.sum(masses)

    if root:
        rog = np.sqrt(rog)

    return rog
```

ChiCheng45:
**Description of work**
Updated in the ROG docs so that it is consistent with the code.

**Fixes**
Closes #404

**To test**
Build the docs and check the ROG maths renders correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,8 @@ share/python-wheels/
 *.egg
 MANIFEST
 
+# doc build files
+Doc/_build/
+
 # pycharm
 .idea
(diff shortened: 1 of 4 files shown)


MBartkowiakSTFC (approved):
Thanks for fixing this.

PR #409, opened 2024-04-04
Updated frames configurator

ChiCheng45:
**Description of work**
Improved the frames configurator testing so that

**To test**
Load up the GUI and try inputting frames setting that shouldn't be allowed. E.g. the last frame was set lower then the first frame.


MBartkowiakSTFC (approved):
It works and the input validation is better now.

PR #410, opened 2024-04-04
Fix action tab for trajectories with the same filename

### LINKED ISSUE #383: [BUG] Frames widget does not detect the trajectory length correctly
**Description of the error**
Frames widget automatically sets the inputs to 0, 100, 1 for trajectories with more than 100 steps.

The Frames widget should set the limits on each of the three fields to (0, len(trajectory)), and set the values of the three fields to 0, len(trajectory), 1.

For short trajectories everything works as expected. If a trajectory ends on frame 5, the Frames widget will show 0, 5, 1.
For longer trajectories (e.g. 1800 frames) the Frames widget shows 0, 100, 1.

**Suggested fix**
Since the mechanism of filling in the fields works in some cases, it probably just needs to be adjusted.

**Additional details**
N/A

ChiCheng45:
**Description of work**
The action tab fails to update when switching to a different trajectory with the same file name.

**Fixes**
Action tab failing to update for trajectories with the same filename.
Closes #383

**To test**
Load two or more trajectories with the same file name. Change the selected trajectory in the actions tab the panel should update.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
@@ -43,7 +43,7 @@ class JobTab(GeneralTab):
         self._job_starter = None
         self._trajectory_combo = QComboBox()
         self._trajectory_combo.setEditable(False)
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
Thanks for the changes, this works better now.

PR #413, opened 2024-04-08
Remove the calculation of formatted time string from Status

MBartkowiakSTFC:
**Description of work**
The calculation of "timedelta" in the Status object causes an occasional random crash with an OverflowError. At the same time, the result of the calculation is not used anywhere.

Note: an individual analysis job should not be spending time calculating its own elapsed time and predicted remaining time. If any process (e.g. GUI) needs this information, they can get the timing information from the Status and do the calculation on their side.

**Fixes**
1. The calculation and formatting of the "timedelta" has been removed from the code.
2. The ._deltas list does not grow indefinitely, and contains only two elements instead. These are [start_time, last_update_time].

**To test**
Unit tests should pass. Please run a converter job and an analysis job from the GUI to be sure that no side effects have been introduced into the status-gui interaction.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/JobStatus.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/JobStatus.py
@@ -16,7 +16,7 @@
 
 import collections
 import pickle
-import datetime
+import time
 import os
 import threading
 import time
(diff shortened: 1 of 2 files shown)


ChiCheng45:
Looks good and seems to work as before.

ChiCheng45 (approved):

PR #415, opened 2024-04-17
Add resolution helper

### LINKED ISSUE #343: [ENHANCEMENT] Improved resolution input
At the moment the instrument resolution is defined by the user using peak function parameters specified in the internal units of MDANSE. This means that a Gaussian function will require the user to input a correct value of $\sigma$ in the units of radian/ps. Most users work with full width at half maximum (FWHM) of the peak, and most likely use meV as the energy unit.

Similar to other widgets, we can have the GUI use an understandable representation of peak width, and then convert it to native MDANSE input.

A specialised tab for instrument parameters would give us more space to visualise the peak, choose the correct unit, etc.

A helper dialog would be another option, but in the long run we will need an instrument tab anyway.

**Additional context**
Some users will not be neutron instrument users at all, so we should dedicate more space to the resolution settings to make it clearer to them which option to pick if they are not familiar with typical values of resolution of neutron instruments.

MBartkowiakSTFC:
**Description of work**
Adds a helper dialog for defining the instrument resolution function. It uses FWHM as the input parameter, and allows the user to pick a unit of energy of the FWHM. Also provides a preview of the peak function.

closes #343

**Fixes**
1. A new dialog has been added to the InstrumentResolutionWidget.
2. If the user changes the main GUI input fields manually, the helper dialog is updated too.

**To test**
Please try running the DynamicIncoherentStructureFactor analysis, and change the resolution function type and peak parameters both in the main GUI and in the helper dialog.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/InstrumentResolutionWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/InstrumentResolutionWidget.py
@@ -69,7 +69,7 @@ class InstrumentResolutionWidget(WidgetBase):
         # first row
         for num in range(1, 3):
             label = QLabel("", parent=self._base)
-            field = QLineEdit("0.0", parent=self._base)
+            field = QLineEdit("N/A", parent=self._base)
             field.setEnabled(False)
             self._layout.addWidget(label, 0, 2 * num)
             self._layout.addWidget(field, 0, 2 * num + 1)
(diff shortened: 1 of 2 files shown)


ChiCheng45:
Looks good, just a few things.

- Would it be better to have an apply button similar to the other helper dialogs? In the atom mapping and atom selection dialogs the setting is only updated after an apply button is pressed. I think that will also get around the issue of the helper dialog not updating when the actual setting is changed.

image

- Some of the setting boxes allow strings which are not numbers. Also some general visual issues. See the pictures below.
 
image

image

MBartkowiakSTFC:
I think that I addressed all the issues other than the matplotlib text overlapping or getting clipped. The problems with matplotlib text labels are mainly a result of bad ratio of the figure dimensions, figure DPI and the Qt widget size; I am not sure how to test for it in the code and correct it if such a problem occurs.

ChiCheng45:
A few more things to do to clean it up a bit before merge - I will add this.

With ideal the setting values should be N/A like before.
image

Ideal peak is not centred and the red line which is used to show the FWHM should not be shown.
Also the FWHM and centre options should be disabled like eta.
image

Text can be entered for the eta value
image

ChiCheng45 (approved):
Looks good and works on my end.

PR #416, opened 2024-04-17
Update to the DOS calculation.

### LINKED ISSUE #353: [BUG] Total DOS should be a sum of the atom DOS?
**Description of the error**
My expectations are that the total DOS should be a sum of the individual atoms DOS. However in MDANSE this is not the case. Maybe MDANSE applies some normalisation to the both the atom and total DOS? What is the standard thing to do in this field?

**Additional details**
This is the DOS I calculated from a H2O trajectory.
image

My expectations would be to see something like this,
image
from this publication
https://pubs.rsc.org/en/content/articlelanding/2013/cp/c3cp52741g

### LINKED ISSUE #408: [BUG] Why does the DOS go less then zero and why do they have negative frequencies?
As far as I understand it the DOS should never be negative but sometimes you can get negative DOS in MDANSE. Also in MDANSE the DOS includes negative frequencies, why is this done? 

**Additional details**
Here is the DOS calculated from the CASTEP test trajectories (croco.md)
image

ChiCheng45:
**Description of work**
Updated the DOS calculation so that results are a bit more clear for new users.

- Removal of the negative frequency part of the DOS.
- Setting of the DOS to arb units. Since MDANSE DOS is actually the FT of a weighted VACF some actual units could be specified like previously. However A.U. are less confusing for new users since you would expect either A.U. or states per unit energy not the previous units of nm2/ps. 
- Set the default weighting scheme to atomic_weight so that the results are now proportional to the actual density of states.
- Changed the DOS so that the sum of atomic density of states equal to the total.

**Fixes**
Closes #408, Closes #353

**To test**
Run a density of states calculation on the protos branch and then on the dos-weight-update branch the total DOS should be the same except the new result should only include positive frequency results and the DOS units should now be au.

**Water diffusion constant DOS(0) test**
Since we know that DOS(0) with equal weighting is proportional to the diffusion constant (D) and d/dt MSD = 6D. I ran the MSD with the long DLPOLY H2O trajectory runs and got the following plot.

image

Doing a simple two point gradient at the endpoints I get a approximate diffusion of 0.633738624175695 / (6 * 50) = 0.00211246208058565 nm^2/ps. Running DOS (equal weight) on the same trajectory I get.

image

DOS(0) = 0.0007601503675207263 and pi * DOS(0) = 0.002388082810226695 (The pi factor occurs because of the 1/2pi factor used in MDANSE for the foward FT). Which is similar to the MSD result above.  

**Total DOS now a sum of atomic**
image

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/InstrumentResolutionConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/InstrumentResolutionConfigurator.py
@@ -78,6 +78,7 @@ class InstrumentResolutionConfigurator(IConfigurator):
         )
         self["n_omegas"] = len(self["omega"])
 
+        # generate the rfftfreq for the positive frequency only results
         self["romega"] = (
             2.0 * np.pi * np.fft.rfftfreq(2 * self["n_frames"] - 1, self["time_step"])
         )
--- a/MDANSE/Src/MDANSE/Framework/Configurators/InstrumentResolutionConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/InstrumentResolutionConfigurator.py
@@ -79,9 +79,7 @@ class InstrumentResolutionConfigurator(IConfigurator):
         self["n_omegas"] = len(self["omega"])
 
         self["romega"] = (
-            2.0
-            * np.pi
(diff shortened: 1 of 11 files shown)


MBartkowiakSTFC (approved):
Thank you for cleaning up the FFT in the resolution functions.

I ran the calculation of the DOS using the old and the new code. The new version is looking good.

PR #417, opened 2024-04-18
Correct RMSD documentation and remove RMSD weight setting.

### LINKED ISSUE #381: RootMeanSquareDeviation weight setting and documentation
A weighting is applied to the RMSD in the RootMeanSquareDeviation job. Looking at the code the weight technically does not follow the procedure described in https://mdanse.readthedocs.io/en/protos/pages/parameters.html#weights. However the current way it is used seems to make some sense.

It seems like when you used something other then equal you are doing some sort of "root weighted mean square deviation"

Also the documentation describes the standard RMSD https://mdanse.readthedocs.io/en/protos/pages/structure.html#root-mean-square-deviation.

**Suggested fix**
Update the documentation and make it clear what MDANSE is doing with RMSD and the weight setting. Or remove the setting so that it follows the documentation.

ChiCheng45:
**Description of work**
Updated and corrected RMSD documentation, the equations in the current version are not correct, see the new version below. Also updated the ROG documentation.

image

I've also removed the weight setting from the RMSD calculation. Here are two reason for why I've decided that the weight setting should be remove from the RMSD calculation.

- RMSD weight setting does not follow the documentation on how the weight setting is applied. In RMSD it is applied inside the property calculation before the application of the root. 
- Keeping the weight setting would mean doing similar things for other calculations or we'd have to have RMSD as some special case which will be confusing for users.

I've also changed the result name from rmsd_total to rmsd_all to make it clear that rmsd_all is not the sum of the atom rmsds.

**Fixes**
Closes #381

**To test**
The weight setting on the RMSD job should be removed. Run RMSD and check the results, this should be the same as the results calculated on the protos branch with equal weight setting.

### DIFF
--- a/Doc/pages/structure.rst
+++ b/Doc/pages/structure.rst
@@ -256,10 +256,10 @@ the difference between two structures. It can be defined as:
 .. math::
    :label: pfx131
 
-   {\mathit{RMSD}{(t) = \sqrt{\frac{\sum\limits_{\alpha = 1}^{N_{\alpha}}\left( {r_{\alpha}{(t) - r_{\alpha}}\left( t_{\mathit{ref}} \right)} \right)}{N_{\alpha}}}}}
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC (approved):
I compared the protos branch results to this branch. Everything worked as expected.

Also, thank you for correcting the documentation.

PR #419, opened 2024-04-19
Atom transmutation update

### LINKED ISSUE #401: [BUG] Atom transmutation
**Description of the error**
Atom transmutation is not implemented and a few analysis jobs have this as an option. 

**Suggested fix**
Something similar to atom_selection and atom_mapping.

ChiCheng45:
**Description of work**
Fixed MDANSE so that atom transmutations works again. The setting has been changed so that a json string is used to map atom indexes to atom types e.g. `{"0": "Be"}`.

A helper dialog is created to help generate the transmutation settings. First the user uses the selection part to select atoms to transmute. Then the transmutation combo box is set to the atom to transmute the selection to and then the transmute button is clicked. The user can do this a few times until they are happy with the transmutation. The use setting button is used to use the setting which updates the setting box in the main window.

Animation16

**Fixes**
Closes #401

**To test**
All unit tests should pass. Use the transmutation and run an analysis job. Open the results and check that results are outputted with the atom types that you transmuted to.


MBartkowiakSTFC (approved):
It all works as expected. I tried creating some deliberately bad inputs, but I did not notice any problems.

PR #422, opened 2024-04-22
Extend unit tests

MBartkowiakSTFC:
**Description of work**
Our current unit tests verify for all the analysis tasks if an .mda file can be created, which indicates that the task completed without an error. However, it turned out that the outputs for DensityOfStates were internally inconsistent (although only for the 'time_window' array), triggering an error when using TextFormat output.

**Fixes**
1. Test of TextFormat output has been added to most analysis types,
2. Test of 'multicore' running mode has been added to unit tests.

**To test**
All tests should pass.


ChiCheng45 (approved):
Looks good.

PR #423, opened 2024-04-24
Fixed and added infrared jobs

ChiCheng45:
**Description of work**
- Added infrared job which calculates the autocorrelation of the time derivative of the molecular dipole which is proportional to the IR intensity. See Eq. 10 from https://doi.org/10.1039/C3CP44302G.
- Dipoles in the dipole autocorrelation and infrared jobs are now calculated for the selected molecule and results are averaged over all instances of that molecule. 
- Dipoles are calculated for the molecule relative to its centre of mass.
- Added the charge mapping helper which is similar to the atom transmutation helper.

**Fixes**
Fixed the dipole autocorrelation jobs.
Fixed angularcorrelation job with a non default frame setting.

**To test**
- Check that the dipole autocorrelation function and Infrared job run successfully and check results plot successfully.
- Generate a trajectory of bulk H2O and run the IR calculation check that the IR results appear reasonable compared to the literature.


MBartkowiakSTFC:
IR_spectrum_MDANSE
For the moment I managed to run the IR spectrum calculation on a solid system. I specified that all the atoms in the cell belong to a single molecule. Was it a meaningful way of running the calculation? Also, I seem to get a lot of negative values (see the plot above) - is the real IR spectrum the square of this curve?

ChiCheng45:
> IR_spectrum_MDANSE For the moment I managed to run the IR spectrum calculation on a solid system. I specified that all the atoms in the cell belong to a single molecule. Was it a meaningful way of running the calculation? Also, I seem to get a lot of negative values (see the plot above) - is the real IR spectrum the square of this curve?

I think this won't be a very good trajectory to run on since there will only be a single dipole. I'd imagine the d/dt dipole autocorrelation function would be highly oscillatory even after a long time so the FFT of this won't be so good. I think the correct thing to have done in this instance would be to average the d/dt dipole autocorrelation function over multiple MD simulations and take the FFT of this average.

MBartkowiakSTFC:
water_IR_macePotential
I have tried the IR calculation out on a trajectory of 216 water molecules. This was a run using the MACE potential (https://arxiv.org/abs/2401.00096) and it was aimed at getting results quickly rather than correctly. Still, I think that this is not a bad spectrum of water.

MBartkowiakSTFC (approved):
The charge setter works well. The IR spectrum calculation works for systems containing molecules, which is a good starting point.

PR #426, opened 2024-04-26
Correct the package tags for PyPI and installation instructions

MBartkowiakSTFC:
**Description of work**
The 'topic' tags in pyproject.toml were not correct at the time of the alpha release. Also, minor corrections were needed in the installation instructions.

**Fixes**
1. Changed topic tags in pyproject.toml to include Physics and Chemistry,
2. Corrected the slash characters in Windows installation instructions,
3. Added the installation instructions for pip install from GitHub,
4. Added `pyqt6-tools` to GUI dependencies. This may not be necessary, but it is possible that some installations on Windows machines will not work without it. Also, **this fixes the PyQt6 version to a slightly older one**.

**To test**
Please install MDANSE_GUI in a clean virtual environment and see if it starts correctly after installing.

### DIFF
--- a/MDANSE/pyproject.toml
+++ b/MDANSE/pyproject.toml
@@ -20,7 +20,8 @@ classifiers = [
         'Development Status :: 3 - Alpha',
 
         'Intended Audience :: Science/Research',
-        'Topic :: Software Development :: Build Tools',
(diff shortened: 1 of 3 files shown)


ChiCheng45 (approved):
I've installed this on a new environment and it loads up without any issues.

PR #429, opened 2024-04-30
Add missing build dependencies to readthedoc files

MBartkowiakSTFC:
**Description of work**
As the requirements on Read The Docs have been gradually changing, we have reached the stage where our documentation build fails. This PR adds the missing configuration which should fix the problem.

**Fixes**
Doc/requirements.txt has been added.
A few broken links on the main page have been corrected.

**To test**
I tested the build manually via my ReadTheDocs account. There is not much testing to do otherwise.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -12,6 +12,6 @@
     configuration: Doc/conf.py
 
  # Dependencies required to build your docs
- # python:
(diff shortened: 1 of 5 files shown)


ChiCheng45 (approved):
OK looks good.

PR #430, opened 2024-04-30
Lammps converter fix

### LINKED ISSUE #421: [ENHANCEMENT] LAMMPS converter unit specification
LAMMPS defines several different unit styles https://docs.lammps.org/units.html that can be used in simulations.
Neither the text format nor the H5MD format can be expected to contain the information about which unit systems was used. This means that, depending on which units were used for a given simulation, the results could be completely meaningless.

At the very least we should add input fields to the LAMMPS converter allowing the user to specify what units were used for each physical quantity in the input files.

We could also scan the LAMMPS input files for the `units` command to recognise which style was used.

**Additional context**
N/A

MBartkowiakSTFC:
**Description of work**
The existing LAMMPS converter works correctly only for a small number of specific cases. LAMMPS gives the user multiple options of the physical unit system output file format.

The converter has been extended to allow the user to pick the unit system, and supports three trajectory formats: LAMMPS custom format (this has been considered to be the ".lammps" format so far), XYZ and LAMMPS implementation of H5MD.

In principle, another change will be necessary in the future to account for different `atom_style` options, which will change the format of the structure file (known in MDANSE as the 'config' file).

closes #421 

**Fixes**
Added multiple readers to the LAMMPS converter.
Added the unit system selector.
Included additional trajectories for unit testing, together with the input files used to generate them.
Added new unit tests for the LAMMPS converter.

**To test**
All tests must pass.
Also, please try to convert a LAMMPS trajectory using the protos branch and again using this branch. The results should be the same.
Optionally, please try to manually convert the new trajectory files from unit test data and check that they all result in the same output.


ChiCheng45:
Looks good. How do we convert lammps h5md and other files from the GUI? or is this not supported?

MBartkowiakSTFC:
I just used the combo box in the converter panel to specify that the input file is h5md or xyz. You are right that we should probably change the filters in the file dialogs. For some reason my LAMMPS files never have the extensions that the converter expects, so I always have to switch to `all files (*.*)` anyway, even when using the LAMMPS custom trajectory.

I will check if there is any official LAMMPS convention for file name extensions. Maybe I have been naming my files wrong, and MDANSE converter is right.

MBartkowiakSTFC:
Looking at the examples in the LAMMPS documentation:
https://docs.lammps.org/dump.html
https://docs.lammps.org/read_data.html
I would say that there is no conventional file name extension for neither trajectory nor structure files. Unless it's an H5MD trajectory (.h5).

ChiCheng45 (approved):
Looks good.

PR #432, opened 2024-05-03
Remove hardcoded background colour in the GUI dialogs

MBartkowiakSTFC:
**Description of work**
Hardcoded font/background colour made it impossible to see the text in some GUI components, notably the chemical element viewer and the data list in the plotter. The solution is to remove the custom formatting wherever possible and have the GUI rely on the system style settings.

**Fixes**
Colour setting code has been removed wherever necessary.

**To test**
Please start the GUI and open the following GUI elements:
1. Periodic table viewer.
2. Atom information in the periodic table viewer for any specific atom.
3. Data plotter with any data file loaded.
Try changing the style sheets. All the background and font colours should change accordingly.


ChiCheng45 (approved):
Looks good. I tested on Windows with dark mode and I can see the text now.

PR #433, opened 2024-05-03
Reduce the number of workflows triggered per commit

MBartkowiakSTFC:
**Description of work**
At the moment, the unit tests are run when code is uploaded to the branch, and again when a PR is created. This is unnecessary. Wheel building is also unimportant now that there are easier ways of installing MDANSE from PyPI or directly from GitHub. This PR switches off the workflows which we can consider to be unnecessary at the moment.

**Fixes**
1. Unit tests will only run when a PR is created.
2. Black will still run every time we upload code.
3. Wheels will only be built when we create a release.

**To test**
N/A


ChiCheng45 (approved):
Looks good.

PR #434, opened 2024-05-08
3D selection update

ChiCheng45:
**Description of work**
Initial implementation of the 3d selection into MDANSE. Added atom name and fullname to the selection options. The invert selection was removed as it would make the updating from the 3d view to the checkbox and comboboxes more difficult.

Animation17

**To test**
Load a trajectory, open the atom selector, and try out the selection of atoms in the 3d view. The checkbox, comboboxes, and textboxes should update. Try changing the selection in the checkbox and combobox selection the 3d view should update. Test out a few different things and click the use settings and run the job. Try testing out the atom charge and atom transmutation as well.


MBartkowiakSTFC:
So far I managed to crash the GUI once. It was using the cobrotoxin trajectory. I was switching different selections on and off  reasonably fast. I ticked all atoms + dummy atoms on and off, all atoms on and off, and when I tried doing dummy atoms on and off, it crashed with this output:
```
Python(55329,0x103188580) malloc: Incorrect checksum for freed object 0x12832e400: probably modified after being freed.
Corrupt value: 0x101010101010101
Python(55329,0x103188580) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```
This is fairly low-level, most likely something within VTK, so I am not sure if we can do much to prevent it from happening. But maybe for some time-consuming operations we should block the GUI until the operations have completed?

MBartkowiakSTFC:
I have been testing it in more detail. So far, I can list a few things that do NOT cause a crash:
1. Switching atoms/bonds on and off in the main 3D viewer.
2. Alternately activating and deactivating the "all atoms" selection in the selection dialog.
However, switching other selections on and off sooner or later leads to segmentation fault or (more frequently) to the malloc error described above.

Of course, the exact text of the malloc error is specific to a Mac, and it would be good to confirm that an equivalent error can be generated on other platforms.

ChiCheng45:
@MBartkowiakSTFC Thanks for spotting that. Not exactly sure why the crash occurred but stopping vtk from drawing bonds when there aren't any seems to have fixed things.

MBartkowiakSTFC (approved):
I don't see any other problems. The current version works well for me.

PR #437, opened 2024-05-20
simplify json selection setting

ChiCheng45:
**Description of work**
Updated the json selection setting so that it is simplified slightly.

Before 
`{"all": false, "element": {"H1": true}, "index": {"144": true, "387": true, "639": true}}`

After:
`{"all": false, "element": ["H1"], "index": [126, 252, 630, 663]}`

**To test**
Run a job with atom selection and check that it behaves correctly.


MBartkowiakSTFC (approved):
I ran a few jobs from the GUI and by saving inputs to a script. Both options worked and the results were consistent.

PR #439, opened 2024-05-23
Atom int name selection fix

ChiCheng45:
**Description of work**
Sometime the atom name could be converted to an int e.g. "1". This causes some issues with the atom selector. This PR fixes this issue.

**To test**
Run an analysis job with atom selection and check that it works correctly.


MBartkowiakSTFC:
Do you remember for what trajectory/analysis combination this problem could occur? I have not noticed it so far.

ChiCheng45:
> Do you remember for what trajectory/analysis combination this problem could occur? I have not noticed it so far.

It was the lammps trajectory from sanghamitra it's on onedrive. p_lammps_10000_fixed.lammps.

MBartkowiakSTFC (approved):
I managed to reproduce the problem and confirm that this PR fixed it. Thank you for the changes!

PR #440, opened 2024-05-23
Add tags to output datasets in MDA files

MBartkowiakSTFC:
**Description of work**
This PR adds an extra attribute called 'tags' to IOutputVariable. As a result, the HDF5 datasets in the MDA output file will have a 'tags' attribute as well. This does not change the output as such, but is a starting point allowing to store useful hints for the data plotter.

The DOS calculation will now label 'dos_total' as 'main', and 'dos_{element name}' as 'main,partial'. The goal is to let the plotter create a quick plot out of an entire output file, where the main result is plotted first (solid line) and the partial results are added (using dashed lines, etc.)

**Fixes**
1. The 'tags' attribute has been added to IOutputVariable,
2. For each job, keyword arguments 'main_result = True' and 'partial_result = True' have been manually added for the outputs considered the main/partial results of a given analysis type.

**To test**
All tests should pass. The tags are not used at the moment.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/AngularCorrelation.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/AngularCorrelation.py
@@ -121,6 +121,7 @@ class AngularCorrelation(IJob):
             (self.configuration["frames"]["number"],),
             axis="time",
             units="au",
+            main_result=True,
         )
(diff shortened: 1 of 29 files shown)


ChiCheng45 (approved):
Looks good.

PR #441, opened 2024-05-23
Output xaxis preview

### LINKED ISSUE #412: [ENHANCEMENT] Preview for Configurator outputs
The GUI takes user inputs and passes them to the MDANSE configurators. If the inputs are invalid, the offending entries are highlighted and the hovertext contains a brief explanation of what went wrong. However, for many input widgets it may still not be clear to users if their inputs worked correctly. For example, if I define Frames=0,100,1 I would like to know what the first, second and last frame numbers are - will it be 0,1 and 100, or will it be 0,1 and 99?

The GUI needs to include an option of displaying the result of each configurator, for the user to verify the inputs. For commonly used configurators such as Frames we should extend the output to include not only the frame numbers but also corresponding time values, etc.

This can be implemented in different ways. Some discussion will be necessary before we decide on one.

**Additional context**
N/A

MBartkowiakSTFC:
**Description of work**
Action widget now includes a QTextEdit preview field, which informs users about the data axis of the output. This allows the user to check how a change to the input parameters affects the output of the analysis. At the moment the main application is to confirm that the frame selection results in a sufficient energy resolution and range.

At the moment only FramesConfigurator, InstrumentResolutionConfigurator and QVectorsConfigurator create a preview of an output axis. The automatic conversion of the units will be added in another PR.

closes #412 

**Fixes**
1. Configurators now have a new method: `preview_output_axis`.
2. Action collects and displays the output axis of each configurator using the `show_output_prediction` slot.
3. Extra checks have been added to prevent the GUI from crashing when only a single frame is selected. A single frame is a valid choice for FramesConfigurator, but makes the InstrumentResolutionConfigurator fail.

**To test**
Load a trajectory into the GUI and select an analysis in the "Actions" tab.
Try changing the frames input and check if it affects the preview in the text field just above the "RUN" button.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/AxisSelectionConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/AxisSelectionConfigurator.py
@@ -85,5 +85,7 @@ class AxisSelectionConfigurator(IConfigurator):
         :return: the information about this configurator
         :rtype: str
         """
+        if "value" not in self:
+            return "Not configured yet\n"
 
         return "Axis vector:%s\n" % self["value"]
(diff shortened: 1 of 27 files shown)


ChiCheng45 (approved):
I've tested this out and it looks good.

Issue #435, opened 2024-05-10
[ENHANCEMENT] change correlation function so that the number of configurations used to determine the average is the same for all times


ChiCheng45:
**Is your feature request related to a problem? Please describe.**
Currently, the correlation calculation is done so that the number of configurations it averages over is different for different times $t_0 + \Delta t$. This is usually why we see slightly larger correlations at the large $t_0 + \Delta t$. 

For example the autocorrelation of 1 1 2 3

t_0 - averages over 4 configurations
```
1 1 2 3
1 1 2 3
```

t_1  - averages over 3 configurations
```
  1 1 2 3
1 1 2 3
```

t_2  - averages over 2 configurations
```
    1 1 2 3
1 1 2 3
```

t_3  - averages over 1 configuration
```
      1 1 2 3
1 1 2 3
```

**Describe the solution you'd like**
The trajectory should be halved or some other fraction and then correlated with the full trajectory.

E.g. with the 1 1 2 3 trajectory

t_0
```
1 1
1 1 2 3
```

t_1
```
  1 1
1 1 2 3
```

t_2
```
    1 1
1 1 2 3
```

Although the number of steps $t_0 + \Delta t$ decreases and the number of configurations they average over is smaller for some times, they all average over the same number of configurations. This can be offset by running longer MD simulations. With the old scheme, you will probably always see some spuriously large correlations at the large $t_0 + \Delta t$ time no matter how long you run the MD simulation for.

**Describe alternatives you've considered**
A truncation procedure could be done. This would be similar in the end but I think it is preferable to use the above since it means the same number of configurations are used for each time step. Additionally, the proposed method is similar to the correlation function defined in mcquarrie statistical mechanics.

![image](https://github.com/ISISNeutronMuon/MDANSE/assets/14276033/961e1997-0209-47a7-ae45-7b0cf3e7da8d)

You see here that the window is of a fixed size (before taking the limit). The truncation procedure always has smaller sized windows at a large time leading to an asymmetry.

**Additional context**
Here is an example PACF from #212. We can see that the correlation spuriously increases at the very end.
![image](https://github.com/ISISNeutronMuon/MDANSE/assets/14276033/f7bcc38f-62d5-4e21-896c-09ce06de89f8)

Jobs to check
---------------
**Dynamics**
- [x] Angular correlation #444
- [x] CurrentCorrleationFunction #566
- [x] DOS #444
- [ ] GeneralAutoCorrelationFunction
[...]

PR #444, opened 2024-05-24
Correlation update for Angular correlation, DOS, MSD, PACF and VACF

ChiCheng45:
**Description of work**
Applied the correlation update to some dynamics jobs following #435. The approach that I'm that is proposed in the PR is similar to the correlation function defined in mcquarrie statistical mechanics.

image

An alternative truncation-type procedure would be different from the above (before taking the limit) since the window size $-T$ to $T$ decreases for increasing $\tau$.

A new option in the frames setting has been added "correlation frames". This gives you the resulting number of frames in your correlations function. E.g. With VACF if "correlation frames" is set to 5 you will have 5 time steps in your VACF.

image

For the default setting I set the correlation frames used to around half of the total frames available.

**Angular Correlation**
Spurious oscillation at long time are removed.

Before
image

After
image

**DOS**
Overall noise of the DOS is reduced.

Before
image

After
image

**MSD**
Some small oscillation at long time are removed. The new MSD appears linear at long times while the old scheme doesn't look straight for long times.

Before
image

After
image

**PACF**
Spurious oscillation at long time are removed.

Before
image

After
image

**VACF**
Spuriously increasing correlation a long times are removed.

Before
image

After
image

**To test**
Run one of the dynamics jobs and check everything is working correctly. Try changing the number of correlation frames and check that it works as expected.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/CorrelationFramesConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/CorrelationFramesConfigurator.py
@@ -42,6 +42,6 @@ class CorrelationFramesConfigurator(FramesConfigurator):
                                  f"should be greater then zero.")
             return
 
-        self["n_cframes"] = c_frames
-        self["ctime"] = self["time"][:c_frames]
-        self["cduration"] = self["ctime"] - self["ctime"][0]
+        self["n_frames"] = c_frames
+        self["time"] = self["time"][:c_frames]
+        self["duration"] = self["time"] - self["time"][0]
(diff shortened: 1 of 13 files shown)


MBartkowiakSTFC (approved):
The results are better now using the default parameters, while the option of changing the correlation window size makes it possible to go back to the old behaviour if necessary. Also, the new input for frames still works well with the x-axis preview widget.

I expect that the new CorrelationFramesConfigurator will require us to add extra documentation, explaining to the users the meaning of the parameter that they are changing.

Overall, I did not find any problems while testing. Thanks for the changes!

PR #445, opened 2024-05-28
Q Vector generator fix

MBartkowiakSTFC:
**Description of work**
The code surrounding the q-vector generation has turned out not to be as stable as the rest of MDANSE.

**Fixes**
1. Corrected the error checks around the individual lines of parameters. This prevents the GUI from crashing on invalid input.
2. Corrected the q-vector generation code in ApproximateDispersionQVectors and MillerIndicesQVectors. There were still problems there resulting from conversion from Python 2.

**To test**
Try typing in wrong parameters for different q-vector generators in the GUI.
Run Dynamic Incoherent Structure Factor calculations using ApproximateDispersionQVectors and MillerIndicesQVectors to check if it completes without errors.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/FramesConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/FramesConfigurator.py
@@ -124,9 +124,17 @@ class FramesConfigurator(RangeConfigurator):
         :rtype: str
         """
 
-        return "%d frames selected (first=%.3f ; last = %.3f ; time step = %.3f)\n" % (
-            self["n_frames"],
-            self["time"][0],
-            self["time"][-1],
-            self["time_step"],
-        )
+        try:
+            result = (
+                "%d frames selected (first=%.3f ; last = %.3f ; time step = %.3f)\n"
+                % (
+                    self["n_frames"],
+                    self["time"][0],
+                    self["time"][-1],
+                    self["time_step"],
+                )
+            )
+        except KeyError:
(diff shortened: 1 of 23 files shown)


ChiCheng45:
I've added a commit to set the background color to the default. Some issues with dark scheme.

Before
image

After
image

ChiCheng45:
Some crashes are still occurring here some I've found from my initial testing.

Setting the following q vector settings:
- CircularQVectors
  - Set axis_1 and axis_2 to the same values.
  - Set axis_1 or axis_2 to a zero vector.

- LinearQVectors
  - Set axis to a zero vector.

Also, the GUI doesn't update properly. When I misconfigure the q-vectors it changes purple, as expected, when I change to a different q-vector generation type e.g. from CircularQVectors to LinearQVectors it stays purple.

Crashes when I try to use ApproximateDispersionQVectors with the DFTB water trajectory.
Dynamic Incoherent Structure Factor calculation using MillerIndicesQVectors works fine.

MBartkowiakSTFC:
Thanks for testing it in detail. I will remove the problems you listed here.

ChiCheng45:
ApproximateDispersionQVectors and MillerIndicesQVectors seem to be working fine now. Everything seems to be working well.
 
 The only problem I've come across so far now is that when you set an invalid setting it still lets you run the calculation. It completes fine, I think it must use the default settings.
 
 
image

ChiCheng45 (approved):
Looks great thanks for sorting the issues out.

PR #449, opened 2024-05-30
update IR jobs to use correlation frames

ChiCheng45:
**Description of work**
Applied the correlation update IR jobs following #435. Here are some tests I did using a water trajectory.

dipole autocorrelation
Before 
image

After
image

Infrared
- derivative dipole autocorrelation
Before 
image
After
image

- IR
Before
image
After
image

**To test**
Run all IR jobs and check that they are completed successfully.


MBartkowiakSTFC (approved):
I did not find any problems on my computer. The results are looking good.

PR #451, opened 2024-05-31
widget window fix

### LINKED ISSUE #448: [BUG] Job trajectory widgets open up in its own windows
**Description of the error**
Job trajectory widgets open up in its own window. This seems to occur when you have a large trajectory selected in the drop-down and you quickly switch between different jobs. 

**Additional details**
image

ChiCheng45:
**Description of work**
Fixes the widget window bug. When the job selection is changed the widgets in that layout need to be deleted before the widgets of the new jobs are created and then added to the layout. This change makes sure that they get deleted.

**Fixes**
Closes #448 

**To test**
Load up a long trajectory and go to the jobs tab. Quickly switch between different jobs the bug seen in #448 should not occur. Run a job and load the results, everything should work correctly.


MBartkowiakSTFC (approved):
I could not reproduce this problem on a Mac, but I can confirm that the GUI still works fine after changes.

PR #452, opened 2024-05-31
Multiprocessing jobs

### LINKED ISSUE #361: [BUG] Mean square displacement job status goes over 100% complete
**Description of the error**
When running a mean square displacement job the percentage complete can go over 100%.

**Additional details**
Animation13

### LINKED ISSUE #447: [BUG] DISF calculation: CPU usage cannot be controlled
**Description of the error**
At the moment a single-core Dynamic Incoherent Structure Factor calculation uses more approx. 250% CPU cores on my computer. While it remains to be seen if there is additional parallelism in place and how the code performance is affected, the main problem is that resources of the host computer cannot be accurately allocated. We need to know in advance how many CPU cores the job will use if we run it on a cluster, etc.

1 CPU core requested = 100% of a CPU core in the system manager

1 CPU core requested = 250% CPU cores used in the system manager

**Suggested fix**
Find the part of the code which uses more resources, and add some way of controlling it.

**Additional details**
Other jobs don't produce this problem.

MBartkowiakSTFC:
**Description of work**
At the moment, multicore jobs work, but do not report progress correctly, and cannot be aborted from the GUI.
This PR adds a mechanism for aborting multicore jobs, and changes the way progress is reported for multicore jobs.

closes #447 
closes #361 

**Fixes**
1. Added an extra Event (end_event) to the JobStatus. This is checked by a running job, and can be used to abort a job
2. Added a new method to Status, which allows a Job to report progress directly from the main thread and not from subprocesses. The progress is now the size of the output queue.
3. Set OMP_NUM_THREADS to 1 in the GUI launcher, suppressing numpy multithreading.

**To test**
Run a job using multiple cores. Check if the progress is updated correctly.
Run a long job using multiple cores. Abort it before it finishes, and confirm that every child process terminates eventually, without being killed manually by the user. (For slower jobs this may take some time, but should happen eventually).

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -272,8 +272,6 @@ class IJob(Configurable, metaclass=SubclassFactory):
                         if not self._status._end_event.is_set():
                             return False
                 output = self.run_step(index)
-                if self._status is not None:
-                    self._status.update()
                 outputs.put(output)
 
         return True
@@ -304,6 +302,8 @@ class IJob(Configurable, metaclass=SubclassFactory):
         while inputQueue.qsize() > 0:
             if self._keep_running:
                 time.sleep(0.1)
+                if self._status is not None:
+                    self._status.fixed_status(outputQueue.qsize())
             else:
                 for p in self._processes:
                     p.terminate()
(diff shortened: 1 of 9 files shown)


ChiCheng45:
The progress update and the terminate seem to kill most of the child processes. 

However, it looks like there is still one more child process that it doesn't kill. Not sure if I needed to wait longer but I gave it a few minutes.

image

Also, I noticed that the MSD job goes to 100% then 0% then back to 100%. I guess this the multiprocessing version of #361.

Animation19

MBartkowiakSTFC:
This PR will be replaced by #471

PR #454, opened 2024-06-06
Split plotting gui

### LINKED ISSUE #332: [ENHANCEMENT] Plotting should be a tab
Since plotting is an important part of the MDANSE workflow, it should be a tab.

Add a plot tab probably something similar to the trajectory tab we have at the moment.

MBartkowiakSTFC:
**Description of work**
A prototype of the new plotting interface has been added as two separate tabs.

**Fixes**
1. It is possible to overplot results originating from different files.
2. Physical unit switching is possible using combo boxes in the plot tab.
3. Single, Grid and Heatmap plotting modes are available at the moment.
4. Sliders can be used to modify the plot contents. Their function depends on the plotting mode.

**To test**
Load several files containing data analysis results.
Try plotting the results in a single plot.
Change the plotting mode and the physical units to check if the changes are applied correctly.


ChiCheng45:
Looking good so far.

So far some bugs I've found from testings.
- Crash when a double left or right click is made on the tree view of a loaded file data.
-  With heatmap plot changing the maximum or minimum too quickly passed one another means it doesn't update to the minimum value maximum can have. 

Animation19


Also some suggestions.
- Let's disable the old plotter.
- Add an option to remove an MDA file from the plot creator?
- Add an option to remove a dataset from the plot holder?
- Add an option to delete a plot?
- Update so that the plot stays zoomed in after a change is made.
- I can plot things like the DOS and VACF to the same plot. Should we stop the user from doing this?

ChiCheng45:
Looks good just have a few more suggestions.

- 2D plotting seems quite slow when it is updating e.g. when the minimum/maximum sliders are moved.
- Can we block the maximum/minimum sliders so they don't go past each other?
- When the units are changed in the 2D plotter the plot range doesn't update.

image

Also I found a crash when you create a few new plots and then delete them.

Animation20

MBartkowiakSTFC:
Thanks for spotting these additional problems!

Latest changes:
- The GUI should not crash now when a plotting tab is deleted.
- The 2D plot update is faster now when slider is used.
- The 2D plots should change the scale correctly when units are changed.

Problems that remain, but maybe are not critical:
- in 1D plots, it may be necessary to click the 'home' button to get the correct plot limits after changing units or plotting offsets,
- The sliders are **not** connected to prevent maximum from becoming lower than minimum. The reason for this is that this is the same GUI component that is used also for 1D plots, and then no such restriction applies anymore.

ChiCheng45:
Looks great I'm happy to merge just one last thing.

image

What's the purpose of the checkboxes in the plot creator?

MBartkowiakSTFC:
The check boxes are only useful for enabling/disabling individual curves in the plot. At the moment both tabs, Plot Creator and Plot Holder, use the same Qt data model, but some parts are only useful once the data model has been attached to a plot.

I guess I _could_ make this check box do something in Plot Creator. As in, only the checked items would be transferred to the plot.

ChiCheng45:
Thanks for the above changes. Some more crashes I've found that seem related.

Crash when attempting to plot nothing.
Animation21

Crash when nothing is plotted and the offset setting is changed.
Animation23

ChiCheng45 (approved):
Looks great!

PR #455, opened 2024-06-07
Change EISF weight default

ChiCheng45:
**Description of work**
Currently the default weight for EISF is b_incoherent. According to https://mdanse.readthedocs.io/en/protos/pages/scattering.html#elastic-incoherent-structure-factor

image

and 

image

therefore the default weight should be b_incoherent2.

**To test**
Check that the default weight for EISF is b_incoherent2.

### DIFF
--- a/Doc/pages/scattering.rst
+++ b/Doc/pages/scattering.rst
@@ -606,7 +606,7 @@ one can write:
 .. math::
    :label: pfx101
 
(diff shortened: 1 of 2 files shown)


MBartkowiakSTFC (approved):
The new default weights are applied correctly. The default values are now the same as in the documentation.

PR #456, opened 2024-06-07
Scattering correlation update 1

ChiCheng45:
**Description of work**
Updated dynamic incoherent structure factor and Gaussian dynamic incoherent structure factor jobs to use the correlation method following #435. 

cp2k water trajectory comparison.

DISF (sphericalqvectors) 
Before
image
image

After
image
image

GDISF (MSD is linear for long times so GDISF should go something like exp(-x))
Before
image
image

After 
image
image

**To test**
Run dynamic incoherent structure factor and Gaussian incoherent structure factor jobs and check they are working as expected.

### DIFF
--- a/MDANSE/Extensions/.gitignore
+++ b/MDANSE/Extensions/.gitignore
@@ -1,4 +1,6 @@
 /*.c
 /*.cpp
+/*.html
 /xtc/xtc.c
 /xtc/trr.c
+/xtc/*.html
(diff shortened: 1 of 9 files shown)


gonzalezma:
Could you check the DISF in several systems? My impression from your figures is that the DISFs before and after are not decaying with the same relaxation rate, which would imply that one of them is not correct.

ChiCheng45:
> Could you check the DISF in several systems? My impression from your figures is that the DISFs before and after are not decaying with the same relaxation rate, which would imply that one of them is not correct.

I think they look similar. The x-axis range is different here are the same plots with the same range.

Before
image
image


After
image
image

MBartkowiakSTFC (approved):
The results look good.

PR #457, opened 2024-06-10
Improve GUI layout

MBartkowiakSTFC:
**Description of work**
Cosmetic changes which should give the users a better starting point when they launch a new GUI.

**Fixes**
1. In each tab, the right panel is now larger than the left panel. The main advantage is that the 3D view of the trajectory is now larger.
2. The parameter table in the QVectorsWidget is now given more space, so that all the rows can be seen simultaneously.

**To test**
Start the GUI and load a trajectory.
Pick a scattering analysis (e.g. DISF) and check the size of the QVectorsWidget.


ChiCheng45 (approved):
Looks great!

PR #460, opened 2024-06-13
Trajectory wrapper: support for H5MD

MBartkowiakSTFC:
**Description of work**
Trajectory class is now a wrapper around MdanseTrajectory or H5MDTrajectory. This way trajectories in H5MD format can be loaded directly into MDANSE.

It is still necessary for the H5MD trajectory to contain the information about the physical units of the quantities saved in the file. This means that an H5MD file created by LAMMPS is not usable.

**Fixes**
1. A new test trajectory file has been added.
2. New unit tests load an H5MD file in Analysis/test_mdmc_h5md.py
3. Most of the code from the Trajectory class is now in MdanseTrajectory. Trajectory decides internally if it should create an MdanseTrajectory or H5MDTrajectory.

**To test**
1. Please run an analysis on an MDANSE trajectory using the main branch (protos), and the same one using the current branch. The results should be the same for both runs.
2. Load the trajectory from MDANSE/Tests/UnitTest/Data/Ar_mdmc_h5md.h5 and confirm that it can be viewed and animated.
3. Run a few analysis types on the trajectory from the GUI.


MBartkowiakSTFC:
At the moment the H5MD trajectory implementation spends some time on unit conversion, making sure that the values presented to the analysis code are expressed in internal MDANSE units. It may be a good idea not to do it and make sure that the output units are changed correctly depending on the input units. The data plotter typically converts the units of the analysis results anyway.

ChiCheng45 (approved):
I've been able to complete all tests. I've run several analysis jobs with the h5md trajectory file and I've not come across any issues.

PR #463, opened 2024-06-13
Disable jobs with no atoms selected

### LINKED ISSUE #461: [BUG] Crash when no atom are selected
**Description of the error**
Running a job such as MSD with no atoms selected causes MDANSE to crash.

**Suggested fix**
Update the selection configurator so that when no atoms are selected the MDANSE blocks the job from running.

MBartkowiakSTFC:
**Description of work**
Fixes the problem of the GUI crashing when starting a job with total_steps=0.

closes #461 

**Fixes**
1. Added a check in JobHolder before dividing by 'total_steps'.
2. AtomSelectionConfigurator becomes invalid if the selection is empty.

**To test**
Load a trajectory, pick an analysis and deselect all the atoms. The widget should become invalid, and the "run" button inactive.


ChiCheng45 (approved):
Looks great thanks for fixing this.

Issue #446, opened 2024-05-30
[ENHANCEMENT] Q vector generation: path in reciprocal space


MBartkowiakSTFC:
**Is your feature request related to a problem? Please describe.**
As far as I can tell, none of the q vector generators in MDANSE can create vectors along a path. (E.g. (0,0,0) -> (0.5,0,0) -> (0.5,0.5,0) ). Also, while I could use the DispersionLatticeQVectors to generate each of these segments separately, it seems that the HKL indices of the points are not written to the output. 

**Describe the solution you'd like**
A q vector generator should be added which creates vectors along an arbitrary path.
The way vectors are written to the output file needs to be changed so that the HKL indices of the vectors are saved.
Plotter has to be extended to plot the results as a function of q. (At the moment it only does |q|).

**Describe alternatives you've considered**
Suggestions welcome.

**Additional context**
N/A

MBartkowiakSTFC:
In an MD simulation of a crystalline solid, the simulation box is most likely going to be a supercell. The HKL values in this situation should correspond to the unit cell, and not the supercell.

gonzalezma:
Are you sure? I do not see how you could handle this easily (in most cases when you read a trajectory you will not know if it is a single unit cell or a n*m*p supercell) and nothing ensures that the MD will maintain the symmetry of the unit cell. I would imagine that if you have e.g. a 4x4x4 supercell, the user will interpret the vector (4,0,0) as the vector (1,0,0) of the unit cell. But I don't work with crystals, so you decide.

MBartkowiakSTFC:
You are right that this will require some effort to get it right. So, the more problems with this idea you point out now, the better we can prepare our implementation.

In my own simulations of crystals, I typically specify the input structure as a single cell, and then use the MD engine commands to replicate it, as in:
https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/CELL.html#CP2K_INPUT.FORCE_EVAL.SUBSYS.CELL.MULTIPLE_UNIT_CELL
or
https://docs.lammps.org/replicate.html
[...]

PR #465, opened 2024-06-14
Average structure analysis

MBartkowiakSTFC:
**Description of work**
A new analysis type has been added in this PR. It is a tool mentioned in #446 allowing the user to output the time-averaged crystal structure of the simulated system. The intention is to use it with an external tool like phonopy or spglib to determine the symmetry of the system and find the single unit cell of the supercell that was being simulated.

**Fixes**
1. New analysis added: AverageStructure.
2. OutputStructureConfigurator added, specifically for choosing an output format supported by ASE.
3. A matching OutputStructureWidget has been added to the GUI.
4. Unit tests for AverageStructure runs check that it executes without crashing for several combinations of parameters.

**To test**
Tests must pass. You can run the AverageStructure analysis on a crystal(like) system; the LAMMPS CuSbS trajectory can be used if you exclude the Cu atoms.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/OutputStructureConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/OutputStructureConfigurator.py
@@ -1,0 +1,127 @@
+#    This file is part of MDANSE.
+#
+#    MDANSE is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


ChiCheng45:
Looks great. I generated some cif files of some averaged structures and they loaded up in mercury successfully.

image

image


It won't make much (if any) difference but what do you think about running average over the atomic positions in fractional coordinates first then converting back to cartesians with the averaged unit cell rather than the current approach which runs the average over cartesian coordinates?

I think the two approaches would be the same if the unit cell is cubic and stays cubic during the simulation but I wonder if it would be different if it is not cubic and the unit cell angles change.

MBartkowiakSTFC:
I thought that using the Cartesian coordinates would result in a slightly lower risk of getting wrong results. The averaging of the unit cell, on the other hand, is something that I would worry about in general, independent of how the coordinates were handled.

I cannot really predict how bad the worst case scenario would be, since generally the expectation we have that the atoms stay around their equilibrium positions normally should correspond to a reasonably stable unit cell shape as well.

ChiCheng45 (approved):

PR #467, opened 2024-06-17
Scattering correlation update 2

### LINKED ISSUE #459: [BUG] NeutronDynamicTotalStructureFactor: mismatch between default DISF and DCSF parameters
**Description of the error**
Normally, it was enough to calculate the Dynamic Coherent and Dynamic Incoherent Structure Factors (DCSF and DISF) with the default parameters to produce matching inputs for the NeutronDynamicTotalStructureFactor (NDTSF) calculation. Now, the correlation window settings for DISF make the normal output shorter (in time units) than the DCSF.

The standard inputs of DCSF and DISF analyses should produce input files which work for the NDTSF analysis.

At the moment, the two analysis types are handled differently, and the NDTSF run fails.

**Suggested fix**
Could NDTSF analysis automatically select the right frames from the DCSF output? The analysis window effect on the DISF time axis could be accounted for by truncation of the DCSF results.

**Additional details**
N/A

### LINKED ISSUE #464: [BUG] DCSF assumes f(q,t)_AB = f(q,t)_BA is this correct?
**Description of the error**
In the DSCF job the following correlation is done.

```
    def combine(self, index, x):
        """
        Combines returned results of run_step.\n
        :Parameters:
            #. index (int): The index of the step.\n
            #. x (any): The returned result(s) of run_step
        """

        if x is not None:
            for pair in self._elementsPairs:
                corr = correlation(x[pair[0]], x[pair[1]], average=1)
                self._outputData["f(q,t)_%s%s" % pair][index, :] += corr

```

The pairs in `self._elementPairs` are unique pairs. In the weight section of the code it uses a symmetric setting so that it weights it according to the fact that the combination AB and BA occur. However, cross-correlations are not symmetric in general. 

**Additional details**
Here I calculate the DCSF with water trajectory and the grid qvector generator. I use this generator so that the qvectors aren't random.

image

Here I switch the pairs around in the code e.g.

`                corr = correlation(x[pair[1]], x[pair[0]], average=1)`

image

ChiCheng45:
**Description of work**
Updated DCSF to use correlation frames following #435.

DCSF with DLPOLY water trajectory and SphericalQVectors
Before
image

After
image

**Fixes**
Fixes #464
Fixes #459

**To test**
Run DCSF and DISF with the default number of correlation frames. Using the output files run NDTSF. All jobs should complete without any issues.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
@@ -18,12 +18,12 @@ import collections
 import itertools
 
 import numpy as np
-
+from scipy.signal import correlate
 
 from MDANSE.Core.Error import Error
 from MDANSE.Framework.Jobs.IJob import IJob
 from MDANSE.Mathematics.Arithmetic import weight
-from MDANSE.Mathematics.Signal import correlation, get_spectrum
+from MDANSE.Mathematics.Signal import get_spectrum
 
 
 class DynamicCoherentStructureFactorError(Error):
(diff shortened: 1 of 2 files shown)


ChiCheng45:
failing tests due to macOS-11 deprecation. #468 opened for this.

MBartkowiakSTFC (approved):
I tried it out on a couple of trajectories and did not run into any problems.

PR #470, opened 2024-06-18
Distance cutoff widget update

### LINKED ISSUE #431: [BUG] Unusual RDF result in pair distribution function job.

### LINKED ISSUE #462: [ENHANCEMENT] Set the maximum r value with the PDF job.
Beyond a specific length, the calculated PDF and other functions are not useful as they begin to add distances that go beyond the box length.

Set the maximum length that the r value setting can be based on the unit cell axis lengths.

ChiCheng45:
**Description of work**
Created distance cutoff widget for the PDF job. The max r value is now set to the radius of the largest sphere that can be placed in the smallest unit cell of the whole trajectory. A more sensible 0.1 A step size is used. Removed hardcoded values from the rangewidget.

image

**Fixes**
Closes #462
Closes #431

**To test**
Run PDF and check that the results are good with the default settings. The results should be the same as the previous version except the max r value is set to a reasonable value.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/DistHistCutoffConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/DistHistCutoffConfigurator.py
@@ -16,7 +16,9 @@ class DistHistCutoffConfigurator(RangeConfigurator):
             A tuple of the range parameters.
         """
         if value[1] > floor(self.get_largest_cutoff() * 100) / 100:
-            self.error_status = "The cutoff distance goes into the simulation box periodic images."
+            self.error_status = (
+                "The cutoff distance goes into the simulation box periodic images."
+            )
             return
 
         super().configure(value)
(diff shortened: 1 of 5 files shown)


MBartkowiakSTFC (approved):
There are several good changes in this PR. Thanks for correcting the RangeConfigurator.

It worked fine for me.

PR #471, opened 2024-06-18
Minimal multiprocessing improvement

### LINKED ISSUE #361: [BUG] Mean square displacement job status goes over 100% complete

### LINKED ISSUE #447: [BUG] DISF calculation: CPU usage cannot be controlled

MBartkowiakSTFC:
**Description of work**
At the moment, multicore jobs work, but do not report progress correctly.
This PR changes the way progress is reported for multicore jobs.

closes #447 
closes #361

Note: this PR should let you pause and resume multicore jobs, but **terminating multicore jobs has not been enabled by this PR.**

**Fixes**
Set OMP_NUM_THREADS to 1 in the GUI launcher, suppressing numpy multithreading.
Removed threadpool from the list of running modes.
Added a method to Status for directly updating the progress of multicore jobs.
Removed the super().combine call from MeanSquareDisplacement.

**To test**
Run a job using multiple cores. Check if the progress is updated correctly.
Specifically, try running MeanSquareDisplacement anaylsis in single- and multicore modes. Progress should update correctly for both.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/RunningModeConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/RunningModeConfigurator.py
@@ -32,7 +32,7 @@ class RunningModeConfigurator(IConfigurator):
     specify the number of slots used for running the analysis.
     """
 
-    availablesModes = ["single-core", "threadpool", "multicore"]
+    availablesModes = ["single-core", "processpool", "multicore"]
 
     _default = ("single-core", 1)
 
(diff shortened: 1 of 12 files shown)


ChiCheng45:
For multicore jobs on my machine I still see the status jump from 0 to 100%.

ChiCheng45 (approved):
Looks good. I've tested MSD and the status updates correctly now for both the single core and multicore jobs. 

I've also tested on other jobs and it works as expected.

PR #476, opened 2024-06-25
documentation update

ChiCheng45:
**Description of work**
- Added PACF and IR job background sections.
- Corrected and cleaned up equations.
- Replace FCA section with a section on the correlation window used in MDANSE.
- Changed the trajectory converter section so that it lists the MD engines it supports instead.
- Added contents to all analysis theory sections.


MBartkowiakSTFC:
Looks good, definitely a step in the right direction.

Please check the following before we merge it:
1. In analysis.rst, line 231 there is a normalisation condition equation. What was it supposed to be? This dot over the 'equals' sign probably wasn't meant to be there.
2. In correlation.rst the dot symbol appears in several equations. Is it supposed to be multiplication? If so, should we just remove it?
3. in scattering.rst, line 170, there is a 'this consider' phrase which needs rewriting.
4. Also in scattering.rst, there are 5 references in total to 'target_CN'. Which section do you think it was supposed to be?
5. In scattering.rst, the equations 78 and 79 can't both be right. Most likely only one of them should have a tag 'isotropic system', but I am not sure if some other changes are needed for the anisotropic case. (Should there be a sum over n somewhere in eq. 79?)
6. trajectory.rst: PDB may not need to be on the list (your own observation).

ChiCheng45:
> Looks good, definitely a step in the right direction.
> 
> Please check the following before we merge it:
> 
> 1. In analysis.rst, line 231 there is a normalisation condition equation. What was it supposed to be? This dot over the 'equals' sign probably wasn't meant to be there.
> 2. In correlation.rst the dot symbol appears in several equations. Is it supposed to be multiplication? If so, should we just remove it?
> 3. in scattering.rst, line 170, there is a 'this consider' phrase which needs rewriting.
> 4. Also in scattering.rst, there are 5 references in total to 'target_CN'. Which section do you think it was supposed to be?
> 5. In scattering.rst, the equations 78 and 79 can't both be right. Most likely only one of them should have a tag 'isotropic system', but I am not sure if some other changes are needed for the anisotropic case. (Should there be a sum over n somewhere in eq. 79?)
> 6. trajectory.rst: PDB may not need to be on the list (your own observation).

1. Corrected
2. Corrected
3. Updated
4. I think this is supposed to link to the weight section in the documentation. I think we will have to update this when we get to updating the weights section.
5. I need to think about what this should be. Note that this was what it was in the original documentation.

image

6. Updated

MBartkowiakSTFC (approved):
Thank you for correcting these sections of the documentation. I think that we should merge the changes for now.

PR #477, opened 2024-06-26
Add cumulative time-averaged results to thermodynamic jobs

ChiCheng45:
**Description of work**
When we calculate the temperature and density they are the instantaneous temperatures and densities. To get the actual temperatures and densities we should time-average those properties, I have added a cumulative time-average results to the temperature and density jobs.

I think this should help with reading the temperature and density results since they fluctuate quite a lot. We can see in the below trajectory hasn't equilibrated until at least 2.5 ps.

image

image

**To test**
Run density (use NPT MD trajectory) and temperature jobs. Open results, cumulative average thermodynamic results be available.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/Density.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/Density.py
@@ -16,9 +16,10 @@
 
 import collections
 
+import numpy as np
 
 from MDANSE.Chemistry import ATOMS_DATABASE
-from MDANSE.Framework.Jobs.IJob import IJob, JobError
+from MDANSE.Framework.Jobs.IJob import IJob
 from MDANSE.Framework.Units import measure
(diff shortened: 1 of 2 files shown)


MBartkowiakSTFC:
Works great for temperature, but I got something unexpected with density.

avg_atomic_density
avg_atomic_density

The trajectory I used was **not equilibrated**, with a steep temperature ramp. (It was the molybdenum from the tutorial.) Should I expect it not to work correctly in such a case, or is there still a chance to make the method more general?

ChiCheng45:
This is expected. Its a cumulative average so that the average at, for example, 70 ps will be the average over all previous time steps.

MBartkowiakSTFC (approved):
The new part of the output should be useful to our users. We will have to add some comment to the documentation and tutorials about it.

PR #481, opened 2024-07-04
Numpy2 cleanup

### LINKED ISSUE #479: [BUG] Potential problems with Numpy 2
**Description of the error**
Since 16 June 2024, Numpy 2 is available on PyPI, and the number of users who have this version installed will gradually increase.

The list of problems that need to be addressed before we can switch to the new Numpy includes:

1. Import errors from `rdkit`. In the rdkit pyproject.toml, the requirement list includes "oldest-supported-numpy". Trying to import rdkit when Numpy 2 is installed results in messages like this one:
```
from rdkit.DataStructs import cDataStructs
AttributeError: _ARRAY_API not found
```
2. IOutputVariable inherits numpy.ndarray. The API changes at the moment result in just warning being issued, mainly due to the __array_wrap__ implementation in IOutputVariable not conforming to the requirements of Numpy 2. The full unit test suite of MDANSE at the moment produces over 180000 warnings.

It would be the best if we could smoothly switch to Numpy 2, before the old branch stops being supported.

N/A

**Suggested fix**
Temporary solution: restrict the numpy version in MDANSE requirements to <2.
Long-term solution: rewrite IOutputVariable so it does not inherit ndarray. It should still use the numpy.ndarray internally, but there is no need to inherit it.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Since the release of Numpy 2.0.0, the MDANSE unit tests have been producing large numbers of DeprecationWarnings. This PR corrects this. There are still errors output by rdkit, but they don't seem to cause immediate problems, and we expect them to be solved by rdkit developers eventually.

closes #479 

**Fixes**
1. Removed the __array_wrap__ from IOutputVariable.
2. Replaced numpy.row_stack with numpy.vstack.
3. Corrected the code in MoleculeFinder which tried to access the trajectory HDF5 file directly, and not by using the methods of the Trajectory wrapper class.

**To test**
All tests should pass without producing warnings.
Try running MoleculeFinder on a trajectory without molecule labels, to see if the InChI strings are assigned as molecule labels by rdkit.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/MoleculeFinder.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/MoleculeFinder.py
@@ -126,15 +126,10 @@ class MoleculeFinder(IJob):
         coords = conf.coordinates
 
         variables = {}
-        if (
-            "velocities"
-            in self.configuration["trajectory"]["instance"]
-            ._h5_file["configuration"]
-            .keys()
-        ):
+        if self.configuration["trajectory"]["instance"].has_variable("velocities"):
             variables = {
                 "velocities": self.configuration["trajectory"]["instance"]
-                ._h5_file["/configuration/velocities"][frameIndex, :, :]
+                .variable("velocities")[frameIndex, :, :]
                 .astype(np.float64)
             }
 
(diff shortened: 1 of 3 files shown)


ChiCheng45 (approved):
Looks good. I created a new env with numpy2 and I don't get any warnings in the unit tests. The molecule finder continues to work.

I also receive rdkit warnings as well.

PR #482, opened 2024-07-05
Add settings editor

### LINKED ISSUE #240: Physical units in the GUI
By popular demand, MDANSE should have a way of selecting the units it will use for different physical quantities.

There is no need to change the way units are handled internally, but on the level of the interface we need to enable at least the units of energy to be chosen according to the preference of the user. Typically we should expect the neutron scatterers to use meV for energy, while conventional spectroscopy is likely to pick cm-1.

The preferred units should be used both for input and plotting.

MBartkowiakSTFC:
**Description of work**
This PR is related to two related areas of the MDANSE_GUI code: persistent storage of user settings, and plotting parameters. LocalSession is replaced by StructuredSession, each GUI tab now has its own settings file, and plotters take their settings more or less directly from their settings file.

closes #240 

**Fixes**
1. Code of the previous plotter has been removed.
2. StructuredSession.py has been added, introducing TOML-based settings files.
3. PlotSettings widget now saves the selected values to the settings file.
4. UserSettingsEditor now edits different files, and has a combo box widget to switch between files.
5. GeneralTab class has additional methods for dealing with settings files.
6. Colour, line style and marker type can be manually changed for individual data sets in the plotter.

**To test**
Go through all the steps of trajectory analysis in the GUI, from conversion to plotting.
Change settings in the plotting tab. Confirm that the new values are still shown after restarting the GUI.


ChiCheng45:
Looks good. I changed the energy units and it stayed changed after I restarted MDANSE.

Some issues.
Units are not shown in the plotter when MDANSE is started for the first time (delete the settings to reproduce the error).
Also there are some issues with the plot range when the units are changed, see the animation below.

Animation24

The table is editable in the plot creator, but this does not do anything when plotted.
Animation25

On Windows, if you go into the color dialog, change the color, and then hit cancel the color it is still changed.

Animation26

ChiCheng45:
It looks good so far. Just a few more issues I've noticed.

- The line style and Marker options have duplicate blank or None selections.

image image

- Some unexpected behavior when changing the max/min percentile on the heatmap plots. Placing the minimum percentile above the maximum and changing it causes both the maximum and minimum to change.

Animation30


- The settings can be manually changed so that it is not a valid setting anymore, this can cause a crash later on.

Animation31


- The comments in the setting can be edited.

image

- Some matplotlib styles (e.g. fast) don't seem to work on my end.

Animation32

MBartkowiakSTFC:
I left the option of editing the comments in the settings files enabled, since the same can be done in a normal text editor.

Matplotlib-supplied options have been filtered to keep only the unique ones, and those that work.
The sliders are now coupled for Heatmap plots, but retain the old behaviour for other plotters.
In most places the GUI components which request settings do their own checks if the value they received is valid, and ask for a default value if the current one did not work. (This also adds a log warning about the current value being wrong).
Finally, the Plot Creator now informs the user that the plotting results are in another tab.

ChiCheng45:
I've stumbled across one another crash, see below. Everything else looks good though.

Crash when you deselect something in the grid or single view and select it in the heatmap view.
Animation33

ChiCheng45:
The reselection on the heatmap works without crashing now but the color range seems off. See the animation below. The entire plot is yellow, I have to change the min percentile so that it corrects itself.

Animation34

ChiCheng45 (approved):
Looks good, everything seems to work well.

PR #485, opened 2024-07-09
Disable spinbox in the RunningModeWidget on startup

### LINKED ISSUE #483: [BUG] The number of cores can be changed when single-core mode is selected.
**Description of the error**
The number of cores can be changed when the single-core running mode is selected.

**Additional details**
image

MBartkowiakSTFC:
**Description of work**
In the GUI, a 'single-core' running mode should only accept number of cores=1.

closes #483 

**Fixes**
Disabled the spin box of the RunningModeWidget on startup.

**To test**
In the GUI, load a trajectory and select an analysis type. You should not be able to change the number of cores from 1 when using the single-core mode.


ChiCheng45 (approved):
Looks good.

PR #486, opened 2024-07-09
Trajectory loading after conversion

### LINKED ISSUE #252: [ENHANCEMENT] GUI improvements
Following user feedback, we need to improve several features of the GUI.

- adjustments are needed to make the GUI fit in on smaller screens
- trajectory name should not contain the path to the file, only the filename
- Converters are not tracked by the Job Status widget, and they should be
- converted trajectories may be loaded automatically into the GUI after conversion
- a menu is needed to switch on and off different parts of the GUI
- the 3D Viewer is impossible to restore once closed; this must be corrected

Some of the changes are more arbitrary than the others, but there is no reason not to implement them all

**Additional context**
N/A

MBartkowiakSTFC:
**Description of work**
Following the request for this feature, a checkbox has been added which will make the GUI automatically load the results of a job upon completion.

closes #252 

**Fixes**
1. Added a checkbox to the Action widget.
2. Connected the job output to the Trajectory and Plot Selection tabs using the JobEntry class and JobHolder model.
3. Removed BackEnd.py and ConvertWizard.py, since they are no longer needed.

**To test**
Run a trajectory converter with the autoload box (next to the 'RUN!' button) checked and then unchecked. Check in the Trajectory tab if the trajectory has been loaded.
Run an analysis with the autoload box checked and then unchecked. In each case check if the Plot Creator tab loaded the output file.
In each case, the GUI is expected to load the results when the box is checked, and do nothing after the job has been completed if the box was not checked.


ChiCheng45 (commented):
Looks good, I've just added/corrected the log level check the in file configurators.

Just a few things I noticed.
- The average structure job does not output an mdt/mda file but has an auto-load result checkbox.
- Trajectory edit jobs e.g. moleculefinder outputs mdt files, checking the auto-load does not load this in the trajectories tab.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
```diff
@@ -246,8 +270,19 @@ def startProcess(self, job_vars: list):
         communicator = JobCommunicator()
         watcher_thread = JobThread(communicator, main_pipe, subprocess_ref)
         communicator.moveToThread(watcher_thread)
-
+        entry_number = self.next_number
+        item_th = JobEntry(
+            command=job_vars[0],
+            entry_number=entry_number,
+            pause_event=pause_event,
+            load_afterwards=load_afterwards,
```
ChiCheng45: Just noticed one more thing, the logs are no longer outputted to the job running tab. The first job entry above this one is the one that has all the logs. I think we need to merge the two JobEntry objects.

MBartkowiakSTFC:
Sorry that I missed the new lines creating the other JobEntry. Now, the logging should work again, auto-loading is disabled for AverageStructure, and trajectories output by analysis jobs get auto-loaded too if the box is checked.

ChiCheng45 (approved):
Looks good.

PR #488, opened 2024-07-10
Add logging to MDANSE and general refactoring

ChiCheng45:
**Description of work**

- Replaced all print and icecream calls with calls to logging - icecream dependency has now been removed.
- Added a new option in the outputfiles configurations so that they can now output log files.
- Remove unused imports.
- Updated some remaining file headers with the current version.
- Added logs textbox in the running jobs. This outputs INFO level or higher. Warning and errors are coloured.
- Generalised the TriplePanel to MultiPanel
- Updated checkablecombobox in the file selection widget. This is now the same as the checkablecombobox in the atom selection widget.

image

image

image

**To test**
Run a converter job with the log level selected. A log file should be outputted in the same directory as the MDT file but with a .log extension. Do the same with an analysis job. Some jobs might not output any logs. ASE and DOS produce some logs. Check that the logs are updated in the running tab. 

Try rerunning the jobs with "no logs" selected, no logs should be saved.

Also, check with the multicore setting.


MBartkowiakSTFC:
This is a big improvement throughout. Thank you for cleaning up the code and improving the GUI components. The MultiPanel layout and single implementation of the checkable combo box make the code more consistent.

One comment: you already added a method (add_log_file_handler) to IJob, which enables logging. Do you think that it would be possible to add another, which will wrap up the entire
```
        if self.configuration["output_files"]["write_logs"]:
            log_filename = self.configuration["output_files"]["root"] + ".log"
            self.add_log_file_handler(
                log_filename, self.configuration["output_files"]["log_level"]
            )
```
sequence, so we avoid repeating the same code? You could just have one method in IJob which checks which one of configuration["output_files"] or configuration["output_file"] is present, or have the Converter implementation of the method replace the "output_files" with "output_file". What do you think?

Otherwise everything is working as expected. Two things may have to be added **later**:
1. We will need to ensure that every job that fails outputs an error message. This is not guaranteed at the moment.
2. The GUI log viewer could benefit from an option of setting the logging level it should display. If we decide to do it, it should only happen after we have merged the changes to the settings files.

ChiCheng45:
Thanks for the suggestions. I've added the updated IJobs so it checks for the "output_files" key and adds the file handler.

```
    def initialize(self):
        if (
            "output_files" in self.configuration
            and self.configuration["output_files"]["write_logs"]
        ):
            log_filename = self.configuration["output_files"]["root"] + ".log"
            self.add_log_file_handler(
                log_filename, self.configuration["output_files"]["log_level"]
            )
```

all jobs now call `super().initialize()`. I've also replaced the converter settings keyword from "output_file" with "output_files".


And yes I agree both points (1) and (2) should be added in the future. For (2) we could put the GUI log level chosen in the settings.

MBartkowiakSTFC (approved):
Thank you, I think this PR is ready to be merged now.

PR #493, opened 2024-07-10
Save as script cancel fix

### LINKED ISSUE #491: [BUG] Crash when hitting cancel in the save script file browser.
**Description of the error**
Crash when hitting the cancel button in the save script file browser.

**Additional details**

Animation28

ChiCheng45:
**Description of work**
Fixes the save as script crash when the cancel button is clicked.

**Fixes**
Fixes #491

**To test**
In the converter or analysis jobs, click the save as script button and hit cancel the GUI should not crash. Save as script should continue to work as normal.


MBartkowiakSTFC (approved):
It solved the problem, thanks!

PR #494, opened 2024-07-11
Startup run converter fix

### LINKED ISSUE #490: [BUG] Run converter job not blocked on startup
**Description of the error**
When you start the GUI the run button is not blocked. This means that you can start an invalid converter job which will fail.

The run button should be blocked.

**Additional details**
Animation27

ChiCheng45:
**Description of work**
Fixes the bug where users were able to run invalid conversion jobs on startup.

**Fixes**
Fixes #490

**To test**
Start up MDANSE and check that the converter RUN button is blocked for invalid settings. Test that the conversion and analysis jobs work as usual.


MBartkowiakSTFC (approved):
Looks good.

PR #496, opened 2024-07-12
Job termination fixes

### LINKED ISSUE #472: [BUG] multiprocessing jobs cannot be aborted
**Description of the error**
At the moment it is not possible to abort a multiprocessing job from the GUI.

After "Terminate" has been clicked in the job's context menu, it should stop running as soon as possible.

The GUI shows the status as "Aborted", but the job keeps running in the background.

**Suggested fix**
An extra multiprocessing.Event object could be added to the inform the subprocesses that they should terminate.

**Additional details**
A lone subprocess seems to be created when a job is started from the GUI, and it exists even after the job has finished. While it does not consume any CPU time, it does take up some memory.

### LINKED ISSUE #495: [BUG] Crash on job termination and deletion from run table
**Description of the error**
GUI crashes when you terminate a job and then delete it from the run table.

**Additional details**
Animation29

ChiCheng45:
**Description of work**
Updated so that for multicore jobs its child processes are properly terminated. Fix crashing GUI when terminated jobs are deleted from the run table. Updated the max value of the QSpinBox for the multicore running mode.

**Fixes**
Fixes #495
Fixes #472

**To test**
Run jobs and then terminate them and delete them from the run table. The GUI should not crash. Check the number of python processes that are running, all the child processes should be killed. Check this for both singlecore and multicore running modes. Check jobs complete and results are the same with both singlecore and multicore jobs.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -137,6 +137,10 @@ class IJob(Configurable, metaclass=SubclassFactory):
 
         self._log_filename = None
 
+        self.inputQueue = Queue()
+        self.outputQueue = Queue()
+        self.log_queue = Queue()
+
     def __getstate__(self):
         d = self.__dict__.copy()
         del d["_processes"]
@@ -279,9 +283,9 @@ class IJob(Configurable, metaclass=SubclassFactory):
         if hasattr(self._status, "_queue_0"):
             self._status._queue_0.put("started")
 
-        inputQueue = Queue()
-        outputQueue = Queue()
-        log_queue = Queue()
+        inputQueue = self.inputQueue
+        outputQueue = self.outputQueue
+        log_queue = self.log_queue
 
         log_queues = [log_queue]
         handlers = []  # handlers that are not QueueHandlers
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -336,7 +336,9 @@ class IJob(Configurable, metaclass=SubclassFactory):
         listener : QueueListener
             The log listener that we need to stop.
         """
-        if not (hasattr(self._status, "_queue_0") and hasattr(self._status, "_queue_1")):
+        if not (
+            hasattr(self._status, "_queue_0") and hasattr(self._status, "_queue_1")
+        ):
             return
         if self._status._queue_1.qsize() > 0:
             if self._status._queue_1.get() == "terminate":
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -309,15 +309,19 @@ class IJob(Configurable, metaclass=SubclassFactory):
             p.daemon = False
             p.start()
 
-        while not inputQueue.empty():
+        n_results = 0
+        while n_results != self.numberOfSteps:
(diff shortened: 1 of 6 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -276,12 +276,11 @@ def process_tasks_queue(self, tasks, outputs, log_queues):
...
+        if hasattr(self._status, "_queue_0"):
+            self._status._queue_0.put("started")
 
-        ctx = multiprocessing.get_context("spawn")
-
-        manager = ctx.Manager()
-        inputQueue = manager.Queue()
-        outputQueue = manager.Queue()
+        inputQueue = Queue()
+        outputQueue = Queue()
```
ChiCheng45, line 283: On my computer, this change ensures that all child processes were terminated when requested from GUI. With the code before it there always seemed to be one last process left that was still running.

However, the new code causes issues with macos as Queue.qsize() is not implemented for some reason. The reimplemented Queue is added as suggested by on of the answers in https://stackoverflow.com/questions/65609529/python-multiprocessing-queue-notimplementederror-macos.

ChiCheng45: I guess the manager bit was used because qsize doesn't work with non-manager queues on macos. However, this seems to cause a leaked process when terminate is requested from the GUI.

The new version doesn't leak processes with terminate from the GUI but it means that a few parts of the code needed to be updated so that it avoided the call to qsize.

MBartkowiakSTFC:
So far things are going well. Jobs can be paused and resumed. They can be terminated both when paused and when running. Also, I don't see leftover Python processes when running jobs from the GUI.

The only thing that looks not quite right is this warning I get when closing the GUI after having run some multiprocessing jobs:
`/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 27 leaked semaphore objects to clean up at shutdown`
It seems to report 9 leaked semaphore objects per multiprocessing job **that got terminated**. (There are no leaked semaphores reported for a job that completed.)

ChiCheng45:
> So far things are going well. Jobs can be paused and resumed. They can be terminated both when paused and when running. Also, I don't see leftover Python processes when running jobs from the GUI.
> 
> The only thing that looks not quite right is this warning I get when closing the GUI after having run some multiprocessing jobs: `/opt/homebrew/Cellar/python@3.11/3.11.6_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 27 leaked semaphore objects to clean up at shutdown` It seems to report 9 leaked semaphore objects per multiprocessing job **that got terminated**. (There are no leaked semaphores reported for a job that completed.)

I don't get the warning on Windows but I'm guessing it's something to do with stuff still being in the job queues. I've made a commit which clears the job queues out. Hopefully, that should sort it out. Can you restest again?

MBartkowiakSTFC:
I still get the same message. I tried adding some extra lines like queue.close(), but that did not help.

This is a _warning_, not an error, and is only triggered on a Mac, when somebody starts a multiprocessing job and then aborts it. Maybe the occurrence of this problem is narrow enough, and consequences harmless enough, that we could just live with it for now...? We can decide tomorrow.

MBartkowiakSTFC (approved):
Now the pausing and aborting jobs works fine, both for single- and multicore jobs. The problem with leaked semaphores has been solved too. Looks good!

PR #497, opened 2024-07-15
Avoid using qsize to avoid macos crash

ChiCheng45:


ChiCheng45:
Not sure why but unit tests on the runner are taking too long. Tests work on my computer. Probably some issue somewhere.

PR #498, opened 2024-07-15
Add nightly test runs to github actions workflows

ChiCheng45:
**Description of work**
Run github actions every night for our test workflows. This is useful for cases when newer versions of libraries we depend on cause our tests to fail.

**To test**
N/A


MBartkowiakSTFC (approved):
As far as I can tell, this should work. Thanks for the changes.

PR #501, opened 2024-07-17
Add a test for pip install MDANSE_GUI

MBartkowiakSTFC:
**Description of work**
Recently a user encountered problems when installing MDANSE_GUI using pip. We should add a workflow that tries to install MDANSE_GUI on different platforms, just to see if `pip install` produces an error.

**Fixes**
A new test has been added, which only installs MDANSE and MDANSE_GUI.

**To test**
All tests must pass.

### DIFF
--- a/.github/workflows/test_pip.yml
+++ b/.github/workflows/test_pip.yml
@@ -1,4 +1,4 @@
-name: Test MDANSE
+name: Test pip install MDANSE_GUI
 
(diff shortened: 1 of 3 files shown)


ChiCheng45 (approved):
I've tested the pip install with a fresh conda env and tests pass and GUI loads up as expected.

PR #503, opened 2024-07-18
Add time-derivative configurator

ChiCheng45:
**Description of work**
- Previously the order of the derivative in infrared job was hard-coded. Added a time-derivative configurator so that the order can be selected.
- Updated the default variable of the interpolation order configurator to 3 like the GUI.
- Corrected a few file headers.

**To test**
Run the infrared job with different derivative orders, jobs should complete without any issues.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/DerivativeOrderConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/DerivativeOrderConfigurator.py
@@ -1,0 +1,59 @@
+#    This file is part of MDANSE.
+#
+#    MDANSE is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


MBartkowiakSTFC:
I tried it out and it worked. I used a short trajectory, so the results here are not really meaningful.
derivatives_IR

One thing I'd like to check with you is the naming and labels. Somehow when I see "numerical derivative order", my first thought is that I can choose to calculate the first, second or third derivative of the dipole autocorrelation function. But it is really the order of the polynomial that we choose here. Is there some other way we could label it, or did I misunderstand things?

ChiCheng45:
I've updated the labels and comments, let me know what you think.

MBartkowiakSTFC (approved):
I think that the meaning of the number is clear now, or at least there is little chance of confusing it for something else. Thanks for the change!

PR #505, opened 2024-07-18
GUI code cleanup

MBartkowiakSTFC:
**Description of work**
This PR mainly removes files which were added in the past but ended up not being necessary as other solutions were being gradually implemented.

**Fixes**
1. Removed InteractiveConverter,
2. Removed Dialog/Wizard code meant to be used with InteractiveConverter,
3. Deleted models from the prototype GUI (before the tabbed GUI was created),
4. Deleted unused widgets (oscillating progress bar),
5. Deleted parts of the Molecular Viewer which we don't use (e.g. residues database, since we have one in the main MDANSE already.)

**To test**
All tests must pass.
Try loading large structures (e.g. a protein) into the Molecular Viewer to check that nothing got broken when the residues database was removed.


MDANSE_GUI/Src/MDANSE_GUI/Scripts/mdanse_plotter.py
ChiCheng45, line 1: We will also need to delete the entry-point in the pyproject.toml if we delete this. I've done some testing and everything looks fine. Can merge if we decide to remove the entry-point or not.

MBartkowiakSTFC:
I cleaned up the entry points. `mdanse_job` was very outdated, and is gone now too.

ChiCheng45 (approved):
Looks good.

PR #506, opened 2024-07-19
Initial fix for HIS files

MBartkowiakSTFC:
**Description of work**
This PR makes the recent DMol3 trajectory load correctly, without affecting other sample trajectories.

**Fixes**
A test has been added to recognise a 1x1x1 AA^3 unit cell as not physical and assume that the system should have no periodic boundary conditions.

**To test**
Please try converting sample trajectories from Discover and the problem trajectory from DMol3.


ChiCheng45 (approved):
Looks good. I've tested it out on the test discover trajectory and the problematic DMol trajectory and it works fine.

PR #508, opened 2024-07-19
Update static structure factor default r and q values.

### LINKED ISSUE #473: [BUG] Change default r and q values for the static structure factor
**Description of the error**
As done for the PDF job (#470) the static structure factor default r and q values need to be updated so the results are meaningful.

Here is the static structure factor with the current default values (0 to 10 step 1 for both r and q).
 
image

Here is what I get when I change the step size for both q and r to 0.01 for example.
image

### LINKED ISSUE #474: [BUG] Why are there large negative values of the static structure factor as q->0
**Description of the error**
Why does the static structure factor have a large negative value for q->0. Here is one I generated using a cp2k water trajectory. I'd imagine there are numerical issues as q=0 but the drop to a large negative value begins before it gets close to 0.

image

Here is an experimental static structure factor of liquid Ar (https://journals.aps.org/pra/abstract/10.1103/PhysRevA.7.2130).

image

it goes to zero as q->0.

ChiCheng45:
**Description of work**
Here are some results for an Ar trajectory.

PDF
image

SSF - 10 ang cutoff 
image

SSF - 20 ang cutoff 
image

SSF - 30 ang cutoff 
image

SSF - 40 ang cutoff 
image

SSF - 50 ang cutoff
image

We can see that if we go beyond 1/2 the smallest unit cell length (33.8 ang) then we get nonsensical results. This change updates the static structure factor calculation setting so that the max r values are set to 1/2 the smallest unit cell length. Default Q values are updated to something reasonable.

**Fixes**
Fixes #474
Fixes #473

**To test**
Run the static structure factor job. Should complete without any issues.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/RangeConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/RangeConfigurator.py
@@ -41,6 +41,7 @@ class RangeConfigurator(IConfigurator):
         toList=False,
         mini=None,
         maxi=None,
+        default=None,
         **kwargs,
     ):
         """
@@ -75,6 +76,8 @@ class RangeConfigurator(IConfigurator):
         self._mini = mini
 
         self._maxi = maxi
+
+        self._new_default = default
 
     def configure(self, value):
         """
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
The results are looking good, flattening out close to 0 for low q. I have checked that I cannot set the maximum r to a value higher than the limit, which is also good.

One (minor) thing that worries me is that I am still able to run the analysis for both q and r set to a range from 0 to 0, which gives me a single point at q=1 with the value 1. But this is the case also for the main branch, so it is not a problem with this PR specifically.

PR #509, opened 2024-07-22
Correct path saving mechanism for actions

MBartkowiakSTFC:
**Description of work**
Last improvements to the code consistency before the second alpha release.

**Fixes**
1. Corrected the path saving mechanism, so it does not save the path if the user cancels the file dialog.
2. Replaced several `print` statements with MLogger equivalents.
3. Limited the maximum PyQt version to 6.7.0 (preventing the problem with PyQt 6.7.1 depending on glibc 2.35).

**To test**
1. Pick some input files for a converter, e.g. `/some/path/to/files/input.txt`
2. Now try to pick another file, but cancel the file dialog instead of loading the file.
3. Open the dialog again. The dialog should open in `/some/path/to/files/`.


ChiCheng45 (approved):
Looks good.

PR #514, opened 2024-07-23
Forward the traceback from jobs to the logger

MBartkowiakSTFC:
**Description of work**
Now that each job can have its own log file, it is important that messages connected to the job failing is 

**Fixes**
Added extra LOG lines to IJob.
In IJob.run() forwarded the traceback to the log if an exception is raised.

**To test**
Run a job with the logging level "INFO" or "DEBUG" and give it input that will make it fail. (E.g. converter with a wrong input file). The error output should appear at the end of the log file.


ChiCheng45 (approved):
Changes Look fine. I tried running MDANSE in script mode with bad inputs and I noticed that it doesn't write a log file out. It crashes before the file handlers are created. I've opened #519 for this.

PR #515, opened 2024-07-23
Change filters for output file names in the GUI

MBartkowiakSTFC:
**Description of work**
At the moment on a Mac the file names in the GUI fields end up getting '.*' appended at the end when filled out via the QFileDialog. This PR prevents it from happening.

**Fixes**
Changed filters in file widgets.

**To test**
Try putting some output file names into the GUI with and without the correct extension.


ChiCheng45 (approved):
I've tested on linux and everything works as expected. Now on windows the extension get automatically added to the filename.

I notice an related issue with file auto-loading. I've opened #521 for this issue.

PR #517, opened 2024-07-23
Give the developer version a distinct number

MBartkowiakSTFC:
**Description of work**
To make the versioning more clear, we give the main branch a different logo and version numbers, so the "About MDANSE" dialog makes it clear that the user is not using an official release.

**Fixes**
1. Changed the splash screen.
2. Replaced the 'a2'/'a3' suffix with a '-dev'

**To test**
If you are using an editable install, you may have to re-install for the version number to be updated.
Start the GUI. Check if the splash screen has been updated. Use 'Help -> About MDANSE' to confirm that a different version number is shown.


ChiCheng45 (approved):
Looks good.

PR #522, opened 2024-07-24
Add file extension on autoload if missing

### LINKED ISSUE #521: [BUG] Autoloading trajectory or results when the extension is put in the filename.
**Description of the error**
When the file extension is added to the outputfile. The file is not autoloaded. The file is written correctly and can be loaded manually.

**Additional details**

Animation37

MBartkowiakSTFC:
**Description of work**
At the moment, the automatic loading of results would fail or work depending on if the GUI input field contained file name with or without the extension, respectively.

closes #521 

**Fixes**
1. Added a check in JobEntry to add the extension only if it is missing.

**To test**
Run a converter and an analysis job twice, once with a name that contains the extension, and once without the extension. The results should be loaded automatically in both cases.


ChiCheng45 (approved):
Looks good.

PR #526, opened 2024-07-29
Energy spectrum

RobBuchananCompPhys:
**Description of work**
Job to calculate energy spectrum of atomic trajectories in a similar way to how we currently perform density of states, only the FFT of the position AC function is used in this case.

**Fixes**
None

**To test**
Tests still to come


MBartkowiakSTFC:
Please have a look at the code of the Position Autocorrelation Function (PACF) here:
https://github.com/ISISNeutronMuon/MDANSE/blob/eb282d6abbcee3843fa66ac97f0338c2c8e4f47e/MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py#L117
It is important that it does **not** call the `differentiate` function. By differentiating the positions we get velocities, and the result ends up being the Velocity Autocorrelation Function.

PR #527, opened 2024-07-29
Log viewer tab

MBartkowiakSTFC:
**Description of work**
Now that we log messages in different parts of the code, it would make sense to show the log messages in the GUI

**Fixes**
1. Reactivated the Logger tab.
2. Connected the LOG from MLogging to the Logger tab.
3. Added a combo box for setting the log level.

**To test**
Start the GUI and load some analysis results for plotting. The log messages should be shown in the logger window.
Try changing the log level to see if the log messages are filtered correctly.


ChiCheng45 (approved):
Looks good.

PR #535, opened 2024-08-09
Reset camera after changing trajectory.

### LINKED ISSUE #528: [BUG] 3D render distance for very large systems
**Description of the error**
For very large systems the 3D viewer does not render the system properly. This mainly occurs with the parallel projection mode. 

You can reproduce the error by first loading a small system setting the projection to parallel and then loading a much bigger system.

**Suggested fix**
Increase the VTK render distance depending on the system size.

**Additional details**
Animation39

ChiCheng45:
**Description of work**
Reset the camera when the trajectory is changed. This is helpful when working with systems of different sizes so that the user doesn't need to zoom in or out after switching trajectories.

Before:
Animation45

After:
Animation44

**Fixes**
Fixes #528

**To test**
Try loading system sizes of different size and switch between them. 3D viewer should reset the camera e.g. zoom out if it switches to the bigger trajectory. Test the 3D viewer in the atom selection dialog.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -135,6 +135,8 @@ class MolecularViewer(QtWidgets.QWidget):
         self._colour_manager = AtomProperties()
         self.dummy_size = 0.0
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
I was able to reproduce the original problem. This PR fixed it, and I did not run into any other difficulties. Looks good.

PR #536, opened 2024-08-09
Frame setting updates animation while it still running.

### LINKED ISSUE #329: [ENHANCEMENT] Allow interactive animation setting changes whilst its running.
When an animation is running we should be able to adjust the 3D view interactively. Currently we need to pause and then change certain settings whilst its running, for example, the time per frame setting. 

Allow animation to be updated whilst its running for all possible setting changes.

**Additional context**

Animation7

ChiCheng45:
**Description of work**
Updated the time per frame and show every N frames setting so that they work while the animation is running. The means that the user doesn't need to stop and start the animation everytime they change those settings.

**Fixes**
Closes #329 

**To test**
Load a trajectory and run the animation, change the time step and frame skip. The animation speed should change while it's still running. Check this is working in the atom selection dialog.


RobBuchananCompPhys (approved):
Tested and approved

PR #537, opened 2024-08-12
Make the sum of partial terms equal to the total for scattering calculations

ChiCheng45:
**Description of work**
Given that

image

the total $F(q, t)$ and $S(q, \omega)$ should be a sum of partial terms. In this PR I have updated the DynamicCoherentStructureFactor, DynamicIncoherentStructureFactor, ElasticIncoherentStructureFactor, and GaussianDynamicIncoherentStructureFactor so that the total is a sum of the partial terms.

This makes the results more clear. For example, with the water trajectory and the DISF, EISF, and GDISF calculations below, we can see that the oxygen atoms with b_inc2 = 6.37e-13 do not contribute much to the total since hydrogen atoms have b_inc2 = 6.39e-8.

Water trajectory
**DCSF** 
Before
image
After
image

**DISF**
Before
image
After
image

**EISF**
Before
image
After
image

**GDISF**
Before
image
After
image

**To test**
Run the DynamicCoherentStructureFactor, DynamicIncoherentStructureFactor, ElasticIncoherentStructureFactor, GaussianDynamicIncoherentStructureFactor, and DOS calculations. All total results should be a sum of the partial terms.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/DensityOfStates.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/DensityOfStates.py
@@ -238,16 +238,21 @@ class DensityOfStates(IJob):
             )
 
         weights = self.configuration["weights"].get_weights()
-        weight(
+        self._outputData["vacf_total"][:] = weight(
             weights,
             self._outputData,
             nAtomsPerElement,
             1,
             "vacf_%s",
-            update_values=True,
+            update_partials=True,
         )
-        weight(
-            weights, self._outputData, nAtomsPerElement, 1, "dos_%s", update_values=True
+        self._outputData["dos_total"][:] = weight(
+            weights,
+            self._outputData,
+            nAtomsPerElement,
+            1,
+            "dos_%s",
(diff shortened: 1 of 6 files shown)


gonzalezma:
I think that having the "non-weighted" partials is still useful (e.g. to compare the dynamics of different atoms in the same scale or to recalculate a total S(Q,w) using a different weighting scheme without rerunning the MDANSE analysis). I would propose to return an output similar to that of the total dynamic structure factor analysis: the total, the non-weighted partials (<exp(iQr(0))exp(iQr(t)>) and the weighted partials (b^2<exp(iQr(0))exp(iQr(t)>).

MBartkowiakSTFC:
Would you accept the solution where the scaling of the partials (for the purpose of comparison) is the task of the plotter? I will have to add normalisation options to the plotter anyway.

gonzalezma:
Yes, I think this could be a good solution.

MBartkowiakSTFC (approved):
Looks good. We will address the relative scaling in the plotter.

PR #538, opened 2024-08-21
Position power spectrum

RobBuchananCompPhys:
**Description of work**
Job to calculate power spectrum spectrum of atomic trajectories in a similar way to how we currently perform density of states, only the FFT of the position AC function is used in this case. Input to a low-pass filter for atomic vibrational modes.

**Fixes**
None

**To test**
Tests still to come


MBartkowiakSTFC (approved):
Mo_metal_DOSvsPPS
The results are looking good. I attached a comparison of DOS and PPS for liquid molybdenum. The two are different, as expected.

Issue #530, opened 2024-08-02
[BUG] How should we deal with the total results when the weight normalization is close to zero or zero?


ChiCheng45:
**Description of the error**
When the sum of all weights is zero, the weight factors are not normalized. Is this the best way to to handle this? For example you can get some odd behaviour by changing the concentrations slightly.

Here I have an argon trajectory where I have transmuted the atoms to 1024 Test3 and 1024 Test4 with bcoh_test3 = 0.2 and bcoh_test4 = -0.2. The denominator in the weights will be zero. The DCSF of this is,

![Image](https://github.com/user-attachments/assets/6bd7abbd-5492-43c7-a7d0-84d6166a20bd)

Now I change the number of atoms to 1023 Test3 and 1025 Test4,

![Image](https://github.com/user-attachments/assets/e75410c1-d1f7-4841-b74d-1206348ec649)

so that the normalization is turned on and the DCSF results are an order a magnitude larger even though they are effectively the same system.

oerc0122:
Potentially fixed by #1279

PR #540, opened 2024-08-23
Text output in the plotter

MBartkowiakSTFC:
**Description of work**
To make it easier for the users to transfer their results to other software (e.g. for fitting), we add another view to the Plot Holder. It is a text viewer, which shows the analysis results as plain text.

**Fixes**
1. Improved the logger handling of exceptions and error messages.
2. Created DataWidget as an equivalent to PlotWidget.
3. Created Text as subclass of Plotter, to be used with DataWidget.

**To test**
Please load analysis results with 1D and 2D arrays (e.g. DOS and DISF). Make sure that they can be plotted.
Create a text output using the new button in plot selector.
"Plot" the results in the new tab. Check the following actions:
1. Changing the units of the axes.
5. Selecting/deselecting datasets.
6. Changing other plot properties in the dataset table on the left.
Plot properties should not affect the text output.

### DIFF
--- a/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
+++ b/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
@@ -37,8 +37,9 @@ def get_weights(props, contents, dim):
         else:
             normFactor += fact
 
-    for k in list(weights.keys()):
-        weights[k] /= np.float64(normFactor)
+    if abs(normFactor) > 0.0:  # if normFactor is 0, all weights are 0 too.
+        for k in list(weights.keys()):
+            weights[k] /= np.float64(normFactor)
 
     return weights, normFactor
 
--- a/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
@@ -107,6 +107,9 @@ class TabbedWindow(QMainWindow):
         self._tabs["Plot Creator"]._visualiser.create_new_plot.connect(
             self._tabs["Plot Holder"]._visualiser.new_plot
         )
+        self._tabs["Plot Creator"]._visualiser.create_new_text.connect(
+            self._tabs["Plot Holder"]._visualiser.new_text
+        )
 
     def createCommonModels(self):
         self._trajectory_model = GeneralModel()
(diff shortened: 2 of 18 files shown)


RobBuchananCompPhys (commented):
Manual testing looks good, made the following suggestions.
For the text view holder
- CheckBox to display the text view as a "preview"
- SpinBox that allows for the precision of the data to be dynamically incremented/decremented
- ComboBox allowing user to change units of y-axis values
- Export button (i.e. text view to .csv)

For the Actions tabs
- Remove lattice QVectors since it is a base class and not intended for use

For the graphical plot holder
- Disable x and y offset sliders in the case of 1D Data plotting

@MBartkowiakSTFC

MBartkowiakSTFC:
There are some improvements following the review:
1. The DataWidget in the plotter text view now has a number of controls for changing formatting. Also, the mode is "preview" by default, so large volumes of text do not appear unless explicitly requested,
2. There is a CSV output option in the DataWidget,
3. Error message handling in the GUI has been improved, and more errors are caught by the logger now instead of crashing the GUI,
4. Related to issue #530 if the sum of weights is 0, normalisation by the sum is **not applied**. This prevents the output files from being filled with NaN values. Now they will be filled with 0 instead.

MBartkowiakSTFC:
@ChiCheng45 I prioritised this PR because it makes it easier for users to view and export the data from the GUI, which could be useful in the workshop tutorials. The stability improvements could be important for the same reason. Your opinion is welcome, especially since the tutorials were written almost exclusively by you.

RobBuchananCompPhys:
I've done manual testing of the new features, they look good and are working well.
Just one suggestion made about simplifying the of the `enable_sliders()` method.

ChiCheng45 (commented):
Looks good. Just a few things.

What are the differences between options in the save file?
image

Can we change the csv output so that there isn't a newline after each row? In excel there is a empty row between each row of data.
image

MBartkowiakSTFC:
I added extra labels in the GUI to explain what the fields are for.
I made a change to the text output, which did not make a difference on my computer, but was intended to remove the extra lines on Windows. Did it work?

ChiCheng45:
> I added extra labels in the GUI to explain what the fields are for. I made a change to the text output, which did not make a difference on my computer, but was intended to remove the extra lines on Windows. Did it work?

Thanks, that has sorted that out.

ChiCheng45 (approved):

PR #541, opened 2024-09-04
Remove explicit PyQt6 imports

MBartkowiakSTFC:
**Description of work**
We still had two lines importing an object from PyQt6 instead of qtpy. I think that the IDE inserts those automatically, that's why they are easy to overlook. If present, these will make it more difficult to switch between PyQt versions.

**Fixes**
- removed the two import lines referring to PyQt6.

**To test**
MDANSE_GUI should start normally using
`mdanse_gui`
but also if PyQt5 is installed and
`QT_API=pyqt5 mdanse_gui`
is used.


ChiCheng45 (approved):
Looks good.

PR #543, opened 2024-09-10
Post-workshop improvements

MBartkowiakSTFC:
**Description of work**
The recent live demonstration of MDANSE has uncovered several unrelated minor problems. A few of them are addressed here.

**Fixes**
1. ASE converter will now reuse the initial unit cell definition if the input file (tried on a PDB file) only contains the definition for the first frame,
2. TrajectoryWriter is now able to resize the datasets to the number of frames that were **successfully** read. This is at the moment only used by the ASE converter (which is most likely to predict the number of frames wrong).
3. Some error messages have been relabelled as warnings, most notably the missing configuration file messages,
4. StructuredSession has been modified to reduce the likelihood of it keeping many files opened at the same time.

**To test**
All tests must pass.
Please try to convert an openmm PDB trajectory using ASE converter. The resulting trajectory should have the correct number of frames and a unit cell defined at all steps.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/AseInputFileConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/AseInputFileConfigurator.py
@@ -102,6 +102,8 @@ class AseInputFileConfigurator(InputFileConfigurator):
         try:
             val = self["value"]
         except KeyError:
-            LOG.error(f"No VALUE in {self._name}")
-
-        return "Input file: %r\n" % self["value"]
+            result = f"No VALUE in {self._name}"
+            LOG.error(result)
+            return result
+        else:
+            return "Input file: %r\n" % self["value"]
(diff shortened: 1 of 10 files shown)


ChiCheng45 (commented):
Looks good. Almost all of the pdb trajectory from openmm I've tried out seems to work fine.

I've found one example where the configuration seems to fail. This was the pdf trajectory (traj.pdb) generated from https://github.com/openmm/openmm_workshops/blob/main/section_1/protein_ligand_complex.ipynb. Not sure if this is an issue with ASE or MDANSE.

image

MBartkowiakSTFC:
The file fails because it has more molecules than ASE can handle.

I created an issue here:
https://gitlab.com/ase/ase/-/issues/1539
to see if anything can be done about it on the ASE side.

ChiCheng45 (approved):

PR #544, opened 2024-09-12
Add new job which calculates the distinct part of the Van Hove function.

ChiCheng45:
**Description of work**
The Van Hove function is related to the intermediate scattering function via a Fourier transform and the dynamic structure factor via a double Fourier transform.

This PR adds the distinct part of the Van Hove function and is spherically averaged so that it is a function of r=|$`\vec{R}`$| and normalized so that $`G(r, t) \rightarrow 1`$ as $`r \rightarrow \infty`$ for liquids or gases. Also, this will mean that $`G(r, 0) = \mathrm{PDF}(r)`$ and will be consistent with the $\mathrm{PDF}(r)$ calculated in MDANSE. Similarly to the PDF calculations, inter and intramolecular parts will be separated.

See Fig. 10 in  https://journals.aps.org/pre/abstract/10.1103/PhysRevE.51.4626 for an example of the distinct part of the Van Hove function in the literature. Here is an example calculated with the new job using the Ar36 trajectory.

image

We can see that the correlation hole from around $r=0$ to $3$ begins to fill up as time evolves. Eventually as  $`t \rightarrow \infty`$,  $`G(r, t) \rightarrow 1`$.

A job which calculates the self part of the Van Hove function and documentation will be added in later PRs.

**To test**
The cython distance histogram function was generalized for the Van Hove function calculations. Calculate the PDF on protos and save the results. Switch to this branch and rebuild MDANSE, calculate the PDF, and compare against the previous results, results should be the same. Run a Van Hove Function Distinct job and check the results, the G(r,t=0) result should be the same as PDF(r).


MBartkowiakSTFC (changes requested):
The calculation itself is working well. There are no changes to the PDF results, while the van Hove result agrees with the PDF at t=0.

Please add a docstring to the van Hove analysis, so an explanation of the analysis can be seen in the GUI. Also, I would be in favour of moving it from 'scattering' to 'dynamics', mainly because the result is in direct units (time, distance) and not reciprocal units (E,q).

ChiCheng45:
> The calculation itself is working well. There are no changes to the PDF results, while the van Hove result agrees with the PDF at t=0.
> 
> Please add a docstring to the van Hove analysis, so an explanation of the analysis can be seen in the GUI. Also, I would be in favour of moving it from 'scattering' to 'dynamics', mainly because the result is in direct units (time, distance) and not reciprocal units (E,q).

Thanks for the review, I've moved the job to dynamics and added a description.

MBartkowiakSTFC (approved):
Looks good! Thanks for the changes.

Issue #545, opened 2024-09-13
[BUG] CenterOfMassesTrajectory changes elements and density


MBartkowiakSTFC:
**Description of the error**
When running CenterOfMassesTrajectory analysis on a DFTB water trajectory (from MDANSE-Examples), the resulting trajectory contains only hydrogen atoms.

**Describe the expected result**
I would expect the following results:
1. If grouping="atom", the output trajectory should be identical to the input,
2. If grouping="molecule", the output trajectory should contain one object per molecule. I would expect it to be an artificial atom with the same mass at the mass of the original molecule. 

**Describe the actual result**
Instead, I got the following:
1. With grouping="atom", from 200 H and 100 O atoms on input I got 300 H atoms in the output.
2. With grouping="molecule" I got 100 H atoms. The number of H atoms is the same as the number of water molecules.

![density_COM](https://github.com/user-attachments/assets/786694d3-7eb8-470a-8634-b3edc73b7d1e)
The figure compares the mass density of original water, molecule-grouped COM water and atom-grouped COM water.

**Suggested fix**
Trajectory should have its own built-in atom database containing only the elements present in the system. This will make trajectories fully transferable and enable replacing molecules will artificial atoms that will conserve the mass of the system.

**Additional details**
Even though it is possible to do statistics on molecule positions using the current implementation of COM trajectory, it will also become impossible to distinguish different molecules if they all get replaced with hydrogen.

PR #546, opened 2024-09-13
Store charge information in trajectory

MBartkowiakSTFC:
**Description of work**
In some cases, the information about the partial charges is already contained in the MD engine files. This can be stored in the trajectory file and used for calculations of charge-related properties.

PR https://github.com/ISISNeutronMuon/MDANSE/pull/518 introduces a Trajectory Editor. That will have to be adjusted to handle charges once that PR is merged.

No attempt has been made to handle charges in CenterOfMassesTrajectory, since that one will require more fundamental changes (see https://github.com/ISISNeutronMuon/MDANSE/issues/545 ).

**Fixes**
1. Several (but not all) converters attempt to find charge information in the input files.
2. The charge values are stored in the MDT trajectory,
3. If the charge values change between frames, the information about charges is stored for each frame,\
4. Infrared analysis reads the charges from the trajectory file, **unless** the user specified charges using charge configurator.

**To test**
All tests must pass.
Please try converting a LAMMPS trajectory or DL_POLY trajectory which contains charges. Infrared analysis should produce a non-zero signal even when charges are not defined. Setting the charges using the charge configurator should change the results.
Analysis types which output a trajectory as a result should not discard the charge information.
Behaviour for trajectories without charge information should be the same as before.


ChiCheng45 (commented):
I've tried it out on a DLPOLY water trajectory and it seems to have produced an IR spectrum when the atom charges setting is an empty dict as expected.

I think we need to add an update to clear up the default behavior. In the latest commit in this PR, when the partial charge helper is used and a charge of zero is set then the charge setting dict does not include that atom idx. The charge of that atom will not be zero but the charge from the trajectory. I think we should update the charge helper so that it specifically sets zeros now. 

Perhaps we should also change the default behavior so that when any of the atoms are set to have a charge in the setting config then all others that are not set should be set to zero. In other words the charges from the trajectory are only used when the charge setting is an empty dict.

MBartkowiakSTFC:
I made changes to the PartialChargeConfigurator. Now the charges should be taken from the trajectory, except for the atoms where the user set the charge manually. Also, the charge dialog in the GUI shows the initial configuration at frame 0.

@ChiCheng45 It seems to me that part of the information about the dipole moment was lost in the loop around here https://github.com/ISISNeutronMuon/MDANSE/blob/5e3d36ac2d9c796bfb35931d43bc3a50715335c3/MDANSE/Src/MDANSE/Framework/Jobs/Infrared.py#L161
I changed it to do `+=` instead of `=`. Please check if this is the desired behaviour. From my point of view, I get the correct results now.

ChiCheng45:
> I made changes to the PartialChargeConfigurator. Now the charges should be taken from the trajectory, except for the atoms where the user set the charge manually. Also, the charge dialog in the GUI shows the initial configuration at frame 0.
> 
> @ChiCheng45 It seems to me that part of the information about the dipole moment was lost in the loop around here
> 
> https://github.com/ISISNeutronMuon/MDANSE/blob/5e3d36ac2d9c796bfb35931d43bc3a50715335c3/MDANSE/Src/MDANSE/Framework/Jobs/Infrared.py#L161
> 
> 
> I changed it to do `+=` instead of `=`. Please check if this is the desired behaviour. From my point of view, I get the correct results now.

Yes, this should have been '+=' and the default operation when the charge mapping setting is used is better now. I can approve once the merge conflict is fixed.

ChiCheng45 (approved):

PR #548, opened 2024-09-16
Visual improvements to GUI

MBartkowiakSTFC:
**Description of work**
Minor improvements to the GUI appearance are suggested. It is mainly following up on the observation that we don't really preview colours, only the colour hex string.

**Fixes**
1. The Running Jobs tab now has a progress bar in the progress field.
2. Colour selectors in trajectory view atom list and in the plotter now show the current colour as the background colour of the corresponding table item.

**To test**
Load two trajectories and check if the atom colours are updated when switching between trajectories, and when changing the colours using the dialog.
Load analysis results and plot several curves. Check if the colour entry in the table has the correct colour, also when switching matplotlib styles.


ChiCheng45 (approved):
Works great. While I was testing I noticed some behaviour issues with the atom properties setting and plotting that existed previously. I've created #569 and #567 for these.

PR #549, opened 2024-09-17
Added self part of van Hove function job and updated documentation theory

ChiCheng45:
**Description of work**
Added the self part of the van Hove function and updated the theory part of the documentation.

Here is the self part of the van Hove function for an Ar36 trajectory. We can see the argon atoms diffusing through the system as time progresses.
image

Here is an example with the lammps trajectory
image

we can see that the Cu atoms (top left) diffuse through the system via specific sites while the S and Sb atoms are fixed in the lattice.

**To test**
Rebuild MDANSE and run the VanHoveFunctionSelf job and check the results. See Figs. 5 and 6 in https://journals.aps.org/pre/abstract/10.1103/PhysRevE.51.4626 for some examples.

### DIFF
--- a/Doc/pages/dynamics.rst
+++ b/Doc/pages/dynamics.rst
@@ -13,6 +13,7 @@ This section contains background theory for following plugins:
 -  :ref:`analysis-msd`
 -  :ref:`analysis-op`
 -  :ref:`analysis-pacf`
+-  :ref:`analysis-vhf`
 -  :ref:`analysis-vacf`
 
 .. _analysis-angular-correlation:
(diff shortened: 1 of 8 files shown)


MBartkowiakSTFC (commented):
The calculation is working well. I tried it on a few examples and had no problems.

The only issue I see is the fact that the two parts of the van Hove function (self- and distinct) are normalised differently. It would be good if they could both be on the same scale. Since there are good reasons to normalise the self part the way it is already done now, could you:
- normalise the distinct part by the same factor,
- save the normalisation factor in the output in case we wanted to undo it later?

ChiCheng45:
> The calculation is working well. I tried it on a few examples and had no problems.
> 
> The only issue I see is the fact that the two parts of the van Hove function (self- and distinct) are normalised differently. It would be good if they could both be on the same scale. Since there are good reasons to normalise the self part the way it is already done now, could you:
> 
> * normalise the distinct part by the same factor,
> * save the normalisation factor in the output in case we wanted to undo it later?

I think we should save the normalization so that it can be applied or removed while plotting. A similar feature was mentioned in #537 but for the weighting factor. Can we do this in another PR? It would be a larger change and then we can do both the normalization and weighting for all calculations, I've created #550 for this feature request.

MBartkowiakSTFC:
Yes, we need to extend the plotter to add different normalisation options anyway. I'm fine for it to be implemented in a separate PR. Does it mean that now both of the van Hove functions will be output without being normalised?

ChiCheng45:
> Yes, we need to extend the plotter to add different normalisation options anyway. I'm fine for it to be implemented in a separate PR. Does it mean that now both of the van Hove functions will be output without being normalised?

Currently, they are both normalized in different ways. Currently, the self and distinct part of the van Hove function are implemented so that they follow what is done in the plots in https://journals.aps.org/pre/abstract/10.1103/PhysRevE.51.4626.

PR #551, opened 2024-09-19
Fix MDANSE when using files with non-standard characters

### LINKED ISSUE #547: [BUG] Actions tab issue with trajectory files with filenames which contain emoji characters
**Description of the error**
A trajectory with a filename containing emoji characters (e.g. 🧲🧲🧲.mdt) can be loaded up and is visible in the 3D view,

image

however when I move to the actions tab and select a job, the settings part of the actions tab does not load up.

image

ChiCheng45:
**Description of work**
Allow MDANSE to be used with files with names with non-standard characters.

**Fixes**
Fixes #547

**To test**
Save an MDT trajectory file with a non-standard character filename, run a job with that trajectory, and save the MDA file with a file name with non-standard characters. Plotting and action jobs should run with no issues.

### DIFF
--- a/MDANSE/Src/MDANSE/Core/Platform.py
+++ b/MDANSE/Src/MDANSE/Core/Platform.py
@@ -197,12 +197,7 @@ class Platform(object, metaclass=abc.ABCMeta):
         :return: the normalized and absolute version of the input path
         :rtype: str
         """
-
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
Thanks for fixing this! It works fine.

PR #553, opened 2024-09-20
Update the eccentricity calculation so that it follows more standard equations

### LINKED ISSUE #378: [BUG] Eccentricity weight setting
The weight setting in the eccentricity job seems to replace the mass in the moment of inertia calculation with the weight setting value so that it might not be a moment of inertia anymore. It also means that the weighting procedure does not follow what is stated in the documentation. Additionally no weighting is applied to the radius of gyration calculation.

**Suggested fix**
Remove the weight setting and change the moment of inertia so that it uses the atomic weight not the weight setting value.

**Additional details**
Here is an example where I set the weight setting to charge. The moment of inertia is zero and additionally the units are wrong in the plot.
image

### LINKED ISSUE #380: [BUG] Why is the Radius of gyration calculated using the Eccentricity job not the same as the one from the RadiusOfGyration job?
**Description of the error**
The ROG calculated in the Eccentricity job is different to the one calculated in the RadiusOfGyration job. Is this expected? I would've expected them to be the same.

**Additional details**
ROG from eccentrictiy job
image

ROG from RadiusOfGyration Job
image

### LINKED ISSUE #552: [BUG] Eccentricity has an orientational dependence
**Description of the error**
The eccentricity calculation in its current form depends on the orientation of the molecule. Here I calculate the eccentricity of a single tip4p water molecule which is rigid.

image

Since it's rigid and its shape does not change, its eccentricity should be constant.

**Suggested fix**
We should change the eccentricity job to follow the definition in rdkit https://www.rdkit.org/docs/source/rdkit.Chem.Descriptors3D.html.

sqrt(pm3^2 -pm1^2) / pm3

where pm3 and pm1 is the largest and smallest principal moment of inertia. This equation will change depending on the shape of the system but not on the orientation of the system.

ChiCheng45:
**Description of work**
Updated the eccentricity calculation so that it follows more standard equations. Specificaly it is now changed so that it follows the equations that are used in rdkit https://www.rdkit.org/docs/source/rdkit.Chem.Descriptors3D.html. 

Single rigid water trajectory
BEFORE
image
At some points, it's nearly zero therefore nearly perfectly spherical this looks wrong.

AFTER
image
Now the rigid water molecule has a near-constant value for its eccentricity.

Single flexible water trajectory
BEFORE
image
It seems to change far more than it should. Again at some points, it's nearly zero therefore nearly perfectly spherical which is wrong.

AFTER
image
A flexible water molecule eccentricity changes with time. What is nice is that we can see that water geometry changes in a nice oscillatory manner but mostly stays the same.

**Fixes**
(Fixes #552) The new equations that are used are do not depend on the orientation of the selected atoms.
(Fixes #378) Weight setting is removed as it is not meaningful for this calculation.
(Fixes #380) ROG calculation has been removed.

**To test**
Run the eccentricity calculation which specific atom selections and check that results are consistent with expectations e.g. spherical atom selections should have small values of the eccentricity.

### DIFF
--- a/Doc/pages/dynamics.rst
+++ b/Doc/pages/dynamics.rst
@@ -373,7 +373,7 @@ averaged and normalised so that for liquid or gaseous systems
 .. math::
     \lim_{r \rightarrow \infty } G_{\mathrm{d}}(r,\Delta t) = \lim_{\Delta t \rightarrow \infty } G_{\mathrm{d}}(r,\Delta t) = 1
 
-and the self part of the van Hove function in *MDANSE* is summed do
+and the self part of the van Hove function in *MDANSE* is summed so
 that it only depends on the distance :math:`r`
 
 .. math::
--- a/Doc/pages/structure.rst
+++ b/Doc/pages/structure.rst
@@ -137,8 +137,8 @@ The eccentricity of a selection of atom is calculated using the equation
 
 where :math:`\lambda_{1}` and :math:`\lambda_{3}` are its smallest and
 largest principal moments of inertia. A spherically symmetric
-selection of atoms will have an eccentricity of 0 and an aspherical
-selection of atoms will have a eccentricity of 1.
+selection of atoms will have an eccentricity approaching 0 and an
+aspherical selection of atoms will have a eccentricity approaching 1.
 
 .. _molecular-trace:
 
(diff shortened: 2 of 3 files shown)


MBartkowiakSTFC (approved):
The current version works correctly for all the cases I tried. Thanks for the changes!

PR #558, opened 2024-09-25
Remove or replace several Cython extensions

MBartkowiakSTFC:
**Description of work**
Two Cython extensions have been identified as unused and removed. A third one mt_fast_calc, has been replaced with a faster numpy equivalent.

**Fixes**
1. Deleted sd_fast_calc.pyx
2. Deleted change_coordinates.pyx
3. Deleted mt_fast_calc.pyx and replaced the corresponding function call with equivalent Python code using numpy.

**To test**
Please run MolecularTrace analysis on a trajectory using the protos branch. Switch to this branch and run the same MolecularTrace again. The results should be the same.


MDANSE/Src/MDANSE/Framework/Jobs/MolecularTrace.py
```diff
@@ -154,13 +153,13 @@ def run_step(self, index):
...
-            self.configuration["spatial_resolution"]["value"],
-            self.min,
-        )
+        resolution = self.configuration["spatial_resolution"]["value"]
+
+        indices = np.floor(
+            (conf["coordinates"][self._indexes, :] - self.min.reshape((1, 3)))
+            / resolution
+        ).astype(int)
+        grid[tuple(indices.T)] += 1
```
ChiCheng45: I think there is a slight difference to the cython code when there are a duplicated set of indices  e.g.

```
a = np.zeros((2, 2))
a
Out[21]: 
array([[0., 0.],
       [0., 0.]])
a[(np.array([0, 1, 1]), np.array([0, 1, 1]))] += 1
a
Out[23]: 
array([[1., 0.],
       [0., 1.]])
```

I ran with a DFTB water trajectory and there are some small differences in the output.


OLD
image

NEW
image

ChiCheng45 (changes requested):
See above

MBartkowiakSTFC:
Thanks for pointing this out. I implemented a correction, and now I get this result for the DFTB water:
corrected_molecular_trace

ChiCheng45 (approved):
Looks good.

PR #561, opened 2024-09-27
Replace atomic trajectory cython version with numpy implementation

raunakkbanerjee:
**Description of work**
Deleted atomic_trajectory.c and atomic_trajectory.pyx, added atomic_trajectory.py with numpy implementation of the cython function
**Fixes**
Readability and portability issues related to using cython.

**To test**
Tested on the DFTB H2O trajectory


MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculeWidget.py
```diff
@@ -13,7 +13,7 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
-from qtpy.QtWidgets import QComboBox
+from qtpy.QtWidgets import QComboBox, QPushButton
```
ChiCheng45: It doesn't look like this is being used. I think you can undo the change in this file.

MDANSE/Src/MDANSE/Trajectory/AtomicTrajectory.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
```
ChiCheng45: I think the file doesn't belong in this directory. Maybe we could put this function somewhere else.

MDANSE/Src/MDANSE/Trajectory/AtomicTrajectory.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def atomic_trajectory(config, cell, rcell, box_coordinates=False):
+    trajectory = np.empty((config.shape[0], 3), dtype=np.float64)
```
ChiCheng45: I think this is slightly different to the cython code, also I think the function can be cleaned up slightly aswell. Also could you add a numpy docstring and type hints to this?

ChiCheng45:
I just tested with DL_POLY water trajectory with a temperature calculation. I see some differences with your changes.

Before
image

After
image

ChiCheng45:
Looks like its fixed now.

image

PR #562, opened 2024-09-27
Simplify the weight widget

MBartkowiakSTFC:
**Description of work**
Weight widget has too many options, most of them unnecessary. It is also not clear what the default option was for each analysis type after the value in the widget has been (accidentally?) changed by the user.

**Fixes**
1. Weights widget shows less entries than before.
2. The possible weights are sorted alphabetically,
3. There is a 'mass' alias for 'atomic_weight',
4. The default value for each analysis is shown in bold font.

**To test**
Load a trajectory and pick a scattering analysis. Change the value of weight, and see if the original value appears in **bold**. Choose mass and check if it does not trigger an error.


ChiCheng45 (approved):
Looks good.

PR #563, opened 2024-09-30
Add missing function name to ResolutionCalculator output

MBartkowiakSTFC:
**Description of work**
PR https://github.com/ISISNeutronMuon/MDANSE/pull/529 restructured the resolution calculation. This interfered with the resolution helper dialog, which requires one additional item in the parameter dictionary.

**Fixes**
1. The missing function name has been added to the output of ResolutionCalculator.
2. InstrumentInfo removes the function name from the dictionary, so it does not appear twice.

**To test**
Try setting the resolution in a scattering job when no instrument is selected.
Try changing the resolution type in an instrument profile in the instrument tab.
Both should work and create valid resolution parameters.


ChiCheng45 (approved):
Looks good. I've tested and its working again.

PR #566, opened 2024-10-04
Fixed the current correlation calculation

ChiCheng45:
**Description of work**
Fixed and refactored the current correlation function job. Removed the interpolation_mode and preloaded frames settings since this greatly complicated the code.

Here is the longitudinal current correlation function of the argon trajectory.

image

Compare against this figure from molecular hydrodynamics, see page 6 in https://www.google.co.uk/books/edition/Molecular_Hydrodynamics/GcTR1xSl0ZkC?hl=en&gbpv=0.
the agreement looks good.

The current correlation function peaks are not particularly sharp and the peak's heights are different for different values of |Q| in the plot so you get the smeared curve above. 

Note: I found it quite difficult to get clean results, the reason why this job is noisy is probably the same reason why the DCSF is typically noisier than the DISF job. It probably has something to do with the increased number of configurations that the time correlation function can occupy. I recommend increasing the number of q vectors and reducing the number of correlation frames so that you average over a larger number of configurations. Additionally, I think that for larger time differences you will need more configurations to get cleaner results. For the above results, I used 800 q vectors and a frame setting of [0, 5000, 1, 200] so it averages over 4800 time steps.

**Fixes**
Fixed the current correlation function so that it can now be run. 
Moved the current correlation function job to dynamics.

**To test**
Run the current correlation function and check that it runs without issues. Unit tests should pass.


MBartkowiakSTFC (approved):
Looks good! I have not found any problems.

PR #568, opened 2024-10-07
Molecular Trace fix and plotting improvements

### LINKED ISSUE #565: [BUG] Molecular Trace output has no axes
**Description of the error**
At the moment the MolecularTrace analysis outputs a 3D array, but the 3 axes associated with the output data are not written to the file.

We should have 3 arrays of position along x, y and z written into the output file, specifying the coordinates of each bin of the grid.

The only entry in the file specifies a single 'index' axis, which means that each point has its own number.

**Suggested fix**
MolecularTrace code has to be corrected to include the missing datasets.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Following a user request, this PR makes it possible to plot Molecular Trace results.

closes #565 

**Fixes**
1. MolecularTrace output has been corrected and now includes plotting axes,
2. Plotting interface now has principal axis selection option,
3. Plotting interface "use it" column now handles text input for selecting a subset of data.
4. Text output of N-dimensional arrays is now possible in the plotter's text view.

**To test**
Run a MolecularTrace analysis on a trajectory.
Load the results into the plotter. Try changing the plotting axis in the Single mode. Choose a specific set of curves by putting a range description into the "Use it" column. (Accepted are single numbers, ranges 5-7, Python slices 2:8:3, and combinations of these. E.g. 5,7,15;18-20 will plot curves 5,7,15,18,19,20.)
Switch to Heatmap mode. Input a single number into the "Use it" column to plot a different plane from the 3D set.
Create a Text widget in the plotter and use it to save molecular trace 3D array to a CSV file.


ChiCheng45 (commented):
Looks good. My tests work fine. 

Can we update the heatmap view so that it rotates the plot when the main axis is changed?

image

MBartkowiakSTFC:
Now it should be possible to swap axes in a heatmap plot by changing the principal axis.

In the course of the changes here I removed the option of plotting multiple heatmaps next to each other. The handling of multiple datasets has become complicated and I think that the plotting code needs to be refactored after November. For now if a user wants to compare two 2D maps, they would have to create them separately in two plotting tabs.

ChiCheng45 (commented):
Looks great just one last thing. Can we update the plotter so that the axis limits change to the new defaults when the main axis is changed? This will save the user from having to click the home button or zoom in after the main axis has been changed.

Animation49

MBartkowiakSTFC:
I think that we will leave it the way it is for now. Let's add it to the list of things that will be fixed in the plotter refactoring.

ChiCheng45 (approved):

PR #571, opened 2024-10-09
Update Gromacs converter

MBartkowiakSTFC:
**Description of work**
Recent work has shown that the current Gromacs converter in MDANSE could not read complicated trajectories correctly. Both the PDB reader and the XTC/TRR Cython extensions were triggering errors when converting a trajectory.

NOTE: `gmx editconf` has been shown to produce _wrong formatting_ in PDB files. When such a file is used, the chemical elements will not be recognised correctly. The most reliable tool for creating the input PDB file so far has been `gmx trjconv`.

**Fixes**
1. Cython extension code (originally copied from mdtraj) has been replaced with a newer version, also from mdtraj.
2. MDANSE Gromacs converter has been adapted to deal with the changed API of the Cython extensions.
3. PDBReader has been replaced with a new MinimalPDBReader.

**To test**
All tests must pass.
A Gromacs trajectory should be converted to the MDANSE format from both .xtc and .trr files. The resulting trajectory should be the same in both cases, except that the .trr file may also contain velocities.

### DIFF
--- a/MDANSE/Extensions/xtc/include/xdr_seek.h
+++ b/MDANSE/Extensions/xtc/include/xdr_seek.h
@@ -12,5 +12,6 @@
 
 int64_t xdr_tell(XDRFILE *xd);
 int xdr_seek(XDRFILE *xd, int64_t pos, int whence);
+int xdr_flush(XDRFILE* xd);
 
 #endif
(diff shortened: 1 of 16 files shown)


ChiCheng45 (approved):
Looks good.

PR #572, opened 2024-10-10
Update the normalization of the van hove functions

ChiCheng45:
**Description of work**
Updated the self-part of the van Hove function so that both $4 \pi r^2 G_s(r,t)$ and $G_s(r,t)$ are saved. Changed the normalization so that it is now consistent with the distinct part. The normalization of $4 \pi r^2 G_s(r,t)$ has been corrected.

Here I have plotted both the self and distinct part of the van Hove function on the same plot for the 0th and 20th time steps.
image
image

**To test**
Run the self-part of van Hove function and check the results, rebuild MDANSE as some cython was changed. Check the self-part van Hove function, there should now be two outputs with different normalizations.  The $4 \pi r^2 G_s(r,t)$ output should be similar to the output from before. A new $G_s(r,t)$ should be available, normalized in the same way as the distinct part.

### DIFF
--- a/MDANSE/Extensions/van_hove.pyx
+++ b/MDANSE/Extensions/van_hove.pyx
@@ -116,6 +116,7 @@ def van_hove_distinct(
 def van_hove_self(
     double[:,:] xyz,
     double[:,:] histograms,
+    double[:] cell_vols,
     double rmin,
     double dr,
     int n_configs,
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC (approved):
I tried it on a few trajectories, and the results looked good.

PR #573, opened 2024-10-10
Beta release 2024

### LINKED ISSUE #492: [BUG] Wrong limits in Density Profile
**Description of the error**

A Density Profile analysis sets the length of the output x axis correctly, but does not detect that the atom positions are offset in respect to the limits.

Screenshot 2024-07-10 at 14 43 02

In the plot above, there should have been peaks within the entire plotting range

Half of the range is filled with zeros. This means that the analysis is trying to find atoms in the wrong range of coordinates. The atoms in this simulation were distributed across the entire simulation box, which was over 22 Angstroms wide.

**Suggested fix**
It may be the easiest just to translate the atoms by the unit cell vector to make them appear within the range sampled by the analysis.

**Additional details**
Additional checks need to be performed for non-periodic simulation boxes.

### LINKED ISSUE #576: Disable some or all of the QVTKRenderWindowInteractor hotkeys
**Description of the error**
I think we might want to try to disable some or all of the QVTKRenderWindowInteractor hotkeys. See https://docs.vtk.org/en/latest/api/python/vtkmodules/vtkmodules.qt.QVTKRenderWindowInteractor.html#vtkmodules.qt.QVTKRenderWindowInteractor.QVTKRenderWindowInteractor. 

The problem is that a user may accidentally hit one of the hotkeys and change the behavior of the 3D view. They might think there is some bug. For example, hitting 'j' key changes it to joystick mode so that holding the mouse button rotates the 3D view. Only by hitting the 't' key does it change back to the default mode where you hold the mouse button and drag the mouse to rotate the 3D view.

### LINKED ISSUE #577: [BUG] Invalid q_vector settings for range and vector setting but GUI still allows the job to be run
**Description of the error**
If you enter an invalid setting the q vector generators for the RangeConfigurator or VectorConfigurator settings the RUN! button is not greyed out so that the job can still be run. MDANSE uses the default values instead.

**Suggested fix**
When the q vector configurator settings are not valid, we should block the job from being able to be run like the other configurators.

**Additional details**
image

### LINKED ISSUE #579: [BUG] Grouping level with RMSF job results plot
**Description of the error**
Running the RMSF job with the molecule grouping level with the DFTB trajectory (which has H2O molecules) and plotting the results creates an empty plot.

It looks like the issue is occurring because MDA stores an indexes array which is the indexes of the atoms instead of the indexes of the molecules.

image

image

### LINKED ISSUE #580: [ENHANCEMENT] Remove the cross from the preview tab in the plot holder
Let's remove the cross from the preview tab in the plot holder as it doesn't do anything.

image

### LINKED ISSUE #582: [ENHANCEMENT] Update default value behaviour of the instruments settings
Some of the instrument setting options appear to have default values. Here I set the resolution_fwhm, q_min, and q_max to an empty cell
image
this produces an instrument that uses some default values for resolution_fwhm, q_min, and q_max. Using this instrument for a job works fine.

When I set an empty cell for q_step, q_width, or vectors_per_shell the instrument setting fails to be created. You can still select this instrument but nothing changes in the q_vectors setting.
image

When the cell is empty then it should use some default values and show what they are similar to other input cells e.g. last frame in the frames configurator.
image

MBartkowiakSTFC:
**Description of work**
Final changes before the beta release of MDANSE 2.

closes #580
closes #579
closes #577
closes #576 
closes #582
closes #492 

**Fixes**
1. Updated version numbers of MDANSE and MDANSE_GUI,
2. Splash screen text changed to 'beta release',
3. Correction to the job progress bar, to ensure that a finished job is 100% complete.
4. Initialised the primary axis field in the plotting GUI with the correct axis name.
5. Correction to the Grid.py file, allowing to plot curves with different primary axes.

**To test**
Run a DOS job on a trajectory. Check that it reached 100% when finished.
Plot the results of DOS and VACF at the same timer. Switch the plot to Grid. All selected curves should be visible.


ChiCheng45:
So far look's good, still doing some testing/checking.

Need to update the version in the help box.

image

ChiCheng45:
Looks, good so far.

I've found one issue with the project coordinates configurator. For the axial and planar projection, you can input the vector of zeros, this causes the jobs to fail. Also, non-numerical numbers can be entered into the boxes, hitting run doesn't do anything but we should stop the user from entering these like the other configurators.

image

MBartkowiakSTFC:
The projection widget should work correctly now.

ChiCheng45:
So far I have found some minor issues (#577, #576, #575 and #574) which I think we need not fix for this release.

There is a lammps trajectory conversion failure with the p_lammps_10000 test file from one drive.  This does not work on protos either and the issue appears to have been introduced in #546. I tested MDANSE at a98af806e87e9c9ba7bdbcf208e4142a8d38af8f and the conversion works fine at this point. I think we should try to fix this before the release.

MBartkowiakSTFC:
The problem is that LAMMPS has multiple possible formats of the configuration file, and specifically this section:
https://docs.lammps.org/atom_style.html
is what creates the problem. Depending on which one is selected in the LAMMPS script, the number of columns in the Atoms section and their meaning will be different.

Some of the files conveniently contain a comment after the word 'Atoms' which specified which atom_style is being used. In the long run we should probably aim to support all of them, but there are quite a few possibilities. We can't really rely on the comment being there, and being correct, so some other mechanism of recognising the atom_style will have to be implemented.

Having said that, I think I fixed the problem with our example trajectory. It should work again.

ChiCheng45:
It looks good all my lammps test trajectories work fine now. 

Can we update the self-part of the van Hove function so that it gives the same error as the distinct part when a simulation without a unit cell is used?

This is error in the self-part

```
179, in run_step
.unit_cell.volume
^^^^^^^^^
AttributeError: 'RealConfiguration' object has no attribute 'unit_cell'
```

This is the error with the distinct-part

```
line 81, in detailed_unit_cell_error
raise ValueError(
ValueError: This analysis job requires a unit cell (simulation box) to be defined. The box will be used for calculating density in the analysis. You can add a simulation box to the trajectory using the TrajectoryEditor job. Be careful adding the simulation box, as the wrong dimensions can render the results meaningless.
```

ChiCheng45:
I have identified two new issues we may want to try to fix #579 and maybe #581. For #581 It might be better to deactivate the unfolded trajectory job for this release.

MBartkowiakSTFC:
Thanks for finding these problems! In the meantime I added the error message to the van Hove self function.

Also, I think that the commits I added today should close #580 #576 #577 and make the AverageStructure analysis fail sooner, which does not solve the problem, but is an improvement.

ChiCheng45:
Looks like #580, #576 and #577 are fixed now. 

Still one minor visual issue with the q_vector generator is that when you use a vector or range of 4 or more elements the cell in the table is not highlighted. The run button is blocked so it is not much of an issue, so I think we can leave this as it is.

image

MBartkowiakSTFC:
To address #582 I added placeholder text to the instrument input fields. The values are not very useful (they are all 1) but at least this produces a valid configuration and it is clear what values were used.

MBartkowiakSTFC:
Also, #579 should be fixed now.
#581 has been addressed by disabling UnfoldedTrajectory for now.

ChiCheng45 (approved):
Looks great, I think this is ready to be released.

PR #588, opened 2024-10-19
Make iterable unpacking backwards compatible

MBartkowiakSTFC:
**Description of work**
The syntax of PlottingContext for n-dimensional arrays was not handled correctly by Python versions before 3.11.

**Fixes**
- replaced the `*` unpacking operator with a conversion to tuple.

**To test**
Create and load a 3D data array (Molecular Trace will do it). Using Python 3.9, make sure that you can start the GUI, load the data and plot a 2D slice using the Heatmap plotter.


ChiCheng45 (approved):
Looks good. I've tested on 3.9 and it works fine. The molecular trace plotting is the same on protos and this branch.

PR #590, opened 2024-10-23
Correct dlpoly and instruments

MBartkowiakSTFC:
**Description of work**
This PR fixes some problems which became apparent when testing on the visa platform.

**Fixes**
- DL_POLY converter does not ignore velocities and forces if they are present in HISTORY,
- MANIFEST.in has been updated to include InstrumentDefinitions.toml
- version numbers have been incremented, so this build can be used on PyPI
- disabled the 'Neutron Dynamic Total Structure Factor' since it does not check if weights had already been applied,
- changed the normalisation factor in Dynamic Incoherent Structure Factor.

**To test**
Install MDANSE on a fresh machine. Check that initial list of instruments is present.
Convert a DL_POLY trajectory with velocities, see that they get written to the output file.
Try to reproduce the argon results from CCPBioSim tutorial 3, using atoms with effective scattering lengths of b_coh=0.0025 and b_inc=1e-10, based on the sigma values reported in the Skold paper. The ratio of DISF to DCSF should be close to the one presented in the paper.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
@@ -230,7 +230,7 @@ class DynamicIncoherentStructureFactor(IJob):
 
             rho = np.exp(1j * np.dot(series, qVectors))
             res = correlate(rho, rho[:n_configs], mode="valid").T[0] / (
-                n_configs * qVectors.shape[1]
+                n_configs * rho.shape[1]
             )
 
             disf_per_q_shell[q] += res.real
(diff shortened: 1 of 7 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -230,7 +230,7 @@ def run_step(self, index):
 
             rho = np.exp(1j * np.dot(series, qVectors))
             res = correlate(rho, rho[:n_configs], mode="valid").T[0] / (
-                n_configs * qVectors.shape[1]
+                n_configs * rho.shape[1]
```
ChiCheng45, line 233: `qVectors.shape[1]` and `rho.shape[1]` should be the same

MBartkowiakSTFC: I have checked it again. You are right, this change makes no difference.

ChiCheng45 (approved):
Looks good.

PR #601, opened 2024-11-11
3d viewer fix when switching trajectories

### LINKED ISSUE #575: [BUG] 3D view trajectory issues when switching to another trajectory
**Description of the error**
The trajectory can disappear from the 3D view. This occurs when the max frame is set in one trajectory (e.g. 25) and then I switch to another trajectory whose max frame is smaller (e.g. 3). This fixes itself when I switch to another trajectory and then back again.

Animation50

A similar issue occurs when switching from one trajectory to another when the timestep is set to something. You can see the trajectory appear to jump. It looks like when you switch trajectories it resets the timestep to 0. This causes the jump when you press the play button when the GUI has the timestep set to something other than 0.

Animation51

ChiCheng45:
**Description of work**
PR to fix issues with the 3D viewer when switching between trajectories.

**Fixes**
Fixes #575.

**To test**
Load up two trajectories with a different number of frames. Switch to the trajectory with the largest number of frames and set the 3d view to its last frame then switch to the other trajectory. The trajectory of the system you switched to should be visible.

Load up two trajectories and set the frame to something other than zero. Switch between the trajectories. The frame in the 3D viewer should not be in the zeroth frame; see #575.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -601,12 +601,11 @@ class MolecularViewer(QtWidgets.QWidget):
         # Update the view.
         self.update_renderer()
 
-    def set_reader(self, reader, frame=0):
+    def set_reader(self, reader):
         """Set the trajectory at a given frame
 
         Args:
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
Looks good, no problems.

PR #602, opened 2024-11-11
Split trajectory frames into chunks

### LINKED ISSUE #596: [ENHANCEMENT] Improve trajectory chunking
At the moment, the HDF5 trajectory file stores the positions of all atoms at a given time step as a single chunk.

For trajectories with larger number of atoms, each frame should be split up further into fixed-size chunks. This will speed up the analysis whenever the calculation is performed per atom and not per frame.

Any attempt at increasing the size of a chunk along the time axis to include multiple frames ended up slowing down the converters.

**Additional context**
N/A

MBartkowiakSTFC:
**Description of work**
Trajectory arrays such as positions, velocities and charges are now split into fixed-size chunks within a single time step.

closes #596

Note: the changes in this PR will **break** the scripts in MDANSE-Examples. The scripts should be updated around the time of the next release.

**Fixes**
1. Changed chunking in TrajectoryWriter. The arrays are padded to ensure that their size is a multiple of the chunk size.
2. Added chunk size as a parameter to OutputTrajectoryConfigurator.
3. Added chunk size to OutputTrajectoryWidget.
4. Replaced test trajectories in unit tests with chunked ones.
5. Updated test inputs to inlcude the new parameter.

**To test**
Convert a trajectory using the protos branch. Run Dynamic Coherent/Incoherent Structure Factor analysis on this trajectory and time both runs.
Repeat the same steps using this branch. Compare the results and the timing.
Expected results: DCSF will slow down by ca. 5%, DISF will speed up by ca. 25%, the results should not change.


ChiCheng45 (approved):
Looks good. Calculations were run on a single core. 

DISF time reduced from 4.5 mins to 1 min. (500 frames, 2000 Atoms and 128 chunk)
DCSF times were similar 25s vs 20s. (2000 frames, 2000 Atoms and 128 chunk)

This is definitely an improvement for trajectories with a larger number of atoms and I find the same results with the new and old chunking schemes. I think there is probably more we could do to improve performances, we should look at this again in the future.

PR #603, opened 2024-11-11
Explicitly set the QLocale in the GUI

### LINKED ISSUE #595: [BUG] German Keyboard GUI Problem
When putting the value 0.2 as a timestep in the trajectory converter it is tranformed to 2E+00, which is wrong and not supported by the software. 

Possible reason for this is using , as a decimal place, like in a german keyboard.

MBartkowiakSTFC:
**Description of work**
The number format will always follow the British convention, where "1 point 0" is 1.0 and not 1,0 or anything else.

closes #595 

**Fixes**
1. Created a QLocale instance set to English, UK, before starting the GUI.

**To test**
On a platform already set to English, there should be no difference.
On platforms using different symbols for the decimal point, the MDANSE GUI should now accept the '.' character instead.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/main.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/main.py
@@ -53,6 +53,10 @@ def startGUI(some_args):
     path = os.path.dirname(os.path.abspath(__file__))
     app.setWindowIcon(QIcon(os.path.join(path, "Icons/MDANSE.ico")))
     fixed_locale = QLocale(QLocale.Language.English, QLocale.Country.UnitedKingdom)
+    fixed_locale.setNumberOptions(
+        QLocale.NumberOption.RejectGroupSeparator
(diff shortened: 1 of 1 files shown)


ChiCheng45:
I've tested this on a German Windows 11 VM and this change allows me to use '.' characters as the decimal point.

One problem I can see is if they use ',' as the decimal point then QT converts it to a different number e.g.

1,3 -> 1.3E+01

This might be confusing to new users. What do you think about disabling the use of commas altogether?

MBartkowiakSTFC:
Unfortunately, commas are allowed in English number formatting, as in "1,000,000.05" so a normal QDoubleValidator will not remove them. But you are right, they should be disabled.

MBartkowiakSTFC:
As far as I can tell, this should work now.

ChiCheng45 (approved):
Looks good.

PR #604, opened 2024-11-12
Indicate job start and completion

### LINKED ISSUE #593: [ENHANCEMENT] Visual feedback while running the file converter
When converting a trajectory to .MDT, there is no visual feedback to confirm that the process is running. This becomes especially relevant for big trajectories, which may take a while to convert. For this reason, a visual confirmation or a system message to confirm that the converter is still running without crashing is quite useful.

A visual confirmation indicating that the converting process is running properly.

**Additional context**
Tested to convert CP2K outputs.

MBartkowiakSTFC:
**Description of work**
Basic mechanism has been added to inform users about job start and objects appearing in other tabs.

closes #593

**Fixes**
1. "RUN!" button becomes inactive after job start, and the text changes to "starting" for 3 seconds.
2. Tabs which received new input now change the text colour and show an info icon.

**To test**
Run a converter. It should highlight the running job tab on start, and trajectory tab on job completion.
Run an analysis. It should highlight the running job tab on start, and plot creator tab on job completion.
Plot the results. Both creating a new plot and sending data to the plot should highlight the plot holder tab.
Check that the tab text and icon get reset correctly when clicked on.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
@@ -269,6 +269,7 @@ class TabbedWindow(QMainWindow):
         self.tabs.addTab(trajectory_tab._core, name)
         self._tabs[name] = trajectory_tab
         self._job_holder.trajectory_for_loading.connect(trajectory_tab.load_trajectory)
+        self._job_holder.trajectory_for_loading.connect(trajectory_tab.tab_notification)
 
     def createInstrumentSelector(self):
         name = "Instruments"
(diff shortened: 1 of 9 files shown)


ChiCheng45 (commented):
Looks good. 

Just one thing, I noticed that if you hit "run!" the button goes to the "starting..." state and stops the user from pressing the button. If you quickly change the settings the "starting..." text goes black and you can click it to run another job. Can you update it so that if the user does change the settings then the text changes to "run!" immediately?

MBartkowiakSTFC:
I changed the button behaviour. It should not allow the reactivation now, as long as the button still says "Starting". Also, if you change parameters to invalid ones, it will actually check if execution should be allowed after the 3 seconds wait.

ChiCheng45 (approved):
Looks good.

PR #607, opened 2024-11-15
Modify the VASP converter to work with several XDATCAR formats

### LINKED ISSUE #592: [BUG] Segmentation fault when trying to convert VASP MD trajectory using ASE.
I get a segmentation fault when trying to convert a VASP trajectory, using the ASE convertoer. 

I can share the relevant file if necessary by email; could likely be a user error.

MBartkowiakSTFC:
**Description of work**
Recent problems with the ASE converter have increased the importance of having a working VASP converter. This PR updates the VASP converter to make if work for several possible formats of XDATCAR, including those with variable unit cell size.

It is still impossible to work with an XDATCAR which does not define the atom types, but a comment has been added to the docstring to explain this to the user and suggest a solution.

closes #592

**Fixes**
- VASP.py has been modified.
- XDATCARFileConfigurator.py has been largely re-written.

**To test**
All tests must pass.
Please try out the VASP converter on a few XDATCAR files, if possible.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/XDATCARFileConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/XDATCARFileConfigurator.py
@@ -25,134 +25,110 @@ class XDATCARFileError(Error):
     pass
 
 
+def read_modern_header(source):
+    cell = []
+    system_name = source.readline().strip()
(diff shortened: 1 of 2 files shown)


ChiCheng45 (approved):
I've tested this on a few XDATCAR files and it works as expected.

PR #610, opened 2024-11-18
Improved output file names

### LINKED ISSUE #599: [ENHANCEMENT] Add checks to output file widgets
The output file widgets at the moment do not check if the output file is going to overwrite an existing file.

The following improvements could be combined:
1. The output widget should suggest an output filename.
2. The path should be the same as the trajectory path, or the path used most recently by the user.
3. The widget should not allow a to overwrite a file.
4. The widget should not allow to write to the same file from more than one job (i.e. it should check if the running jobs are going to use the same file name).

N/A

**Additional context**
N/A

### LINKED ISSUE #608: [BUG] Jobs writing to opened files.
**Description of the error**
When you run a job that save to opened files, it can cause them to end up being in a bad state. For example on windows if I run the trajectory converter with the output file set to a file already loaded then MDANSE will attempt to run the job. The job fails but it seems to clear the file in the process. This means this cleared file will still be loaded up in the MDANSE GUI.

**Suggested fix**
If the MDT or MDA file is already open in MDANSE then the job should be blocked from running. Alternatively the user should be warned that the loaded file will be overwritten, MDANSE should then unload those files, overwrite and reload them.

MBartkowiakSTFC:
**Description of work**
Introduces sensible suggestions of output file names. Prevents from writing into a file that it already open (which would fail anyway).

This PR does **NOT** prevent overwriting existing files, since users could have valid reasons to do so. The initial filename suggested by the GUI will prevent overwriting, and the file dialog will issue a warning when overwriting. Once a job has started, the output filename becomes protected, preventing from running another job with the same output filename. However, after the job has ended, it will be possible to overwrite the file ONLY after removing the output file from the GUI (e.g. from the plot creator of from the trajectory viewer).

closes #599 
closes #608 

**Fixes**
- improved communication between widgets and session settings,
- created a list of protected filenames in the session object,
- added filename suggestions to the output file widgets.

**To test**
Please run at least two trajectory converters. See if the output trajectory widget suggests a valid path and filename. It should be impossible to run the same analysis twice without changing the output filename.
Try running an analysis. Check if the suggested output filename is valid.
Load an analysis result into the plotter. Check if you can run an analysis that overwrites the file. It should be impossible to overwrite as long as the file is on the list of open files in the plot creator.


ChiCheng45 (commented):
Looks good just a few suggestions.

Can we update the trajectory converter widget so that the full path of the suggested trajectory is written out? I think it might not be obvious to a new user that ".\DL-POLY_trajectory1" will be the place where they ran mdanse_gui.

In the analysis job output file suggestion it uses a mix forward and back slash on windows e.g. E:/MDANSE-Examples-main/Data/Trajectories/DFTB\DensityOfStates_result2. Can you update it so on windows it only uses the forward or back slash?

ChiCheng45 (commented):
I still find #608 occurring for a few cases. 

This can occur when the path of the trajectory loaded is the absolute path and the input is a relative path. For example, I can load a trajectory which has the path "D:/Users/chi-y/DFTB_trajectory1.mdt" and I used a relative path in the converter tab ".\DFTB_trajectory1". Running the job causes the conversion to fail and the file to be cleared.

I'm also still able overwrite MDA files that are loaded in the plotter. Could this be some issue with window paths?

MBartkowiakSTFC:
> Looks good just a few suggestions.
> 
> Can we update the trajectory converter widget so that the full path of the suggested trajectory is written out? I think it might not be obvious to a new user that ".\DL-POLY_trajectory1" will be the place where they ran mdanse_gui.
> 
> In the analysis job output file suggestion it uses a mix forward and back slash on windows e.g. E:/MDANSE-Examples-main/Data/Trajectories/DFTB\DensityOfStates_result2. Can you update it so on windows it only uses the forward or back slash?

Normally, the full path should appear once you have selected one of the input files. Has this happened at all?

MBartkowiakSTFC:
> I still find #608 occurring for a few cases.
> 
> This can occur when the path of the trajectory loaded is the absolute path and the input is a relative path. For example, I can load a trajectory which has the path "D:/Users/chi-y/DFTB_trajectory1.mdt" and I used a relative path in the converter tab ".\DFTB_trajectory1". Running the job causes the conversion to fail and the file to be cleared.
> 
> I'm also still able overwrite MDA files that are loaded in the plotter. Could this be some issue with window paths?

Thanks for pointing this out. I will change the way paths are handled.

ChiCheng45:
> > Looks good just a few suggestions.
> > Can we update the trajectory converter widget so that the full path of the suggested trajectory is written out? I think it might not be obvious to a new user that ".\DL-POLY_trajectory1" will be the place where they ran mdanse_gui.
> > In the analysis job output file suggestion it uses a mix forward and back slash on windows e.g. E:/MDANSE-Examples-main/Data/Trajectories/DFTB\DensityOfStates_result2. Can you update it so on windows it only uses the forward or back slash?
> 
> Normally, the full path should appear once you have selected one of the input files. Has this happened at all?

Yes, this seems to occur now.

ChiCheng45 (commented):
Looks good. Just a few more issues I've found.

Windows file systems are not case-sensitive so C:\Users\xcb63893\Documents\dftb and C:\Users\xcb63893\Documents\DFTB are the same file. MDANSE sees these as different files so I can try to run another job to write to the same file and cause the job to fail.

On windows, when I use the file browse dialog the file path switches from backslashes to forward slahes. Can we change it so that it is always backslahes?

Animation52

ChiCheng45:
The issues above are fixed now. The only thing now is that I'm unable to overwrite files if I've deleted them from the trajectories tab or plot holder.

Animation53

MBartkowiakSTFC:
The latest changes fixed the following problems:
1. The JobEntry would execute on_finish method twice, and so it would remove the protected filename from the list twice.
2. AverageStructure jobs would not check if the output file was protected.
The only problem that remains is that, if the user unchecks auto-loading of the results, the output file widget will not unlock automatically when the running job finishes. It will unlock once the user has changed to another tab instead. This should not be a serious problem, since there are not many cases where you would want to run a job, wait for it to finish and then overwrite the results immediately without checking the original results.

ChiCheng45 (approved):
Looks good.

PR #611, opened 2024-11-18
Add MDAnalysis converter

### LINKED ISSUE #585: [ENHANCEMENT] Add MDAnalysis converter.
Similarly to the ASE converter, we should have an MDAnalysis converter. Basically, we should have some sort of converter that converts the trajectory using MDAnalysis and then using their trajectory object we create our MDT trajectory file.

Also, this will allow us to investigate whether parts or all of the MDANSE trajectory objects could be replaced with the MDAnalysis trajectory objects.

ChiCheng45:
**Description of work**
Added MDAnalysis converter so that a larger range of files can now be converted to the MDANSE format. Trajectories are now loaded up with a topology file and optionally a multiple number of coordinate files, see https://userguide.mdanalysis.org/stable/formats/index.html#formats for details. In MDAnalysis atom can be group following this scheme

image

I have not mapped the MDAnalysis atom grouping into the MDANSE atom grouping system as it is not clear what exactly the mapping should be and whether we are keeping the chemical entity objects in the short/medium term. 

Features and changes
- The input file format for the topology and coordinate format can be specified or the AUTO setting can be used to let MDAnalysis determine the format.
- Multiple files can be loaded up into the coordinates files. MDAnalysis will stich together these trajectories. A continous frame stiching option has been added, see https://userguide.mdanalysis.org/stable/reading_and_writing.html.
- Atom mapping has be updated so that the selection boxes are in a scrollable area. This should be better for the larger system that can now be converted.

Closes #585 

**To test**
1) Convert some gromacs trajectories from the MDANSE-Examples repository. 
Enter PDB files into the topology setting and trr or xtc files should be entered into coordinate files setting. Once entered the atom mapping and time step should update automatically. Check that the files convert with any issues
2) Convert a DL POLY trajectory.
Convert the dlpoly trajectory using the history file from MDANSE-Examples. For this case we only need to enter a file in the topology file input. As the file extension is not added to these files you will need to change the combobox to HISTORY. Once converted check that the files convert with any issues.
3) Try converting a trajectory using multiple coordinate files.
Either generate multiple coordinate files or copy and rename them. Convert the trajectory, if you copy and renamed it then the MDANSE trajectory should be twice as long and repeat itself.
4) Try converting some trajectories from https://www.ccpbiosim.ac.uk/slim-md/slimmd-database, most of these should convert without any issues.


MBartkowiakSTFC:
I have one observation so far: now that it is possible to input multiple trajectory files, we may need to change the input widget and the button set. Maybe a list would work better? Something where we can see all the trajectories that have been selected so far, and easily add or remove one. The 'Browse' button overwrites the existing selection at the moment, so it is hard to combine trajectories that are located in different directories.

MBartkowiakSTFC:
When I convert trajectories using .pdb or .gro files as topology files, the atom labels (HGA1, OW, etc.) get discarded, and the only information left is the chemical element symbol. However, if a .psf file is used as the topology files, the labels from the topology files are preserved and can be used in atom selection (under "atom name"). The preferred situation would be to have the labels available for any input file in which they are present.

MBartkowiakSTFC (commented):
Overall, the converter works very well. Finally we have a way of converting GROMACS trajectories with the optional change to atom type mapping. Also, so far the chemical elements have been recognised correctly every time I tried it.

I suggest that we open a new issue after merging this PR for an improved input widget for multiple trajectories. For now, please check if uniform behaviour can be achieved for atom labels read from the topology files (see comment above). Once this is addressed, we should be ready to merge.

MBartkowiakSTFC:
One more thing is that unit tests now leave behind two files, .md.xtc_offsets.lock and .md.xtc_offsets.npz
We should probably either clean them up or at least add them to .gitignore.

ChiCheng45:
> I have one observation so far: now that it is possible to input multiple trajectory files, we may need to change the input widget and the button set. Maybe a list would work better? Something where we can see all the trajectories that have been selected so far, and easily add or remove one. The 'Browse' button overwrites the existing selection at the moment, so it is hard to combine trajectories that are located in different directories.

Yes that might be better. I guess if we go for this, we might as well add the optional to have different formats for each file. I've created #612 for this.

> When I convert trajectories using .pdb or .gro files as topology files, the atom labels (HGA1, OW, etc.) get discarded, and the only information left is the chemical element symbol. However, if a .psf file is used as the topology files, the labels from the topology files are preserved and can be used in atom selection (under "atom name"). The preferred situation would be to have the labels available for any input file in which they are present.

This has been updated, the atom names after conversion should be the same.

> One more thing is that unit tests now leave behind two files, .md.xtc_offsets.lock and .md.xtc_offsets.npz
We should probably either clean them up or at least add them to .gitignore.

I've updated the test so that these files get removed. It seems for performance reasons and for certain files MDAnalysis generates these temp files.

ChiCheng45:
Thanks for the suggestions. I've made the suggested changes and this should be ready for another review.

PR #613, opened 2024-11-20
Set charges using TrajectoryEditor

### LINKED ISSUE #331: [ENHANCEMENT] Trajectory editor
It should be possible to change at least the atom types in the chemical system definition. A window/dialog or a tab in the GUI would have to enable editing the properties and saving them to a file. This way it would be possible to correct a trajectory after a wrong conversion.

The changes must be written to the trajectory file. I think that the original definition of the system should be kept in the file for reference, and to allow the user to undo the changes.

We could output a new trajectory every time we edit, but this is not optimal for storage.

**Additional context**
Additional changes the trajectory editor should be able to make:
- assign charges
- find molecules

### LINKED ISSUE #525: [ENHANCEMENT] Save per-atom electric charge in the trajectory
At the moment we support two trajectory formats: H5MD and MDF. The former stores charge information, and the latter does not.

MDANSE should:
1. Use the charge values specified by the user, if needed and specified.
2. Use charges stored in the trajectory, if available and not specified by the user.
3. Add charge information to a trajectory using TrajectoryEditor.

N/A

**Additional context**
Now that the TrajectoryEditor is available, setting the charges becomes more feasible.

MBartkowiakSTFC:
**Description of work**
Adds the option of writing fixed (time independent) partial charges into a trajectory using TrajectoryEditor. Also, TrajectoryEditor will now replace MoleculeFinder.

closes #525 
closes #331 

**Fixes**
- Added PartialChargesConfigurator to TrajectoryEditor,
- Added charge writing to TrajectoryEditor.
- Added molecule search to TrajectoryEditor.

**To test**
Try setting charges in different trajectories. The output trajectory should preserve the original charges and overwrite those that were explicitly set in the inputs.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/OptionalFloatConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/OptionalFloatConfigurator.py
@@ -1,0 +1,135 @@
+#    This file is part of MDANSE.
+#
+#    MDANSE is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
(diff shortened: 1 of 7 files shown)


ChiCheng45 (commented):
Great look's like it does the job. Any reason why we are keeping the moleculefinder job?

MBartkowiakSTFC:
No reason, really. #616 removes the MoleculeFinder, but I guess it could have been done here.

ChiCheng45:
> No reason, really. #616 removes the MoleculeFinder, but I guess it could have been done here.

Great we can remove it in #616. I will approve this change.

ChiCheng45 (approved):

PR #616, opened 2024-11-29
Simplify Chemical System

### LINKED ISSUE #351: [ENHANCEMENT] Trajectory and chemical system refactor.
Both the trajectory and chemical system object codes are both quite convoluted and need either replacement or some refactoring. Here are some problem with the chemical entity and trajectory objects that I have thought of so far.

- Bonding is set in the atom objects using a list of atoms. This makes adding and then tracking bonding more difficult. This is unintuitive, the bonding should be stored in an object which contains the atoms e.g. the molecule object.
- The chemical entity is time-independent and never changes. There could be cases where the molecule dynamics simulation makes or breaks bonds. This would mean that the chemical entity objects would need to change over time.
- The trajectory object treats some data as time dependent or independent e.g. bonding is time-independent and the unit cell is time-dependent. However the trajectory object should be able to support either possibility. For example, we may be given an input trajectory with only a single unit cell so that we'd have a time-independent unit cell. 
- It will be useful for the trajectory or chemical entity objects to know whether an atom belongs to some bulk structure or molecule for atom selection or other purposes. 
- For large trajectories loading from the HDF file can be slow because of the literal_eval calls when building the atoms objects.
- The current trajectory object is unable to deal with trajectories from PIMD simulations.

Both objects are quiet central to the whole MDANSE code and would be difficult to replace easily but I think we should begin a discussion on its possible replacement.

### LINKED ISSUE #478: [BUG] Coordination Number default parameters
**Description of the error**
Coordination Number analysis is derived from DistanceHistogram, same as the Pair Distribution Function. For the PDF we have changed the default limits of r to go from 0 to the maximum distance allowed by the simulation box. We should make the same change for Coordination Number.

The analysis should be run using parameters which sample the distance range covered by the simulation.

Default inputs make the calculation run on distances from 1 to 10 nm in the steps of 1 nm, independent of the simulation box size. The resulting curve for small boxes is usually flat (i.e. constant as a function of r), which is not informative and not correct.

**Suggested fix**
The same mechanism of setting the initial values of the range of r that was used for Pair Distribution Function should be used here.

**Additional details**
N/A

### LINKED ISSUE #545: [BUG] CenterOfMassesTrajectory changes elements and density

### LINKED ISSUE #641: [BUG] XRayStaticStructureFactor failure
**Description of the error**
XRayStaticStructureFactor fails to complete. The following error is seen in the logs.

```
2025-01-24 10:08:15,414 - CRITICAL - process[32092] - IJob 419 - Job failed with traceback: Traceback (most recent call last):
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 413, in run
    self.finalize()
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\XRayStaticStructureFactor.py", line 199, in finalize
    xssfIntra = weight(
                ^^^^^^^
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Arithmetic.py", line 48, in weight
    weights = get_weights(props, contents, dim)[0]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Arithmetic.py", line 40, in get_weights
    if abs(normFactor) > 0.0:  # if normFactor is 0, all weights are 0 too.
       ^^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

Process Subprocess-2:
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 413, in run
    self.finalize()
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\XRayStaticStructureFactor.py", line 199, in finalize
    xssfIntra = weight(
                ^^^^^^^
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Arithmetic.py", line 48, in weight
    weights = get_weights(props, contents, dim)[0]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Arithmetic.py", line 40, in get_weights
    if abs(normFactor) > 0.0:  # if normFactor is 0, all weights are 0 too.
       ^^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

During handling of the above exception, another exception occurred:

  File "C:\Users\***\AppData\Local\anaconda3\envs\MDANSE\Lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE_GUI\Src\MDANSE_GUI\Subprocess\Subprocess.py", line 61, in run
    self._job_instance.run(self._job_parameters)
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 420, in run
    raise JobError(self, tb)
MDANSE.Framework.Jobs.IJob.JobError: Traceback (most recent call last):
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 413, in run
    self.finalize()
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\XRayStaticStructureFactor.py", line 199, in finalize
    xssfIntra = weight(
                ^^^^^^^
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Arithmetic.py", line 48, in weight
    weights = get_weights(props, contents, dim)[0]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\***\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Arithmetic.py", line 40, in get_weights
    if abs(normFactor) > 0.0:  # if normFactor is 0, all weights are 0 too.
       ^^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```

MBartkowiakSTFC:
**Description of work**
Most of the classes derived from ChemicalEntity were only used for a very specific case of a trajectory created by GROMACS with a PDB file describing the topology. Since MDANSE is moving to MDAnalysis for loading GROMACS trajectories, the selection mechanism based on fixed databases needs to be replaced with a more general solution. This is provided by rdkit and substructure matching. Since rdkit introduces its own indexing of atoms, not guaranteed to be consistent with the internal indexing of the ChemicalSystem class, changes are necessary to improve the consistency of the atom indexing.

closes #545 
closes #351
closes #641 
closes #478

**Fixes**
1. ChemicalEntity and all derived classes have been removed from the code.
2. The only database remaining in the Chemistry section is the atoms database.
3. Atom indexing is now fixed and matches the indexing in the coordinate array.
4. New ChemicalSystem implementation stores lists of indices.
5. ChemicalSystem does not have a "configuration" attribute anymore. Configuration is accessed only via the Trajectory class.
6. A backup ChemicalSystem loader allows the users to load old trajectories. The new ChemicalSystem instance will be created in memory and populated with information from the old /chemical_system data structure.
7. The trajectory stores a subset of the atom database, preserving the user-made changes to atom definitions and making trajectories portable even if they use artificial atoms.
8. Atom selection works for trajectories independent of the molecular dynamics engine used to run the simulation. However, it is **necessary to detect bonds in the system using the TrajectoryEditor job** before it is possible to select structures.
9. Most of the Cython extensions have been replaced with Python equivalents.

**To test**
All unit tests must pass.
A detailed comparison of the results between this and main branches is necessary due to the fundamental changes to the way the code operates.
Please run TrajectoryEditor with **molecule detection** on a trajectory and check if all the expected Atom Selection options are working correctly in the GUI.

Jobs that require the most attention:
- CenterOfMassTrajectory
- Voronoi
- SolventAccessibleSurface
- PairDistributionFunction
- VanHoveSelf
- VanHoveDistinct


ChiCheng45:
I'm going through the code and program. I'll keep posting any issues I find.

I found a bonding issue in the 3D viewer, bonds are being created between the dummy atoms and non-dummy atoms.

image

ChiCheng45:
I've been testing the solvent SolventAccessibleSurface job. According to the MDTraj documentation https://mdtraj.org/1.9.4/examples/solvent-accessible-surface-area.html they also use the method by Shrake and Rupley. Here I run the SolventAccessibleSurface with a trajectory from SlimMD https://www.ccpbiosim.ac.uk/slim-md/slimmd-database/proteins/cyclosporin-a-37ms. 

Here are the results for the first 1000 time steps.

Protos.
image

This Branch.
image


MDTraj.
image


Concentrating on the timestep at 550 (top) and 670 (the peak, bottom) we can see the molecule geometry change.
Animation54
Animation55

At 670 the molecule is in a more open geometry which agrees with the increase in the solvent accessible surface in MDTraj and this branch. Looks like there were some issues with the implementation of the SolventAccessibleSurface on protos.

ChiCheng45 (commented):
Since the extension are all gone we can also delete the MDANSE/Src/MDANSE/Extensions directory.

ChiCheng45 (commented):
I've tested the pair distribution function and the self and distinct part of the van Hove function. Here are the timing with the dftb water trajectory, I think we should try to speed up those numpy functions some more if possible. 

Before
pair distribution function:  0.7s
van Hove function self: 37s
van Hove function distinct: 20s 

After
pair distribution function: 4s
van Hove function self: 33s
van Hove function distinct: 1m 55s

The self part of the van Hove function looks good but I see differences for the distinct part. Looks like the intra and intermolecular parts haven't been split up. 

Here is the the inter and intra PDF for the dftb water trajectory with this branch.
image

Here is the one on protos.
image

It looks like the bonding information got lost when the DFTB trajectory was converted.

ChiCheng45:
I've found one issue with the forcite nylon66_rho100_500K_v300K trajectory. The trajectory converts without any issues but there are some errors when it is loaded up in the 3D viewer. Apparently, there are inf or nan values in the coordinates.

```
    tree = KDTree(rs)
           ^^^^^^^^^^
  File "_ckdtree.pyx", line 564, in scipy.spatial._ckdtree.cKDTree.__init__
ValueError: data must be finite, check for nan or inf values

```

ChiCheng45:
I've found one issue when I use the LAMMPS converter with the glycyl_L_alanine_charmm trajectory from MDANSE-Examples.

Here is a the trajectory at the 0th and 12th time step on protos.
imageimage


Here is the same trajectory on this branch.
imageimage

It looks like the atom ordering might have gotten mixed up compare, for example, the nitrogen atoms.

ChiCheng45:
I found that compared to Protos, the molecule information gets lost. On Protos, when I run the DFTB converter and select a job that requires molecule information, such as the angular correlation, the molecule name is detected with the name H2O1. On this branch, the molecule information appears to have gotten lost, when I select angular correlation the molecule setting doesn't get filled.

MBartkowiakSTFC:
> I found that compared to Protos, the molecule information gets lost. On Protos, when I run the DFTB converter and select a job that requires molecule information, such as the angular correlation, the molecule name is detected with the name H2O1. On this branch, the molecule information appears to have gotten lost, when I select angular correlation the molecule setting doesn't get filled.

There is a question of what the right approach should be here. If I run TrajectoryEditor with molecule detection option, I will end up with a trajectory that has both molecules and bonds defined. Some of the converters would give me grouping of atoms into molecules, but no bonds. In the MDANSE School tutorials we had the example of DL_POLY trajectories where AgI molecules are defined, even though physically there are no chemical bonds.

The question now is if we decide to trust the MD engine information, which may be provided or not, or just consistently require the molecule detection to be performed for every trajectory.

MBartkowiakSTFC:
I think that all the problems you highlighted so far have been fixed in the latest version of this PR. Also, now the MDTraj converter checks for bonds in the topology files, and uses them to identify molecules.

ChiCheng45:
> I think that all the problems you highlighted so far have been fixed in the latest version of this PR. Also, now the MDTraj converter checks for bonds in the topology files, and uses them to identify molecules.

Thanks, yes looks good as far as I can tell all the converters are working correctly now.

One bug I've noticed is that when you delete an open trajectory from the GUI, the file still appears to be locked so I can't delete or move the file in Windows. Previously we were able to do this and we can still do this with the MDA files.

ChiCheng45:
> > I think that all the problems you highlighted so far have been fixed in the latest version of this PR. Also, now the MDTraj converter checks for bonds in the topology files, and uses them to identify molecules.
> 
> Thanks, yes looks good as far as I can tell all the converters are working correctly now.
> 
> One bug I've noticed is that when you delete an open trajectory from the GUI, the file still appears to be locked so I can't delete or move the file in Windows. Previously we were able to do this and we can still do this with the MDA files.

Looks like this is fixed now.

ChiCheng45:
COM trajectory to work well and analysis calculation with COM trajectory also seems to be working fine.

Just a few things about COM atoms and the 3D viewer. I ran the the COM trajectory on the discover cubane/buckyball system.

image

this produces the following trajectory 

image

the atom radii are maybe too big? We also need to block the bonding between COM atoms similarly to dummy atoms, here is what I see when I set a smaller atom radius.

image

We also need to think about what the radius setting should mean exactly because it could mean vdw radii or covalent radii or something else.

ChiCheng45:
I'm getting some odd trajectories after running the trajectory editor with default values so no changes to the trajectory should occur. I'm using the NAMD dmpc_in_water trajectory.

Before
image

After (this branch)
image

After (protos)
image

MBartkowiakSTFC:
> COM trajectory to work well and analysis calculation with COM trajectory also seems to be working fine.
> 
> Just a few things about COM atoms and the 3D viewer. I ran the the COM trajectory on the discover cubane/buckyball system.
> 
> the atom radii are maybe too big? We also need to block the bonding between COM atoms similarly to dummy atoms, here is what I see when I set a smaller atom radius.
> 
> We also need to think about what the radius setting should mean exactly because it could mean vdw radii or covalent radii or something else.

Screenshot 2025-01-20 at 16 05 38
In the end I decided to define the radius of a COM atom as the SUM of two terms:
1. the mass-weighted average of the atom distances from the centre of mass,
2. the mean covalent radius of the atoms in the groups.
The atom size scaling is still less than 1 (0.4 by default, so atoms don't block the view). If I change the atom scaling to 1.0, for the fullerene/cubane system I get the picture above.

MBartkowiakSTFC:
> I'm getting some odd trajectories after running the trajectory editor with default values so no changes to the trajectory should occur. I'm using the NAMD dmpc_in_water trajectory.
> 

Two problems had to be addressed:
1. MinimalPDBReader was grouping molecules wrong, adding a water molecule to a chain,
2. contiguous_configuration normally shifts atoms to minimise their distance from the _first_ atom in the group. I added an option to group them around the _centre_ of the group instead. (Not the centre of mass, just the geometric centre). For now I made TrajectoryEditor use that option, as it is less likely to break long chains. We can discuss if we should always use this option or not. For now I left it as "false" to stay compatible with the old code.
If you convert the trajectory again and run a null trajectory editor, nothing should change.

ChiCheng45:
I'm not sure how this happened since the PDF function is the same on protos and this branch but there are some issues wth the SSF calculation on this branch. This worked before, I think d500f10d14013a409828485ee615a6fff2d86ebd should be the first bad commit.

image

image

MBartkowiakSTFC:
> I'm not sure how this happened since the PDF function is the same on protos and this branch but there are some issues wth the SSF calculation on this branch. This worked before, I think [d500f10](https://github.com/ISISNeutronMuon/MDANSE/commit/d500f10d14013a409828485ee615a6fff2d86ebd) should be the first bad commit.
> 
I completely missed that SSF also inherits from DistanceHistogram. PDF needed a change of one scaling factor in this PR. Now SSF has it too. For the trajectories I checked, the results matched those from protos.

ChiCheng45:
Some differences between the RMSD results on protos and this branch.

image

Also the RMSD calculation is quite a bit slower on this branch.

ChiCheng45:
RMSF job on this branch does not give the correct number of data points with the DFTB water trajectory and the molecule group selection.

On this branch there are only two data points, there should be 100 since there are 300 atoms. Also even with the atom group selection the results are different.

image

ChiCheng45:
Here is the EISF for the VASP trajectory, protos and this branch produce the same results.

image

Here is the EISF for the DFTB trajectory, protos and this branch produce different results.

image


I made sure the set the seed to 1 so they should generate the same q-vectors. I think there is some issue with trajectories with atoms that go outside of the unit cell for some calculations.

DFTB - some atoms go outside the unit cell
image

VASP trajectory - all atoms stay inside the unit cell
image

I think this is whats causing the RMSD and RMSF issues above. I used the VASP trajectory to test quite a few of the MDANSE jobs so there are probably a few more that do not work with the DFTB trajectory.

MBartkowiakSTFC:
I have compared the results from protos to the current branch. It was correct for:
- EISF, DISF
- MSD atoms and molecules grouping,
- RMSD, RMSF,
- DOS, VACF, PACF
- angular correlation,
- infrared.
DFTB water and LAMMPS CuSbS trajectories were used in the tests, since one has molecules going out of the box, and the other has clear jumps across the simulation boundary.

I think that the read_com_trajectory method in MdanseTrajectory was the main source of problems before, and it has been fixed now.

ChiCheng45:
I've done some more testing and it looks like there might be some issues with loading h5md files in this branch. I'm using MDANSE\Tests\UnitTests\Data\Ar_mdmc_h5md.h5.

When I load the h5md file I also get the following error.

"KeyError: "Unable to synchronously open object (object 'chemical_system' doesn't exist)""

but it seems to continue to work anyway.

Anyway here are the following results for PACF, RMSD, and RMSF.

image
image
image

MBartkowiakSTFC:
> I've done some more testing and it looks like there might be some issues with loading h5md files in this branch. I'm using MDANSE\Tests\UnitTests\Data\Ar_mdmc_h5md.h5.
> 
> When I load the h5md file I also get the following error.
> 
> "KeyError: "Unable to synchronously open object (object 'chemical_system' doesn't exist)""
> 
> but it seems to continue to work anyway.

True, this message was not really meant to be this dramatic. Now there is just a warning that MdanseTrajectory class could not read the file, so we will try H5MD instead.

> 
> Anyway here are the following results for PACF, RMSD, and RMSF.
> 

I corrected it, now the results match for me.

ChiCheng45:
> > > I think that all the problems you highlighted so far have been fixed in the latest version of this PR. Also, now the MDTraj converter checks for bonds in the topology files, and uses them to identify molecules.
> > 
> > 
> > Thanks, yes looks good as far as I can tell all the converters are working correctly now.
> > One bug I've noticed is that when you delete an open trajectory from the GUI, the file still appears to be locked so I can't delete or move the file in Windows. Previously we were able to do this and we can still do this with the MDA files.
> 
> Looks like this is fixed now.

This is fixed when you load an already converted trajectory but I found one more edge case. When you convert with autoload and then delete it from the GUI the trajectory file is still locked so it can't be deleted or moved.

MBartkowiakSTFC:
> 
> This is fixed when you load an already converted trajectory but I found one more edge case. When you convert with autoload and then delete it from the GUI the trajectory file is still locked so it can't be deleted or moved.

When does it go away? Do you have to switch off the GUI, or is it enough just to wait?

As far as I know this issue is Windows-specific.

ChiCheng45:
> > This is fixed when you load an already converted trajectory but I found one more edge case. When you convert with autoload and then delete it from the GUI the trajectory file is still locked so it can't be deleted or moved.
> 
> When does it go away? Do you have to switch off the GUI, or is it enough just to wait?
> 
> As far as I know this issue is Windows-specific.

Strange it seems to be working now. Could be that my windows got into a bad state or something,

ChiCheng45 (approved):
Looks great. I'm unable to find any more obvious issues. I think this should be good to merge now.

PR #617, opened 2024-12-02
General GUI optimisations and atom selection inversion button

ChiCheng45:
**Description of work**
- An invert selection button was added to the atom selection.
- Atom selection functions were optimized.
- Checkablecombo box has been optimized.
- Bond calculation in the 3D view has been optimized.
- apply_instrument has been updated so that the prediction output is only determined once.
- Changing certain settings that do not affect the output prediction will be faster as they will skip the output determination step.
- Fixed the time per frame so that the spin box is set to the same value as the value it is set to in the objects __init__. This means it doesn't suddenly change from 80 to 6 when it is changed.

According to my profiling tools, most of the time is taken up by the literal_eval calls in the ChemicalEntity objects. I think we should see a performance improvement with #616. We can make another PR to see what else can be optimized once #616 has been merged.

**To test**
- Load up a large trajectory and go to the 3D view, the trajectory animation with the bond calculation should be faster.
- Go to the actions tab and load up the atom selection helper, atom selections should be slightly more responsive. Check that the inversion button works as expected.
- Check that in a job like DISF, applying an instrument should be much faster as the prediction is only determined once.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
+++ b/MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
@@ -110,7 +110,7 @@ class Selector:
         }
 
         # figure out if a match exists for the selector function
-        self.match_exists = self.full_settings
+        self.match_exists = self.create_default_settings()
         for k0, v0 in self.match_exists.items():
             if isinstance(v0, dict):
                 for k1 in v0.keys():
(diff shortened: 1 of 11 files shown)


MBartkowiakSTFC (approved):
It works. Big improvement for the GUI responsiveness when changing fields such as the output file name.

PR #618, opened 2024-12-03
GUI optimisation - qvector and instrument resolution type change

ChiCheng45:
**Description of work**
Optimized some more parts of the GUI. When the instrument resolution or qvector types are changed their fields are changed so updateValue and the output prediction will be recalculated. I've updated the code so this does not happen. With large trajectories, changing the instrument resolution or qvector type should be a lot faster.

Fixed units tests.

**To test**
Load up a large trajectory and select an action job, change the qvector or instrument resolution type. The change should be a lot faster than before. Check that the prediction output changes with the same values as before.


MBartkowiakSTFC (approved):
The GUI performance is better now. I did not find any problems with the changes.

Issue #550, opened 2024-09-19
[ENHANCEMENT] Update the plotter so that the results can be interactively weighted, normalised and etc


ChiCheng45:
**Is your feature request related to a problem? Please describe.**
In some cases, the user may want results that are (not) weighted and/or (not) normalized and/or have a instrument resolution function applied to it for results which have an FFT applied to it e.g. f(q, t) -> s(q, w).  Currently, the user will need to run calculations multiple times to get the results they might want. 

**Describe the solution you'd like**
The normalization and weight factors should be stored in the MDA file to be applied or removed interactively while plotting the results.

**Additional context**
The feature has been mentioned in #537 and #549.

ChiCheng45:
Closing, as this is mostly done with the applied weight option and the normalisation settings in the plotter. The option to FT a result or do other operations on a result is more troublesome to implement and doesn't seem like it would be worth the effort at this moment.

PR #619, opened 2024-12-05
Improve DISF+DCSF summation

MBartkowiakSTFC:
**Description of work**
Neutron total dynamic structure factor analysis does not test if the input datasets (DCSF and DISF) were produced with the correct weights applied. As a result, the datasets frequently end up having the weights applied twice.

**Fixes**
- Re-enabled the NDTSF analysis
- Check what weights were used to produce the input datasets. If "equal" apply weights, if "b_coherent" and "b_incoherent2" don't apply weights, reject other inputs.

**To test**
Try reproducing a known result of dcsf/disf calculation and add the results to each other using the neutron total dynamic structure factor analysis.


ChiCheng45:
I found some inconsistency when different equal or b_inc2/b_coh weighs are used. Here are some examples with the Ar36 trajectory.

DCSF (equal) / DISF (equal)
image

DCSF (b_coh) / DISF (b_inc2)
image

DCSF (equal) / DISF (b_inc2)
image

DCSF (b_coh) / DISF (equal)
image

gonzalezma:
I am not sure how this is intended to work. Originally, the NTDSF analysis was introduced to compute a signal as similar as possible to the measured one, i.e. something similar to Sum_ij b_i*b_j*sqrt(c_i*c_j)*Scoh_ij + Sum_i b_i*c_i*Sinc_i, with output units in barn/sterad (or fm^2/sterad). By providing also the different contributions, it was then possible to determine the relative importance of the coherent and incoherent contributions, etc. As the DCSF and DISF are normalized, one needs to be careful when summing their results to give the NTDSF. I agree that it is possible, but my advice would be to use the partials of both analysis (so it does not matter the weighting scheme used in their calculation) and recalculate all the needed sums.

As for the issue with the Argon simulation, if the element used in the analysis is Ar36, its b_incoherent is null, so I wonder if this could not affect the tests that you are trying to do.

MBartkowiakSTFC:
@gonzalezma Let's start the discussion of this PR with this paper https://journals.aps.org/pra/abstract/10.1103/PhysRevA.6.1107 where the coherent and incoherent contributions to the total S(q,w) are described for a mixture of gases (but mainly argon). The S_inc shows the strongest signal at 10 nm^-1 and the S_coh at 20 nm^-1, the respective intensities given as 2.6 and 2.4 meV^-1. So, the coherent and incoherent contributions to S(q,w) are expected to be of comparable intensity.
Using the same scattering lengths as those in the paper, and a trajectory of argon, I can calculate the DCSF and DISF in MDANSE. The Q dependence of the signal at 0 energy transfer is shown here:
Sqw_vs_Q_separate
The maxima of the two curves are still comparable.

If I use the existing main branch to combine the DCSF and DISF, the result looks like this:
Sqw_vs_Q_protos_merged
The scale of the signal is different, and the incoherent contribution is gone.

The code proposed in this PR takes the same DCSF and DISF, and produces the following result:
Sqw_vs_Q_newbranch_merged

It appears to me that the result produced by the main branch is not correct. While it is possible that the calculation implemented in this PR is not optimal, it still seems like an improvement over the previous version. What is your opinion? What result would you expect from the NDTSF analysis in this situation?

MBartkowiakSTFC:
@ChiCheng45 I was getting a similar result at some point, but then I noticed that I was using an artificial element and the value of 'equal' in the table was set to 0 for this one. All the other elements have 'equal' set to 1. I think we should have 'equal' always evaluate as 1, and not take it from the table.

Here is my comparison of the b_coh/b_incoh2 case and the equal/equal case. The results are still different, but this could also be due to the different q vectors, insufficient trajectory length, system size, etc. Blue curve has DCSF and DISF calculated using 'equal' weights and then summed up. Orange used b_coh, b_incoh2 for DCSF and DISF.
ndtsf_from_equal

gonzalezma:
I don't know what it is implemented in the main branch, so I would need to look at the code. In MDANSE 1.5 this was an analysis on its own, while now I think that it is a "secondary" analysis that uses as input the output of the DISF and DCSF analysis that should be run before. Correct me if I am wrong.

If this is the case, I don't have anything against this approach, but I don't understand the use of the weighting. My comment was to indicate that it should not matter which weighting is used in calculating DISF and DCSF, if one uses the partials and not the total DISF and total DCSF.

As for your example, my answer to what result I would expect for the NTDSF is "it depends". E.g. if I assume that the sample is Ar36, the incoherent cross section is null, so the old result where NTDSF = DCSF, with no incoherent contribution seems right. But if you assume that you have natural Ar, then the incoherent contribution should be visible. In fact, I don't think that a monoatomic system such as Ar is the best one to test this. I would suggest using water and testing the total S(Q,w) and their corresponding coh/inc contributions for H2O and D2O.

ChiCheng45 (commented):
I've tested with the latest version and as far as I can tell the results are consistent now when you use equal or scattering length weights or a mixture of the two.
image

ChiCheng45:
Just wondering was `self._outputData["f(q,t)_coh_%s%s" % pair]` and `self._outputData["f(q,t)_inc_%s" % element]` normalized by `1/np.sqrt(ni * nj)` and `1/ni` before this part of the code?

https://github.com/ISISNeutronMuon/MDANSE/blob/8bd79099137af0895358c080076e041bb7573667/MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py#L426-L438

https://github.com/ISISNeutronMuon/MDANSE/blob/8bd79099137af0895358c080076e041bb7573667/MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py#L462-L468

MBartkowiakSTFC:
> Just wondering was `self._outputData["f(q,t)_coh_%s%s" % pair]` and `self._outputData["f(q,t)_inc_%s" % element]` normalized by `1/np.sqrt(ni * nj)` and `1/ni` before this part of the code?

In the protos branch I'm pretty sure this normalisation would be applied twice. In this PR I do not apply the normalisation to the partials. Instead, the scaling factor is calculated and stored as a dataset attribute, so the plotter can apply the scaling if needed, but the data array itself is written to the file unchanged. So, the partials read by the NDTSF job are taken not normalised, and then the normalisation by `1/np.sqrt(ni * nj)` and `1/ni` is applied.

ChiCheng45:
I think a test to check that this is working would be to generate a liquid argon trajectory and use the Ar scattering lengths to calculate the neutron total dynamic structure factor. Using the same trajectory assign the atoms to Ar36, Ar38, and Ar40 randomly in their natural abundances. There will be no incoherent scattering but this should get absorbed into the coherent part. If the neutron total dynamic structure factor from the first calculation is equal to the coherent scattering in the 2nd then this PR should be correct (we might need a big system for this to work).

Here is my reasoning for why I think this. The intermediate scattering functions are

image

image

In the second calculation, we will specify the isotopes so we use the scattering lengths for Ar36, Ar38, and Ar40. To be precise the scattering functions are now

image

image

The incoherent part disappears. However, what we have done is absorbed the incoherent part into the coherent as the coherent part contains the diagonal part as well as the off-diagonal.

image

If this system is large and the isotopes are placed randomly then I should be able to find situations where the value of the correlation functions in the summation are approx the same at least for that specific choice of q and t. 

I can now group the terms up so that I can write something like this.

image

If I change the summation over $K \neq L$ to include the diagonal part I must remove it again. 

image

which is the sum of the coherent and incoherent parts at the top. Does this make sense?

gonzalezma:
I don't know (if I can I will try to read more carefully what you wrote), but I keep thinking that Ar (or any monatomic system) is not the most appropriate choice to test this. Why don't you run the analysis on a water trajectory and compare the scattering cross sections for H2O and D2O? There are even experimental data for the S(Q), i.e. F(Q,t=0), that you can compare to.

ChiCheng45:
Here is an example of what I was trying to explain above. Here I calculate the DCSF of argon36, argon38, and a 50/50 mix of argon36 and argon38. The DISF of all three will be zero since the isotopes were specified. The 50/50 mix has atoms placed randomly in a trajectory with 2048 atoms. Here is the peak at q=10 ang^-1.

image

The 50/50 mix should give something for the incoherent scattering but we can see above that it got absorbed into the DCSF. This is analogous to how MDANSE treats chemical incoherence which is always added to the DCSF.

gonzalezma:
I am not sure of following everything, in particular what you exactly mean by chemical  incoherence, but I think that you are somewhat mixing the concepts.  As you say, if you have only 36Ar, b_inc = 0 and you will not have any incoherent scattering. 
Same for 38Ar. But I do not agree that the incoherent scattering of the 50/50 mix will be zero.

In the case of the 3 isotopes of Ar, the nuclear spin is 0 for all of them. This implies that there is no "spin incoherent" scattering, and all the incoherence must come from the random distribution of isotopes. In this case, it is easy to calculate the total and coherent scattering cross sections as (see Sears 1992 for the values) as:
sigma_coh = 4 * pi * (c(36Ar)*b(36Ar) +c(38Ar)*b(38Ar) + c(40Ar)*b(40Ar) )**2
sigma_tot = 4 * pi * (c(36Ar)*b(36Ar)**2 +c(38Ar)*b(38Ar)**2 + c(40Ar)*b(40Ar)**2)
and then
sigma_inc = sigma_tot - sigma_coh

So if we look to some possible cases, we get:

1) Pure 36Ar: sigma_tot = sigma_coh = 77.9 barn and sigma_inc = 0 (and, of course, sigma_inc = 0 also for pure 38Ar or 40Ar)

2) Natural Ar: sigma_tot = 0.683 barn, sigma_coh = 0.458 barn, sigma_inc = 0.225 barn

3) 50% 36Ar + 50% 38Ar: sigma_tot = 39.7 barn, sigma_coh = 25.3 barn, sigma_inc = 14.4 barn

So I do not understand why you say that the incoherent signal for the 50/50 mix should be zero.

By the way, this is the method that Maciej proposed in our talk yesterday to calculate the coh and inc scattering lengths for any given isotopic substitution. I agree with him that this is correct for cases like this one, where each isotope can be characterized by a single scattering length (b_coh). But I am not sure that this will work for elements where we also have nuclear spin incoherence. In such cases, I think that using only b_coh will not be enough, and one should do the same calculation as above, but using b+ and b- with the corresponding concentrations c+ and c- for each isotope having an spin I (b+ corresponding to the I+1/2 case, and b- for the I-1/2 case). Such values are available (e.g. http://www.ati.ac.at/~neutropt/scattering/ScatteringLengthsAdvTable.pdf), but they are not in the current elements database of MDANSE, as far as I know. In any case, as agreed yesterday, I do not think that this is a priority for the moment.

ChiCheng45:
> So I do not understand why you say that the incoherent signal for the 50/50 mix should be zero.

Yes, the 50/50 mix does have an incoherent signal. But in MDANSE, if you **specify each isotope manually**, you don't get one because the b_inc for each isotope is zero. This is fine because the incoherent signal will appear in the coherent calculation, I showed this in the above plot. Does this make sense?

gonzalezma:
OK, I understand now what you mean. But I am not sure that things "automatically" work, as you seem to suggest.
First of all, when you have spin incoherence (e.g. with H or D), I do not think that your reasoning applies.
And in this case (Ar), I think that it may be correct, but there is something that I do not understand in your signals. If they correspond to the total scattering (i.e. not normalized), the area under the 36Ar S(Q,w) should be about 50 times larger than the area under the 38Ar S(Q,w), as this is the ratio between their total scattering cross sections (77.9/1.5). And then, from the numbers I wrote above, the 50/50 mix should have an area about half of that of 36Ar.

ChiCheng45:
> OK, I understand now what you mean. But I am not sure that things "automatically" work, as you seem to suggest. First of all, when you have spin incoherence (e.g. with H or D), I do not think that your reasoning applies. And in this case (Ar), I think that it may be correct, but there is something that I do not understand in your signals. If they correspond to the total scattering (i.e. not normalized), the area under the 36Ar S(Q,w) should be about 50 times larger than the area under the 38Ar S(Q,w), as this is the ratio between their total scattering cross sections (77.9/1.5). And then, from the numbers I wrote above, the 50/50 mix should have an area about half of that of 36Ar.

I think it should still be the case with spin incoherence the arguments I made with the math above should still hold, we can check this with a calculation if you like.

For the second issue, I think the similarity between the 36Ar and 38Ar might be down to the weighting scheme used in MDANSE. The scattering lengths get divided by the sum of the scattering lengths of the system. For the 50/50 mix, there is incoherent scattering so we get that large peak. The plot above is not the total (doesn't have the proper scaling because of MDANSEs weight scheme) it's the result from the DCSF calculation, I wanted to show that you get incoherent scattering in the DCSF result when I set the isotopes manually with 50/50.

gonzalezma:
Not sure for the spin incoherence. I will let you do the maths and convince me ;-)

As for the second point, there are two things that I do not understand:
1. I thought that the whole point of this analysis was to give the scattering in absolute units (so no normalization by the sum of the scattering lengths).
2. What's the argument to interpret all the additional signal as incoherent scattering?

ChiCheng45:
1. I only calculated DCSF as an example to show the incoherent signal gets added to the DCSF results when isotopes are set manually, I agree once complete the total shouldn't be using rescaled scattering lengths
2. It isn't completely incoherent but should be mostly incoherent since it's low q and it is so much larger than either pure Ar36 or Ar38.

ChiCheng45:
Here is the total scattering with a trajectory with 50/50 mix of randomly set 36Ar and 38Ar and an Ar_Mix atom which I created with b_coh = 14.2E-05 and b_inc = 10.7E-05.

image

As you can see despite b_inc = 0 for both 36Ar and 38Ar, the incoherent scattering got added into the DCSF calculation. Small differences between the two are due to the short trajectory I used which means there is a bit of noise.

image

I think the above results suggest that we are adding things correctly. I've also removed the copying of the DISF and DCSF results into the total mda as I think you can end up with a lot of different results in the mda file especially when there are a lot of atom types.

I've also added a new feature so that there is a scaling setting in the plotter so that the partials are not scaled, this will mean that the partials will be the same as the old MDANSE, something similar to this was suggested in #353. This should also set it up so we can resolve #669 with #550 by using different scaling/weight schemes.

image

image

Here I used equal weights.

MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
```diff
@@ -109,20 +109,24 @@ def configure(self, value):
...
                 else:
-                    weights[name] = p
+                    weights[name] = property
 
-        for k, v in list(ascfg.get_natoms().items()):
-            weights[k] /= v
+        for element, num_atoms in list(
+            atom_selection_configurator.get_natoms().items()
+        ):
+            weights[element] /= num_atoms
```
oerc0122: Don't need to cast to list
```suggestion
        for element, num_atoms in atom_selection_configurator.get_natoms().items():
            weights[element] /= num_atoms
```

MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
```diff
@@ -109,20 +109,24 @@ def configure(self, value):
...
+        ]
 
         weights = {}
-        for i in range(ascfg["selection_length"]):
-            name = ascfg["names"][i]
-            for el in ascfg["elements"][i]:
-                p = self._trajectory.get_atom_property(el, self["property"])
+        for i in range(atom_selection_configurator["selection_length"]):
+            name = atom_selection_configurator["names"][i]
+            for element in atom_selection_configurator["elements"][i]:
```
oerc0122: Bit verbose due to length of names, but maybe clearer?
```suggestion
        for name, elements in itertools.islice(
            zip(atom_selection_configurator["names"],
                atom_selection_configurator["elements"]),
            atom_selection_configurator["selection_length"]
        ):
            for element in elements:
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
```diff
@@ -251,34 +251,35 @@ def finalize(self):
...
-            self._outputData[f"f(q,t)_{pair_str}"][:] /= np.sqrt(ni * nj)
+            extra_scaling = 1.0 / np.sqrt(ni * nj)
+            self._outputData[f"f(q,t)_{pair_str}"].scaling_factor *= extra_scaling
             self._outputData[f"s(q,f)_{pair_str}"][:] = get_spectrum(
                 self._outputData[f"f(q,t)_{pair_str}"],
                 self.configuration["instrument_resolution"]["time_window"],
                 self.configuration["instrument_resolution"]["time_step"],
                 axis=1,
             )
+            self._outputData["s(q,f)_%s%s" % pair].scaling_factor *= extra_scaling
```
oerc0122:
```suggestion
            self._outputData[f"s(q,f)_{pair_str}"].scaling_factor *= extra_scaling
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -253,33 +253,31 @@ def finalize(self):
...
-            self._outputData[f"f(q,t)_{element}"][:] /= number
+            extra_scaling = 1.0 / number
+            self._outputData[f"f(q,t)_{element}"].scaling_factor *= extra_scaling
             self._outputData[f"s(q,f)_{element}"][:] = get_spectrum(
                 self._outputData[f"f(q,t)_{element}"],
                 self.configuration["instrument_resolution"]["time_window"],
                 self.configuration["instrument_resolution"]["time_step"],
                 axis=1,
             )
+            self._outputData["s(q,f)_%s" % element].scaling_factor *= extra_scaling
```
oerc0122:
```suggestion
            self._outputData[f"s(q,f)_{element}"].scaling_factor *= extra_scaling
```

MDANSE/Src/MDANSE/Framework/Jobs/GeneralAutoCorrelationFunction.py
```diff
@@ -168,12 +168,16 @@ def finalize(self):
                 if self._outputData[f"gacf_{element}"][0] == 0:
                     raise ValueError("The normalization factor is equal to zero")
                 else:
```
oerc0122: Don't need `else` after `raise`

MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
```diff
@@ -394,58 +400,33 @@ def finalize(self):
...
+            self._outputData["f(q,t)_coh_total"][:] += self._outputData[
+                f"f(q,t)_coh_{pair_str}"
+            ][:]
+            self._outputData["s(q,f)_coh_total"][:] += self._outputData[
+                f"s(q,f)_coh_{pair_str}"
+            ][:]
 
         # Compute incoherent functions and structure factor
-        for element, ni in nAtomsPerElement.items():
+        for element in nAtomsPerElement.keys():
```
oerc0122: If this is a normal dict, `keys` is default.
```suggestion
        for element in nAtomsPerElement:
```

MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py
```diff
@@ -194,6 +187,23 @@ def finalize(self):
             main_result=True,
         )
 
+        if self.configuration["normalize"]["value"]:
+            for element in nAtomsPerElement.keys():
+                if self._outputData[f"pacf_{element}"][0] == 0:
+                    raise ValueError("The normalization factor is equal to zero !!!")
+                else:
```
oerc0122: Don't need `else` after `raise`

MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py
```diff
@@ -194,6 +187,23 @@ def finalize(self):
...
+                    self._outputData[
+                        f"pacf_{element}"
+                    ].scaling_factor *= normalisation_factor(
+                        self._outputData[f"pacf_{element}"], axis=0
+                    )
+            if self._outputData["pacf_total"][0] == 0:
+                raise ValueError("The normalization factor is equal to zero !!!")
+            else:
```
oerc0122: As above

MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py
```diff
@@ -194,6 +187,23 @@ def finalize(self):
             main_result=True,
         )
 
+        if self.configuration["normalize"]["value"]:
+            for element in nAtomsPerElement.keys():
```
oerc0122:
```suggestion
            for element in nAtomsPerElement:
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -13,29 +13,50 @@
...
 import itertools
 
 import numpy as np
 
 
-def get_weights(props, contents, dim):
+def get_weights(props: Dict[str, float], contents: Dict[str, int], dim: int):
+    """Calculates the scaling factors to be applied to output datasets
+    of an analysis. Returns an dictionary of scaling factors, where the
+    chemical elements identifying each dataset are the keys.
```
oerc0122:
```suggestion
    """Calculate the scaling factors to be applied to output datasets. 
    
    Returns a dictionary of scaling factors, where the
    chemical elements identifying each dataset are the keys.
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -13,29 +13,50 @@
...
+    Returns
+    -------
+    Tuple(Dict[Tuple[str], float], float)
+        Dictionary of scaling factors per dataset key, and a sum of all the factors
+    """
     normFactor = None
 
     weights = {}
 
     cartesianProduct = set(itertools.product(list(props.keys()), repeat=dim))
```
oerc0122: Doesn't need to be cast as a `set`, `product` already unique.
```suggestion
    cartesianProduct = itertools.product(props, repeat=dim)
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -13,29 +13,50 @@
...
 
-        weights[elements] = np.float64(np.copy(fact))
+        weights[elements] = np.float64(np.copy(factor))
 
         if normFactor is None:
-            normFactor = fact
+            normFactor = factor
         else:
-            normFactor += fact
+            normFactor += factor
```
oerc0122: Why not just `normFactor = 0` at the start?

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -46,30 +67,85 @@ def get_weights(props, contents, dim):
...
+    Returns
+    -------
+    np.ndarray
+        total sum of all the component arrays scaled by their weights
+    """
+    matches = {key % k: k for k in weights if k not in ["sum"]}
+    dim = key.count("%s")
+
+    for k in values.keys():
         if k not in matches:
```
oerc0122:
```suggestion
    for k in values.keys() & matches:
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -46,30 +67,85 @@ def get_weights(props, contents, dim):
...
             continue
 
         if symmetric:
             permutations = set(itertools.permutations(matches[k], r=dim))
-            w = sum([weights.pop(p) for p in permutations])
+            w = sum([weights[p] for p in permutations])
```
oerc0122:
```suggestion
            permutations = itertools.permutations(matches[k], r=dim)
            w = sum(weights[p] for p in permutations)
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -46,30 +67,85 @@ def get_weights(props, contents, dim):
...
+        total sum of all the component arrays scaled by their weights
+    """
+    weightedSum = None
+    matches = {key % k: k for k in weights if k not in ["sum"]}
+
+    for k, val in values.items():
+        if k not in matches:
+            continue
 
         if weightedSum is None:
```
oerc0122: Again, `weightedSum = 0` should do, right?

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -46,30 +67,85 @@ def get_weights(props, contents, dim):
...
+    -------
+    np.ndarray
+        total sum of all the component arrays scaled by their weights
+    """
+    weightedSum = None
+    matches = {key % k: k for k in weights if k not in ["sum"]}
+
+    for k, val in values.items():
+        if k not in matches:
+            continue
```
oerc0122:
```suggestion
    for val in (val for key, val in values.items() if key in matches):
```

oerc0122: In fact, this entire thing could be
```python
weightedSum = sum(val * val.scaling_factor
                  for key, val in values.items() if key in matches)
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -46,30 +67,85 @@ def get_weights(props, contents, dim):
...
+    key : str
+        A string data set name with formatting elements (placeholders for chemical element labels)
+
+    Returns
+    -------
+    np.ndarray
+        total sum of all the component arrays scaled by their weights
+    """
+    weightedSum = None
+    matches = {key % k: k for k in weights if k not in ["sum"]}
```
oerc0122: This doesn't need to be a `dict` if we're not checking any values in it, can just be a `set`

MDANSE/Src/MDANSE/Mathematics/Signal.py
```diff
@@ -139,6 +139,16 @@ def normalize(x, axis=0):
     return nx
 
 
+def normalisation_factor(x, axis=0):
```
oerc0122: No docstring

MDANSE/Tests/UnitTests/Analysis/test_scattering.py
```diff
@@ -105,10 +108,15 @@ def test_dcsf(traj_info, qvector_grid):
     assert path.isfile(temp_name + ".mda")
 
     result_file = os.path.join(result_dir, f"dcsf_{traj_info[0]}.mda")
-    with h5py.File(temp_name + ".mda") as actual,  h5py.File(result_file) as desired:
-        keys = [i for i in desired.keys() if any([j in i for j in ["f(q,t)", "s(q,f)"]])]
+    with h5py.File(temp_name + ".mda") as actual, h5py.File(result_file) as desired:
+        keys = [
+            i for i in desired.keys() if any([j in i for j in ["f(q,t)", "s(q,f)"]])
+        ]
```
oerc0122: Slightly more explicit expectations? Also similarly elsewhere?
```suggestion
        keys = [
            key for key in desired.keys() 
            if any(key.startswith(j) for j in ["f(q,t)", "s(q,f)"])
        ]
```

MDANSE/Tests/UnitTests/TrajectoryEditor/test_editor.py
```diff
@@ -98,9 +100,14 @@ def test_editor_frames():
     original.close()
     changed.close()
 
-    with h5py.File(temp_name + ".mdt") as actual,  h5py.File(short_traj) as desired:
-        np.testing.assert_array_almost_equal(actual["/configuration/coordinates"], desired["/configuration/coordinates"][0:501:10])
-        np.testing.assert_array_almost_equal(actual["/time"], desired["/time"][0:501:10])
+    with h5py.File(temp_name + ".mdt") as actual, h5py.File(short_traj) as desired:
```
oerc0122, line 103: Thinking we probably want a test-global `compare_mdt` function we can use here and elsewhere. Not this PR.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -145,6 +151,12 @@ def longest_axis(self) -> str:
                 best_axis = aname
         return best_unit, best_axis
 
+    @property
+    def data(self):
```
oerc0122, line 155: Docstring

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -213,15 +236,21 @@ def planes_vs_axis(self, axis_number: int) -> List[np.ndarray]:
                     fixed_argument = perpendicular_axis[plane_number]
                     slice_def[axis_number] = plane_number
                     data = self._data[tuple(slice_def)]
-                    self._planes[plane_number] = data
+                    if self._use_scaling:
+                        self._planes[plane_number] = data * self._scaling_factor
```
oerc0122: Wonder if some kind of `with_scaling` function might be nicer and eliminate a lot of this `if self._use_scaling`. 

Could also make a `self.scaling_factor` `@property` which returns `1` if `not self._use_scaling`?

Alternatively, could consider tertiary assign
```python
self._planes[plane_number] = data * self._scaling_factor if self._use_scaling else data
```
or
```python
self._planes[plane_number] = data * (self._scaling_factor if self._use_scaling else 1)
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -406,6 +442,11 @@ def datasets(self) -> Dict:
                 result[key] = (self._datasets[key], colour, style, marker, ds_num, axis)
             else:
                 self._datasets[key]._data_limits = None
+            if set_scaling:
+                self._datasets[key]._use_scaling = True
+            else:
+                self._datasets[key]._use_scaling = False
```
```suggestion
            self._datasets[key]._use_scaling = set_scaling
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/DataPlotter.py
```diff
@@ -68,14 +68,16 @@ def __init__(self, *args, unit_lookup=None, **kwargs):
...
+            self._selection_viewer.hideColumn(col_num)
 
     @Slot(object)
     def add_dataset(self, dataset: SingleDataset):
         if not dataset._valid:
             return
         self._model.add_dataset(dataset)
         self._selection_viewer.resizeColumnsToContents()
-        for col_num in range(4, 9):
+        for col_num in range(4, 10):
```
oerc0122, line 83: Is there particular meaning to `4, 10`? Should this be a constant?

MDANSE/Tests/UnitTests/TrajectoryEditor/test_editor.py
ChiCheng45: Yea I think we should do this.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/DataPlotter.py
ChiCheng45: This just hides specific columns in the PlottingContext table.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
ChiCheng45: Added

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
ChiCheng45: I've refactored the codes to clean this up a bit.

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
ChiCheng45: The change from 

```
permutations = set(itertools.permutations(matches[k], r=dim))
```

to

```
permutations = itertools.permutations(matches[k], r=dim)
```

causes some test failures because matches[k] could be ("S", "S") which will give ("S", "S") twice which we don't want. I've left in the original form.

MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
```diff
@@ -114,18 +114,21 @@ def get_weights(self):
...
+                atom_selection_configurator["elements"],
+            ),
+            atom_selection_configurator["selection_length"],
+        ):
+            for element in elements:
                 property = self._trajectory.get_atom_property(element, self["property"])
                 if name in weights:
                     weights[name] += property
                 else:
                     weights[name] = property
```
oerc0122: This form makes it more obvious that what might be happening here is:
```suggestion
            weights[name] = sum(self._trajectory.get_atom_property(element, self["property"])
                                for element in elements)
```

oerc0122 (approved):
Apart from a minor hopefully clearer (hopefully correct) stylistic thing, if this does what it sets out to functionality-wise, I'm happy for it to go in.

MBartkowiakSTFC:
I have tried out the new behaviour of the GUI. We can still load the old results (i.e. from previous versions) and the new results have the extra option of reverting to the unweighted numbers, which is good. Thanks for the changes!

MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
```diff
@@ -113,20 +114,18 @@ def get_weights(self):
             self._dependencies["atom_selection"]
         ]
 
-        weights = {}
+        weights = defaultdict(lambda: 0.0)
```
oerc0122: This will work too
```suggestion
        weights = defaultdict(float)
```

MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
ChiCheng45: I've added the change but needed to update it slightly see [43d4118](https://github.com/ISISNeutronMuon/MDANSE/pull/619/commits/43d41186c31e556519b217a85170fa44840bc32b).

MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
ChiCheng45: Ah right, thanks I have updated this.

PR #622, opened 2024-12-18
Add MDTraj converter and update to atom mapping code

ChiCheng45:
**Description of work**
Added MDTraj converter and optimized the atom mapping code. 

Similarly to the MDAnalysis converter, the MDTraj converter converts trajectories with trajectory files and a topology file. Trajectory files can include one or more files but should only be of a single type e.g. xtc. In MDTraj topology files are optional if the trajectory files contain topology information.

**To test**
1) Convert some gromacs trajectories from the MDANSE-Examples repository. 
Enter PDB files into the topology file setting and trr or xtc files into the trajectory files setting and convert the trajectory. Atom mapping should be generated faster than before.
2) Convert some charmm trajectories from the MDANSE-Examples repository. 
Enter PDB files into the topology file setting and dcd files into the trajectory files setting and convert the trajectory.
3) Try converting a trajectory using multiple coordinate files.
Either generate multiple coordinate files or copy and rename them. Convert the trajectory, if you copy and renamed it then the MDANSE trajectory should be twice as long and repeat itself.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/AtomMapping/atom_mapping.py
+++ b/MDANSE/Src/MDANSE/Framework/AtomMapping/atom_mapping.py
@@ -39,6 +39,9 @@ class AtomLabel:
         if self.grp_label == other.grp_label and self.atm_label == other.atm_label:
             return True
 
+    def __hash__(self):
+        return hash((self.atm_label, self.grp_label, self.mass))
+
 
 def guess_element(atm_label: str, mass: Union[float, int, None] = None) -> str:
     """From an input atom label find a match to an element in the atom
@@ -179,6 +182,29 @@ def fill_remaining_labels(
             mapping[grp_label][atm_label] = guess_element(atm_label, label.mass)
 
 
+def mapping_to_labels(mapping: dict[str, dict[str, str]]) -> list[AtomLabel]:
+    """Converts the mapping back into a list of labels.
+
+    Parameters
+    ----------
+    mapping : dict[str, dict[str, str]]
+        The atom mapping dictionary.
+
+    Returns
+    -------
+    list[AtomLabel]
+        List of atom labels from the mapping.
+    """
+    labels = []
+    for grp_label, atm_map in mapping.items():
+        kwargs = {}
+        for k, v in [i.split("=") for i in grp_label.split(";")]:
+            kwargs[k] = v
+        for atm_label in atm_map.keys():
+            labels.append(AtomLabel(atm_label, **kwargs))
+    return labels
+
+
 def check_mapping_valid(mapping: dict[str, dict[str, str]], labels: list[AtomLabel]):
     """Given a list of labels check that the mapping is valid.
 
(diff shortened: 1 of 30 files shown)


MBartkowiakSTFC:
One problem that is worth looking at is the time step of the trajectory.

I tried converting the hDAT_mdma trajectory from SlimMD database. The atom types, positions and unit cell dimensions were correct, but the time axis should be checked. With the MDTraj converter I get
```
Frame times (1st, 2nd, ..., last) in ps:
0.0, 1.0, ..., 105.0
```
and with MDAnalysis
```
Frame times (1st, 2nd, ..., last) in ps:
0.0, 0.04888821, ..., 5.13326205
```

ChiCheng45:
Those are the timesteps from MDTraj, I guess the default must be 1ps. I think I'll add a widget to let the user change the timestep if they want to, similarly to the MDAnalysis job.

ChiCheng45:
I've updated the MDTraj converter to add a time step setting which should update similarly to the MDAnalysis one when files are inputted. This is ready to be reviewed again.

MBartkowiakSTFC:
Now my GUI has the input field for time step, but it does not seem to find the same value that MDAnalysis found. Also, I manually changed the value of the time step from 1.0 to 0.048, but when I loaded the resulting trajectory, it still had 1.0 step, independent of what I set.

ChiCheng45:
> Now my GUI has the input field for time step, but it does not seem to find the same value that MDAnalysis found. Also, I manually changed the value of the time step from 1.0 to 0.048, but when I loaded the resulting trajectory, it still had 1.0 step, independent of what I set.

Thanks for that. It should be fixed now so it will pick up the timestep setting. 

I'm not sure why MDTraj doesn't pick up the same numbers as MDAnalysis but I'm now a bit confused about why MDAnalysis picked up the 0.048 ps timestep when according to https://www.ccpbiosim.ac.uk/slim-md/slimmd-database/protein-membrane/human-dopamine-transporter-embedded-in-a-membrane-bound-to-mdma-250ns it should be 2.5 ns.

ChiCheng45:
Apparently, DCD files are a bit of a nightmare https://github.com/mdtraj/mdtraj/pull/1163. I think we might just have to accept that MDAnalysis and MDTraj might suggest different timesteps and leave it to the user to set a correct one.

Also, the 0.048 seems to come from some timefactor see https://github.com/mdtraj/mdtraj/pull/1163#issuecomment-237337314.

MBartkowiakSTFC:
> I'm not sure why MDTraj doesn't pick up the same numbers as MDAnalysis but I'm now a bit confused about why MDAnalysis picked up the 0.048 ps timestep when according to https://www.ccpbiosim.ac.uk/slim-md/slimmd-database/protein-membrane/human-dopamine-transporter-embedded-in-a-membrane-bound-to-mdma-250ns it should be 2.5 ns.

According to our own DCD code, the DCD file stores time values in AKMA time units, which are also defined in MDAnalysis: https://userguide.mdanalysis.org/stable/units.html
If you divide 1 ps by 20.4548, you get the 0.04888 value. Of course, this only raises more questions. Why divide 1 ps by this factor? Is the information about the time values even in this DCD file?

Getting back to the main topic: thanks for the changes!

MBartkowiakSTFC:
Another problem I have is that my atom mapping values are ignored by the converter. I tried converting a GROMACS trajectory of DNA from SlimMD. I changed some of the dummy atoms to Cr, and some of the phosphorus to Pb208 isotope, but the output trajectory still had the original composition.

ChiCheng45:
> Another problem I have is that my atom mapping values are ignored by the converter. I tried converting a GROMACS trajectory of DNA from SlimMD. I changed some of the dummy atoms to Cr, and some of the phosphorus to Pb208 isotope, but the output trajectory still had the original composition.

Thanks, nice one. I think it should be sorted now.

MBartkowiakSTFC (approved):
Now that we can set the time step and correct the atom types, I think that the converter works fine.

PR #623, opened 2024-12-18
Fix tab switching performance issues.

ChiCheng45:
**Description of work**
Updated test_file_outputs so that it only updates the output file widgets. Previously it would update all widgets which could be slow for large trajectories.

**To test**
Test that the output file widgets are correctly blocked/unblocked when the MDA/MDT files are loaded or unloaded into the GUI.


MBartkowiakSTFC (approved):
I did not find any problems. Works fine.

PR #626, opened 2025-01-02
Fixes runtime error seen after the GUI is closed when jobs have been terminated.

### LINKED ISSUE #502: [BUG] RuntimeError after GUI close
**Description of the error**
The following error is received when you run MDANSE and then close the GUI. This only occurs when you terminate a job from the GUI.

```
Exception ignored in atexit callback: <function shutdown at 0x000002AC56EB20C0>
  File "C:\Users\xcb63893\AppData\Local\anaconda3\envs\MDANSE\Lib\logging\__init__.py", line 2263, in shutdown
    if getattr(h, 'flushOnClose', True):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: wrapped C/C++ object of type JobEntry has been deleted
```

**Additional details**
A quick Google of the error message show that is something related to PyQt.

ChiCheng45:
**Description of work**
Fixes the runtime error seen after the GUI is closed when a job has been terminated.

**Fixes**
Fixes #502

**To test**
Run any job in single or multicore mode, the MDANSE Logs text box in the running job tab should update as usual. Run another job, terminate it, and close the GUI. No runtime error should be seen in the console.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
@@ -89,7 +89,25 @@ class JobThread(QThread):
                 self._job_comm.status_update(status_update)
 
 
-class JobEntry(Handler, QObject):
+class JobLogHandler(Handler):
+
+    def __init__(self):
+        super().__init__()
+        self.records = []
+
+    def msgs_and_levels(self):
+        msgs = []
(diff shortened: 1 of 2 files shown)


MBartkowiakSTFC (approved):
It took me a while to reproduce the problem properly, but in the then this PR fixed it as intended. Also, the code has become a bit cleaner as a result. Thanks!

PR #631, opened 2025-01-06
Output files setting fix

### LINKED ISSUE #627: [BUG] Directories keep getting created when the output file setting is changed
**Description of the error**
Whenever the output files setting is manually changed in the GUI, multiple directories can be if those directories do not exist.

For example if I type,

C:\Users\user\Desktop\abcdefg\test1.mdt

in the output files box then the following directories are created on my desktop.

image

### LINKED ISSUE #630: [BUG] Stop users from entering reserved filenames and illegal characters on Windows for output files.
**Description of the error**
On Windows, some filenames are reserved, for example, CON, PRN, and AUX. There are also some illegal characters such as *. Currently, these can be entered into the output files widget but this will cause the job to fail. 

We should probably update the configurator or the Qt widget to block these file names/paths from being used.

ChiCheng45:
**Description of work**
Updated the output file configurations so that directories are not created each time their configure methods are called. File configurations now also check the filename can be created so that, for example, on Windows reserved filenames cannot be used. The data view has been updated so that files can now be saved into nested directories that do not exist yet and are also blocked from saving into files with file paths that are not valid.

**Fixes**
Fixes #627
Fixes #630

**To test**
Run the trajectory converter and manually edit the output file directory. No directories should be created until the job is run and finished. The configurator should continue to block the user from writing to protected directories. For example, on Windows, I am unable to edit the C:\Users directory without admin privileges. Check that the analysis jobs and average structure jobs work as usual. 

Check the save .csv file with the data view plot works with nested directories that do not exist yet, they should get created when the file is saved. Check that you are blocked from saving to invalid filepaths, for example on windows CON is a reserved filename.

### DIFF
--- a/MDANSE/Src/MDANSE/Core/Platform.py
+++ b/MDANSE/Src/MDANSE/Core/Platform.py
@@ -22,6 +22,8 @@ import inspect
 import os
 import re
 import subprocess
+import tempfile
+
 
 from MDANSE.Core.Error import Error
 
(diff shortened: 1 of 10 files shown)


MBartkowiakSTFC:
The changes worked as expected. The directories get created only on write, and there can be several levels of nesting that get created. Average structure, converters and analysis jobs work correctly.

One thing that could be added for consistency is that at the moment the file output widget in the plotter's text view (for saving .csv files) does not warn/stop the user if the chosen path is not writeable.

ChiCheng45:
> The changes worked as expected. The directories get created only on write, and there can be several levels of nesting that get created. Average structure, converters and analysis jobs work correctly.
> 
> One thing that could be added for consistency is that at the moment the file output widget in the plotter's text view (for saving .csv files) does not warn/stop the user if the chosen path is not writeable.

Thanks, I've added the suggestion to this PR. The user should be blocked from saving to invalid filenames.
image
Here I try to save to a file with a filename con which is a reserved filename on Windows.

MBartkowiakSTFC (approved):
Thanks for the changes. I think that the current version is working well.

PR #632, opened 2025-01-07
Fixed dl-poly gradient conversion fix

### LINKED ISSUE #628: [BUG] DLPOLY trajectory with gradient information conversion failure.
**Description of the error**
Failure when DLPOLY trajectory with gradient information is converted.

**Additional details**
Convert for example the FIELD_water_with_gradients/HISTORY_water_with_gradients files from the MDANSE examples repo.

ChiCheng45:
**Description of work**
Fixed the dl-poly gradient conversion bug and added a test.

**Fixes**
Fixes #628

**To test**
Convert a few dl-poly trajectories and some with gradient information. All trajectories should convert without issues.


MBartkowiakSTFC:
Works fine, solved the problem. One question: does/should the new unit test check specifically if the gradients array is created in the output trajectory file?

ChiCheng45:
> Works fine, solved the problem. One question: does/should the new unit test check specifically if the gradients array is created in the output trajectory file?

Yes, that is a good idea. I've updated the unit test to check whether the trajectory has velocities and gradient variables.

MBartkowiakSTFC (approved):
It works fine, and the test looks good too.

Issue #213, opened 2023-08-07
Handling of NPT trajectories


gonzalezma:
I have recently learnt (see e.g. this paper https://doi.org/10.1021/acs.jctc.3c00308 and references therein), that the handling of PBC in NPT simulations is far from trivial. Furthermore, different codes will use different wrapping/unwrapping schemes. While the effect should be null or negligible when simulating a solid or performing short simulations, this is no longer true for liquids simulated over very long time scales. The paper presents the effect on the m.s.d. and the self-diffusion coefficient, but any dynamical property would be affected.

I do not think that there is any simple solution, so for the moment I can only think of 3 possible actions:

1. Do nothing and let the user with the full responsibility of how he/she uses MDANSE.
2. Check (when converting or loading a trajectory) if we have an NPT trajectory (i.e. cell dimensions change) and if so, show a warning message preventing of the possible problems and citing this reference, so it is up to the user to decide what to do.
3. Be strict and allow MDANSE to work only with constant-cell trajectories.

A 4th solution could be to agree that the TOR algorithm in the paper is the right way to do the wrapping and implement it, but I am afraid there are at least two difficulties:
1. We need to handle many different types of trajectories (see discussion in the paper about Gromacs, NAMD and Lammps)
2. The algorithm cannot be applied on individual atoms, but on centers of masses, and then the molecules reconstructed with the right dimensions.

I don't think that this will be an easy job, so my preference would go for the 2nd one. But this is open for discussion.

oerc0122:
Duplicate of #763

PR #634, opened 2025-01-14
Save Q vectors in the output file

### LINKED ISSUE #564: [ENHANCEMENT] Save the some information about the q-vectors that were generated.
Currently, in MDANSE the q-vectors are generated randomly. This means that there is no way of reproducing results exactly when the seed was not set. It would be useful for MDANSE to output the exact q-vectors so that results can be reproduced and so that we could analyze the q-vectors that were generated to determine whether a sufficient number of them were generated and etc.

Save the q-vectors that were generated to the MDA file.

### LINKED ISSUE #721: [BUG] QVectorsConfigurator docstring.
**Description of the error**
The `MDANSE.Framework.Configurators.QVectorsConfigurator` docstring makes multiple references to being able to save vectors:

> Q vectors can be saved to a user definition and, as such, can be further reused in another MDANSE session.

> To define a new Q vectors generator, you must inherit from MDANSE.Framework.QVectors.QVectors.QVector interface.

> :param value: the Q vectors generator definition. It can be a 2-tuple, whose 1st element is the name of the Q vector generator \
> and 2nd element the parameters for this configurator **or a string that matches a Q vectors user definition.**

However, in the code:
```python
        if isinstance(value, tuple):
            ...
        else:
            self.error_status = f"Q vectors setting must be a tuple {value}"
            return
```            

Docstring should be accurate.

Will return as error if string passed.

**Suggested fix**
Fix implementation or fix docstring.

**Additional details**
N/A

### LINKED ISSUE #727: [BUG] Regression: error from instrument selection in GUI
**Description of the error**
The instrument combo box does not set the q vector generator in the GUI. The error traceback:
```
  File "MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py", line 129, in set_current_instrument
    self.action.apply_instrument()
  File "MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py", line 338, in apply_instrument
    q_vector_tuple = self._current_instrument.create_q_vector_params()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentInfo.py", line 122, in create_q_vector_params
    qvec_generator = IQVectors.create(cov_type, None)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MDANSE/Src/MDANSE/Core/SubclassFactory.py", line 176, in create
    return specific_class(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MDANSE/Src/MDANSE/Framework/QVectors/LatticeQVectors.py", line 29, in __init__
    raise QVectorsError("No configuration set for the chemical system")
MDANSE.Framework.QVectors.IQVectors.QVectorsError: 'No configuration set for the chemical system'
```

Changing the instrument in the combo box should set the resolution and q vector generator input to the values defined in the instrument data base.

The parameters do not change; an error message is output.

**Suggested fix**
N/A

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
The Q vectors used in MDANSE calculations are typically created using a generator based on a limited number of input parameters. To improve the transparency and reproducibility of the calculations, it is meaningful to save the information about the exact values of the q vectors that were used in each calculation.

closes #564 
closes #727 

**Fixes**
1. The calculation between HKL values and reciprocal space vectors has been standardised.
2. The Q vector information is now saved in the output files of each analysis using Q vectors.
3. DCSF calculation now uses average unit cell parameters to generate Q vectors. This should not change the results for NVT simulations, while it may slightly reduce the noise for the NPT simulations.

**To test**
Run a scattering analysis. Confirm that the q vectors have been written into the output files.
Use SphericalLatticeQVectors in a DCSF run on a trajectory with fixed simulation box dimensions. Check that the HKL values are all integer numbers.


ChiCheng45:
I get a job failure when running DCSF with SphericalQVectors. However, I'm wondering whether or not the user should be able to choose this qvector generator for DCSF.

Also, I think we need to decide on what our stance on NPT simulations should be. This was brought up some time ago in #213 which mentions the issue for MSD and related jobs but there will be issues with the qvector generation which we try to fix here with the averaged unit cell. I'm wondering if we should follow dynasor (https://dynasor.materialsmodeling.org/workflow.html#limitations) and only allow for constant-volume MD simulations.

gonzalezma:
I think that choosing to treat only constant-volume simulations is a valid decision, especially if we decide that we cannot handle volume changes fully correctly. Perhaps it could be useful to do a small internal survey at ISIS and the ILL and ask some of our known users if they ever analyse NPT trajectories. My guess is that allowing only for constant-V simulations is probably not a big loss.

MDANSE/Src/MDANSE/Framework/QVectors/IQVectors.py
```diff
@@ -54,3 +61,90 @@ def generate(self) -> bool:
...
+    def write_vectors_to_file(self, output_data: "OutputData"):
+        """Writes a summary of the generated vectors to the output
+        file using an OutputData class instance.
+
+        Parameters
+        ----------
+        output_data : OutputData
+            An object managing the writeout to one or many output files
+        """
+        q_values = [float(x) for x in self._configuration["q_vectors"].keys()]
```
oerc0122: Keys is redundant.
```suggestion
        q_values = [float(x) for x in self._configuration["q_vectors"]]
```

MDANSE/Src/MDANSE/Framework/QVectors/SphericalLatticeQVectors.py
```diff
@@ -43,27 +43,27 @@ def _generate(self):
...
+            -hklMax[2, 2] : hklMax[2, 2] + 1,
         ]
 
-        vects = vects.reshape(
-            3, int(2 * hklMax[0] + 1) * int(2 * hklMax[1] + 1) * int(2 * hklMax[2] + 1)
+        hkl_vects = hkl_vects.reshape(
+            3,
+            int(2 * hklMax[0, 0] + 1)
+            * int(2 * hklMax[1, 1] + 1)
+            * int(2 * hklMax[2, 2] + 1),
```
oerc0122:
```suggestion
            np.prod(2 * np.diag(hklMax) + 1, dtype=int),
```

MDANSE/Tests/UnitTests/Analysis/test_qvectors.py
```diff
@@ -24,6 +26,34 @@ def trajectory():
...
+            print(f"{qvector_generator} has no shells")
+            continue
+        else:
+            print(f"Calculating q vectors for {qvector_generator}")
+        for q in instance._configuration["shells"]["value"][:2]:
+            try:
+                original_qvectors = instance._configuration["q_vectors"][q]["q_vectors"]
+            except KeyError:
+                continue
+            print("q_vectors", original_qvectors)
```
oerc0122: Leftover `print`s?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -87,6 +87,11 @@ def __init__(self, name: str, source: "h5py.File"):
         self._axes_tag = source[name].attrs["axis"]
         self._axes = {}
         self._axes_units = {}
+        if self._axes_tag == "index":
+            for dim_number, dim_length in enumerate(self._data.shape):
+                self._axes["index" + str(dim_number)] = np.arange(dim_length)
+                self._axes_units["index" + str(dim_number)] = "N/A"
```
```suggestion
                self._axes[f"index{dim_number}"] = np.arange(dim_length)
                self._axes_units[f"index{dim_number}"] = "N/A"
```

MDANSE/Src/MDANSE/Framework/QVectors/IQVectors.py
```diff
@@ -54,3 +61,90 @@ def generate(self) -> bool:
...
+        output_data.add(
+            "vector_generator_q",
+            "LineOutputVariable",
+            q_values,
+            units="1/nm",
+        )
+        qvector_lengths = [
+            self._configuration["q_vectors"][q]["q_vectors"].shape[1] for q in q_values
```
oerc0122: Given you refer to `self._configuration["q_vectors"]` a lot, would recommend adding
```
qvec_info = self._configuration["q_vectors"]
```

MDANSE/Tests/UnitTests/Analysis/test_qvectors.py
MBartkowiakSTFC: I removed them.

MDANSE/Src/MDANSE/Framework/QVectors/IQVectors.py
MBartkowiakSTFC: I added the substitution you recommended.

MBartkowiakSTFC:
The current situation in this PR is:
1. The Q vectors are written to the output files,
2. The DCSF job produces a warning if the user picks a vector generator that is not "Lattice"-based.
3. The instrument profiles now suggest the "SphericalLattiveQVectors" as the generator.
4. The DCSF job produces a warning if the unit cell dimensions change.
5. The DCSF job takes the HKL values of the peaks and calculated the Q vectors out of them. The results are still the same for fixed unit cell, and the noise is reduced for variable cells.

My preference would be not to stop the users completely from analysing NPT trajectories, as I expect that for a very small cell size variation the results will still be usable. The only thing is that I would like the warnings for specific jobs to be more visible to the users. I opened a new issue for this.

MDANSE/Src/MDANSE/Framework/Configurators/QVectorsConfigurator.py
```diff
@@ -99,23 +98,33 @@ def configure(self, value):
...
+        self["generator"] = generator
         self.error_status = "OK"
 
     def preview_output_axis(self):
+        """Output the values of |Q| from current parameters.
+
+        Returns
+        -------
+        list[float], str
+            Values of |Q| together with their physical unit
```
oerc0122:
```suggestion
        list[float]
            Values of |Q|.
        str
            Physical unit.
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
```diff
@@ -212,9 +244,35 @@ def run_step(self, index):
...
+            cell_present = True
+            cell_fixed = True
             # loop over the trajectory time steps
             for i, frame in enumerate(self.configuration["frames"]["value"]):
-                qVectors = self.configuration["q_vectors"]["value"][shell]["q_vectors"]
+                unit_cell = traj.unit_cell(frame)
+                if unit_cell is None:
+                    cell_present = False
+                elif not np.allclose(
+                    unit_cell._unit_cell, self._average_unit_cell._unit_cell
```
oerc0122: Is there any reason why this is being accessed through the private `_unit_cell` rather than the public `direct`?

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
```diff
@@ -182,6 +192,28 @@ def initialize(self):
...
+                    all_cells,
+                    axis=0,
+                )
+            )
+            self._cell_std = UnitCell(
+                np.std(
+                    all_cells,
+                    axis=0,
+                )
+            )
```
oerc0122: Is this used? Does it make sense physically to use?

MDANSE/Src/MDANSE/Framework/QVectors/ApproximateDispersionQVectors.py
```diff
@@ -24,7 +24,7 @@
 
 
 class ApproximateDispersionQVectors(LatticeQVectors):
-    """ """
+    """Generates Q vectors along a direction."""
```
oerc0122:
```suggestion
    """Generates Q vectors along a path between two points."""
```

MDANSE/Src/MDANSE/Framework/QVectors/GridQVectors.py
```diff
@@ -75,7 +82,7 @@ def _generate(self):
         dists.sort(key=operator.itemgetter(1))
         qGroups = itertools.groupby(dists, key=operator.itemgetter(1))
         qGroups = collections.OrderedDict(
-            [(k, [item[0] for item in v]) for k, v in qGroups]
+            [(k, [item[0] for item in v]) for k, v in qGroups],
         )
```
oerc0122:
```suggestion
        qGroups = {
            k: list(map(itemgetter(0), v)) for k, v in qGroups
        }
```

ChiCheng45 (commented):
Looks good. There is a problem saving the data to csv and sometimes some issues when the q_vector data is plotted to the data view (e.g. nan values).

MBartkowiakSTFC:
Now the `vector_generator_...` datasets have extended axis tags which makes the behaviour consistent with the rest of the plotting interface. One of the axes is `vector_coordinate` and is always [0,1,2], which is not optimal since the user would have to piece together each vector from three different lines - but it does not fail when producing the .csv files.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
```diff
@@ -182,6 +192,28 @@ def initialize(self):
...
+
```
ChiCheng45: The CCF is closely related to the DCSF, so if we are going to do the unit cell averaging here, then I think we should also do it in the CCF as well.

ChiCheng45:
> Now the `vector_generator_...` datasets have extended axis tags which makes the behaviour consistent with the rest of the plotting interface. One of the axes is `vector_coordinate` and is always [0,1,2], which is not optimal since the user would have to piece together each vector from three different lines - but it does not fail when producing the .csv files.

Alright, I've tested the csv file saving, and it works great, and I think it's still relatively easy to work out what's going on. I've just left a comment for the CCF calculation which we might want to update.

MBartkowiakSTFC:
cff_Jlong_total_cuts
I extended the Current Correlation Function calculations to recalculate the Q vectors from the HKL values if the unit cell dimensions are not constant. Now the behaviour should be consistent with the Dynamic Coherent Structure Factor calculations. The screenshot above shows that the results are different now, but not by much.

ChiCheng45 (approved):
Looks good, we will need to add changes in the future to warn users about the issues with NPT trajectories in MDANSE.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
MBartkowiakSTFC: The standard deviation of the unit cell is only used in the text of the warning. My intention was to give the user an idea of which cell parameters are changing, and how much. I agree that this isn't a very useful indicator, but maybe at least it can give the user an idea of what to check (for example, if they did not expect the cell to change.)

PR #635, opened 2025-01-14
Numerical derivatives configurators now check there are enough MD frames.

### LINKED ISSUE #583: [BUG] Numerical derivatives job failure with small trajectories
**Description of the error**
Job failure when we set the numerical derivatives order to something higher than the number of time steps of the trajectory.

**Suggested fix**
Currently, the errors in the logs is

```
ts[0] = np.add.reduce(coefs[0, :] * a[:4])
~~~~~~~~~~~~^~~~~~~
ValueError: operands could not be broadcast together with shapes (4,) (2,)
```

probably need to raise a more meaningful exception message. Or we can block the job from being from the configurators.

ChiCheng45:
**Description of work**
Updated the numerical derivatives configurators now check there are enough MD frames and corresponding widgets so that they go into the invalid state when the frames and derivatives orders are not valid.

Animation62

**Fixes**
Fixes #583

**To test**
Test calculations that apply the numerical derivatives such as DOS and VACF. Using the GUI check that the configurator widgets go into an invalid state when there aren't enough frames for the selected interpolation order e.g. order 5 with the number of MD frames equal 2.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurable.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurable.py
@@ -124,17 +124,21 @@ class Configurable(object):
 
         return self._configuration
 
-    def setup(self, parameters):
-        """
-        Setup the configuration according to a set of input parameters.
+    def setup(self, parameters: dict, rebuild: bool = True):
+        """Builds and sets the configuration according to a set of
+        input parameters.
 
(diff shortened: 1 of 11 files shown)


MBartkowiakSTFC (approved):
The derivative/interpolation order widget gets blocked and unblocked as expected.

PR #636, opened 2025-01-16
Documentation update: new converged results section

ChiCheng45:
**Description of work**
General documentation cleanup and link fixes.

A new section was added that explains some points about obtaining converged results. It contains an example for obtaining clean DCSF results, basically long trajectories are required so that a larger number of configurations are used per time step of the correlation function.

Updated the current correlation function documentation so that it gives a better explanation of what is calculated.


MBartkowiakSTFC (approved):
Thanks for updating this part of the documentation. You're welcome to merge if you agree with the changes I made.

PR #645, opened 2025-01-28
Fixed discover fold setting

ChiCheng45:
**Description of work**
Previously the Discover converter always folded the trajectory whether or not the setting had been selected.

**Fixes**
The discover converter fold setting.

**To test**
Run the discover converter job and check that the fold setting works correctly.


MBartkowiakSTFC (approved):
Looks good. The fold setting now makes a difference.

PR #647, opened 2025-01-28
Update analysis and converter unit tests so they check against reference values

ChiCheng45:
**Description of work**

- Updated analysis unit tests so they check against reference values. 
- Updated converter unit tests so they check certain values against already converted files. improvedase has not been updated as I find that the trajectories they converted didn't look correct, this should be updated when improvedase is fixed.
- Updated IR unit tests so that charges are specified. 
- Updated the DCD converter so the time series starts from zero.
- Added scattering reference results by changing qvector generator to the grid one since that doesn't generate vectors randomly.
- Added PositionPowerSpectrum, CoordinationNumber, PairDistributionFunction, StaticStructureFactor and XRayStaticStructureFactor tests.

This change should reduce some of the manual testing that we'll need to do, especially when we need to test results against previous versions.

**To test**
Unit tests should pass. The CHARMM and NAMD trajectory converters should work as usual except the times should start from zero.


ChiCheng45:
If #616 gets merged before this one then we will need to change the van Hove function distinct results since #616 changes how it is calculated by adding both g_ij(r,t) and g_ji(r,t). Also the solvent-accessible surface and voronoi calculation since they were fixed or updated.

MBartkowiakSTFC (approved):
I tried changing the code of several analysis types. The changes made the unit tests fail, as expected.
Thanks for adding more analysis types to this PR.

PR #652, opened 2025-01-31
Bond tolerance update and trajectory editor widget fixes.

### LINKED ISSUE #637: [ENHANCEMENT] Change how trajectory editor determines molecules
Currently, the trajectory editor in MDANSE determines bond by the sum of covalent radii multiplied by 1 + some tolerance. 

I far as I remember the standard was the sum of covalent radii plus ~0.4 angstroms. This is what we did when I worked on a crystal structure prediction code when I was a PhD student. See https://gitlab.com/mol-cspy/mol-cspy/-/blob/main/cspy/chem/molecule.py#L101 and is also something done by openeye aswell https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEDetermineConnectivity.html. This was the method I chose to implement for the bonding calculation in the 3D viewer.

ChiCheng45:
**Description of work**
Adjusted the way that the bonds are determined, previously it was the sum of the covalent radii multiplied by (1 + tolerance).  I have changed this so that it is the sum of covalent radii plus ~0.4 angstrom. This is the standard method, I don't know where this came from I think probably from the CCDC since they did some work determining the covalent radii using the CSD. But anyway here are some places where this method is used.

https://www.ccdc.cam.ac.uk/media/superstar-1.pdf (see section 10.2.9 COVALENT_TOLERANCE)
https://pubs.acs.org/doi/10.1021/ci700028w
https://jcheminf.biomedcentral.com/articles/10.1186/1758-2946-3-33 (see section Bond Perception and Atom Typing)
https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEDetermineConnectivity.html

**Fixes**
Fixes #637

**To test**

- Run the trajectory editor with molecule finder and check that the expected molecules are found. 
- Check that the default values are filled and used after running the job when the entry boxes are empty in the unit cell and molecule finder widgets.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryEditor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryEditor.py
@@ -79,7 +79,7 @@ class TrajectoryEditor(IJob):
         "OptionalFloatConfigurator",
         {
             "default": [False, 0.04],
-            "label_text": "Search for molecules (using the following bond length tolerance)",
+            "label_text": "Search for molecules (covalent radii plus the tolerance in nm)",
         },
     )
     settings["output_files"] = (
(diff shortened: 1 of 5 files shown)


MBartkowiakSTFC (approved):
I tried it out on a trajectory which was known to have problems in the past. The molecules were found correctly. Looks good.

PR #655, opened 2025-02-03
Improve H5MD trajectory support

MBartkowiakSTFC:
**Description of work**
Initially, the H5MD support was added to MDANSE using sample H5MD trajectories produced by MDMC. In this PR changes were made to the code to ensure that trajectories generated using https://github.com/oerc0122/castep_outputs_tools can also be loaded.

**Fixes**
1. Extend the code finding physical units in H5MD files.
2. Add a check for chemical element dictionary in the `species` dataset metadata.
3. Improve the code reading unit cells and time values.

**To test**
Tests must pass.
Use the trajectory https://github.com/ISISNeutronMuon/MDANSE-Examples/tree/main/Data/Trajectories/CASTEP and convert it to H5MD using https://github.com/oerc0122/castep_outputs_tools . It should load and show correct atom positions and simulation box size.

PR #657, opened 2025-02-05
Visualise atom trace

MBartkowiakSTFC:
**Description of work**
This PR adds the option of visualising the 'atom trace' in the 3D view. It is intended to show the accumulated volume that was taken by an atom within the entire time of the simulation. This may be useful for identifying hopping paths in porous materials, ionic conductors, etc.

Screenshot 2025-02-05 at 10 31 51

**Fixes**
1. Adapted the code from Eric's molecular viewer.
2. Enabled multiple surfaces in a single view.
3. Reimplemented the trace calculation method to limit the grid size and account for the atom radius.

**To test**
1. Load a trajectory.
2. Calculate the trace for several atoms.
3. Check that the isosurface does not disappear when switching to a different frame.
4. Check that the isosurface disappears when switching to a different trajectory.


MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -52,30 +52,47 @@ def __init__(self, h5_filename):
...
         # Load the first configuration
         coords = self._h5_file["/particles/all/position/value"][0, :, :]
         try:
             pos_unit = self._h5_file["/particles/all/position/value"].attrs["unit"]
-        except:
+        except Exception:
             conv_factor = 1.0
         else:
-            if pos_unit == "Ang":
+            if pos_unit == "Ang" or pos_unit == "Angstrom":
```
```suggestion
            if pos_unit in ("Ang", "Angstrom"):
```
And elsewhere

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -258,9 +276,16 @@ def _load_unit_cells(self):
...
+                    if cell.shape == (3, 3):
+                        temp_array = np.array(cell)
+                    else:
+                        temp_array = np.array(
+                            [
+                                [cell[0], 0.0, 0.0],
+                                [0.0, cell[1], 0.0],
+                                [0.0, 0.0, cell[2]],
+                            ]
+                        )
```
```suggestion
                        temp_array = np.diag(cell)
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -258,9 +276,16 @@ def _load_unit_cells(self):
         else:
             if len(cells.shape) > 1:
```
```suggestion
            if cells.ndim > 1:
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -258,9 +276,16 @@ def _load_unit_cells(self):
...
                 for cell in cells:
-                    temp_array = np.array(
-                        [[cell[0], 0.0, 0.0], [0.0, cell[1], 0.0], [0.0, 0.0, cell[2]]]
-                    )
+                    if cell.shape == (3, 3):
+                        temp_array = np.array(cell)
+                    else:
```
oerc0122, line 280:
```suggestion
                    elif cell.shape = (3,):
```
With an `else: raise ...` otherwise risk ignoring mis-shapen cells.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -39,7 +41,7 @@
 )
 
 
-def array_to_3d_imagedata(data, spacing):
+def array_to_3d_imagedata(data: np.ndarray, spacing: Tuple[float]):
     nx = data.shape[0]
     ny = data.shape[1]
     nz = data.shape[2]
```
```suggestion
    nx, ny, nz = data.shape
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -189,36 +224,75 @@ def _new_atom_parameters(self):
...
+
+        coords = self._reader.read_atom_trajectory(index)
+        element = self._reader._atom_types[index]
+        radius = self._reader._trajectory.get_atom_property(element, "covalent_radius")
+        upper_limit = np.max(coords, axis=0) + radius
+        lower_limit = np.min(coords, axis=0) - radius
+        grid_step = radius / fine_sampling
+
+        span = upper_limit - lower_limit
+        grid_steps = [int(x) for x in span / grid_step]
```
```suggestion
        grid_steps = list(span // grid_step)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -189,36 +224,75 @@ def _new_atom_parameters(self):
...
+        gdim = (grid_steps[0], grid_steps[1], grid_steps[2])
+        grid = np.zeros(gdim, dtype=np.int32)
+
+        indices = np.floor((coords - lower_limit.reshape((1, 3))) / grid_step).astype(
+            int
         )
```
```suggestion
        indices = (coords - lower_limit.reshape((1, 3))) // grid_step).astype(int)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -189,36 +224,75 @@ def _new_atom_parameters(self):
...
+        coords = self._reader.read_atom_trajectory(index)
+        element = self._reader._atom_types[index]
+        radius = self._reader._trajectory.get_atom_property(element, "covalent_radius")
+        upper_limit = np.max(coords, axis=0) + radius
+        lower_limit = np.min(coords, axis=0) - radius
+        grid_step = radius / fine_sampling
+
+        span = upper_limit - lower_limit
+        grid_steps = [int(x) for x in span / grid_step]
+        gdim = (grid_steps[0], grid_steps[1], grid_steps[2])
```
```suggestion
        gdim = tuple(grid_steps[0:3])
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -488,38 +565,40 @@ def get_render_window(self):
     def iren(self):
         return self._iren
 
-    def on_change_atomic_trace_opacity(self, opacity):
+    def on_change_atomic_trace_opacity(self, surface_number, opacity):
```
oerc0122: What is `surface_number`? `surface_index` or `n_surfaces`?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -488,38 +565,40 @@ def get_render_window(self):
...
         """Event handler called when the opacity level is changed."""
 
-        if self._surface is None:
+        if len(self._surfaces) <= surface_number:
```
oerc0122: Because we want to check it's out of range it makes more sense to me this way round. However `len(self._surfaces)` is 1 more than the largest index, so are you sure you wanted `<=`?
```suggestion
        if surface_number > len(self._surfaces):
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
```diff
@@ -0,0 +1,158 @@
...
+        self.update_limits()
+        self._fraction_spinbox.setMaximum(100)
+        self._fraction_spinbox.setValue(5)
+        self._grid_spinbox.setMaximum(10)
+        self._grid_spinbox.setMinimum(1)
+        self._grid_spinbox.setValue(3)
+        self._opacity_spinbox.setMaximum(1.0)
+        self._opacity_spinbox.setValue(0.5)
+        self._opacity_spinbox.setSingleStep(0.01)
+        for pair in [
```
oerc0122:
```suggestion
        for label, widget in [
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
```diff
@@ -0,0 +1,158 @@
...
+        if self._molviewer is None:
+            return
+        if len(self._molviewer._surfaces) == 0:
+            self.remove_trace_button.setEnabled(False)
+        else:
+            self.remove_trace_button.setEnabled(True)
+        if self._molviewer._n_atoms < 1:
+            self.add_trace_button.setEnabled(False)
+        else:
+            self.add_trace_button.setEnabled(True)
```
oerc0122:
```suggestion
        self.remove_trace_button.setEnabled(len(self._molviewer._surfaces) != 0)
        self.add_trace_button.setEnabled(self._molviewer._n_atoms > 0)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
```diff
@@ -0,0 +1,158 @@
...
+        params["atom_number"] = self._atom_spinbox.value()
+        params["surface_number"] = self._surface_spinbox.value()
+        try:
+            params["surface_colour"] = [
+                float(x) / 256 for x in self._colour_lineedit.text().split(",")
+            ]
+        except ValueError:
+            pass
+        except TypeError:
+            pass
```
oerc0122:
```Python
from contextlib import suppress
...
```

```suggestion
        with suppress(ValueError, TypeError):
            params["surface_colour"] = [
                float(x) / 256 for x in self._colour_lineedit.text().split(",")
            ]
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: All occurrences have been replaced with surface_index.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: I swapped the variables around.

An example: an array of length 3 has valid indices 0, 1, 2. Indices 3, 4, ... are out of range. This will be captured by the condition `index >= length`.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
MBartkowiakSTFC: Both occurrences replaced.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
MBartkowiakSTFC: I added the import and the change.

MBartkowiakSTFC:
There are two features missing in this PR:
1. There is no way to indicate to the user in the 3D view which atom the trace will be calculated for.
2. The colour input field has no validation.
This will be summarised in an issue and added in a later PR.

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -115,7 +134,7 @@ def __getitem__(self, frame):
         except:
             conv_factor = 1.0
         else:
-            if pos_unit == "Ang":
+            if pos_unit == "Ang" or pos_unit == "Angstrom":
```
```suggestion
            if pos_unit in ("Ang", "Angstrom"):
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -193,7 +212,7 @@ def coordinates(self, frame):
...
```
```suggestion
            if pos_unit in ("Ang", "Angstrom"):
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -243,22 +262,27 @@ def _load_unit_cells(self):
         self._unit_cells = []
         try:
             box_unit = self._h5_file["/particles/all/box/edges/value"].attrs["unit"]
-        except:
-            conv_factor = 1.0
+        except (AttributeError, KeyError):
+            conv_factor = 0.1
         else:
-            if box_unit == "Ang":
+            if box_unit == "Ang" or box_unit == "Angstrom":
```
```suggestion
            if pos_unit in ("Ang", "Angstrom"):
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -243,22 +262,27 @@ def _load_unit_cells(self):
...
-            if len(cells.shape) > 1:
+            if cells.ndim > 1:
                 for cell in cells:
-                    temp_array = np.array(
-                        [[cell[0], 0.0, 0.0], [0.0, cell[1], 0.0], [0.0, 0.0, cell[2]]]
-                    )
+                    if cell.shape == (3, 3):
+                        temp_array = np.array(cell)
+                    elif cell.shape == (3,):
+                        temp_array = np.diag(cell)
```
oerc0122, line 279: **N.B.** L. 288-289, can also use L. 280-281.

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -70,10 +88,10 @@ def __init__(self, h5_filename: Union[Path, str]):
         coords = self._h5_file["/particles/all/position/value"][0, :, :]
         try:
             pos_unit = self._h5_file["/particles/all/position/value"].attrs["unit"]
-        except:
+        except Exception:
```
```suggestion
        except KeyError:
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -52,16 +52,34 @@ def __init__(self, h5_filename: Union[Path, str]):
...
-                byte.decode() for byte in self._h5_file["/parameters/atom_symbols"]
+                reverse_lookup[type_number] for type_number in particle_types
             ]
-        except KeyError:
-            chemical_elements = self._h5_file["/particles/all/species"]
         self._chemical_system = ChemicalSystem(self._h5_filename.stem)
-        self._chemical_system.initialise_atoms(chemical_elements)
+        try:
+            self._chemical_system.initialise_atoms(chemical_elements)
+        except Exception:
```
oerc0122: More specific exception from `initialise_atoms`?

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -270,14 +294,17 @@ def _load_unit_cells(self):
     def time(self):
         try:
             time_unit = self._h5_file["/particles/all/position/time"].attrs["unit"]
-        except:
+        except KeyError:
```
oerc0122: Do you need `AttributeError` here too as you do below? I think they always have `attrs` though.

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -369,7 +396,7 @@ def read_com_trajectory(
         except:
             conv_factor = 1.0
         else:
-            if pos_unit == "Ang":
+            if pos_unit == "Ang" or pos_unit == "Angstrom":
```
```suggestion
            if pos_unit in ("Ang", "Angstrom"):
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -467,7 +494,7 @@ def read_atomic_trajectory(
...
```
```suggestion
            if pos_unit in ("Ang", "Angstrom"):
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
```diff
@@ -264,7 +267,22 @@ def createSidePanel(self):
         layout.addWidget(wrapper5)
         # the database of atom types
         # self._database = TrajectoryAtomData()
-        self.layout().addWidget(base, 0, 2, 2, 1)  # row, column, rowSpan, columnSpan
+        self.layout().addWidget(
+            absolute_base, 0, 2, 2, 1
```
oerc0122, line 270:
```suggestion
            absolute_base, 
            row=0, 
            column=2, 
            rowSpan=2, 
            columnSpan=1,
```
Then no need for comment.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -39,10 +41,8 @@
...
 
-def array_to_3d_imagedata(data, spacing):
-    nx = data.shape[0]
-    ny = data.shape[1]
-    nz = data.shape[2]
+def array_to_3d_imagedata(data: np.ndarray, spacing: Tuple[float]):
+    nx, ny, nz = data.shape
     image = vtk.vtkImageData()
     image.SetDimensions(nx, ny, nz)
     dx, dy, dz = spacing
```
oerc0122, line 48: I realise it's not really part of this PR, but for later reference, l. 57-60 could be:
```python
    for (i, j, k), val in np.ndenumerate(data):
        image.SetScalarComponentFromDouble(i, j, k, 0, val)
```
or even
```python
    for loc, val in np.ndenumerate(data):
        image.SetScalarComponentFromDouble(*loc, 0, val)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -230,27 +302,28 @@ def _draw_isosurface(self, index):
...
         mapper.Update()
 
-        self._surface = vtk.vtkActor()
-        self._surface.SetMapper(mapper)
-        self._surface.GetProperty().SetColor((0, 0.5, 0.75))
-        self._surface.GetProperty().SetOpacity(0.5)
-        self._surface.PickableOff()
+        new_surface = vtk.vtkActor()
+        new_surface.SetMapper(mapper)
+        new_surface.GetProperty().SetColor((r, g, b))
```
oerc0122: Is there a reason to unpack the tuple only to repack it here?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
```diff
@@ -0,0 +1,151 @@
...
+            temp_layout.addWidget(widget)
+            layout.addWidget(temp_box)
+        layout.addWidget(self.add_trace_button)
+        for label, widget in [
+            ("Remove the surface with index: ", self._surface_spinbox)
+        ]:
+            temp_box = QGroupBox(label, self)
+            temp_layout = QVBoxLayout(temp_box)
+            temp_layout.addWidget(widget)
+            layout.addWidget(temp_box)
```
oerc0122, line 186: Any reason for this to be split off?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
```diff
@@ -0,0 +1,151 @@
...
+        if self._molviewer is None:
+            return
+        self.remove_trace_button.setEnabled(len(self._molviewer._surfaces) != 0)
+        self.add_trace_button.setEnabled(self._molviewer._n_atoms > 0)
+
+    def get_values(self):
+        params = {
+            "atom_number": 0,
+            "fine_sampling": 3,
+            "surface_colour": (0, 0.5, 0.75),
```
oerc0122: Does this value want to be a top-level constant somewherE?

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
MBartkowiakSTFC: These changes are part of PR #655

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
MBartkowiakSTFC: The layout is populated sequentially. I needed to add a button, which is not wrapped in a QGroupBox, before this input field, but the logic of adding the input fields was the same before and after the button.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
MBartkowiakSTFC: On my platform, the modified code triggers:
```
TypeError: arguments did not match any overloaded call:
  addWidget(self, w: Optional[QWidget]): 'row' is not a valid keyword argument
  addWidget(self, a0: Optional[QWidget], row: int, column: int, alignment: Qt.AlignmentFlag = Qt.Alignment()): 'row' is not a valid keyword argument
  addWidget(self, a0: Optional[QWidget], row: int, column: int, rowSpan: int, columnSpan: int, alignment: Qt.AlignmentFlag = Qt.Alignment()): 'row' is not a valid keyword argument
```

oerc0122: Huh. Maybe it's positional only, guess that's part of the C interface. Nevermind then.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/TraceWidget.py
MBartkowiakSTFC: Now this dictionary is a top-level constant and is used both in TraceWidget.py and MolecularViewer.py.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: I added input validation in the widget to make sure the user can only input a tuple of 3 numbers, and removed unpacking from this part of the code.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: Change added.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: Change added with an extra `(` before coords.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: Change added.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
MBartkowiakSTFC: OK, the code now uses `np.ndenumerate`, the former implementation.

MBartkowiakSTFC:
I think that I addressed the points dealing with the 3D view and isosurface calculation. Some other comments about H5MD were covered by #655 

Also, I tried to remove unused code from the 3D viewer and add/improve docstrings.

Do you think that these changes could be merged?

ChiCheng45:
I think this is fine to go in. A few points from me.

- With the coarse option, there are holes in the 3D mesh.

image

- I've been thinking whether you could instead create a 3d mesh using Delaunay3D from vtk (see https://examples.vtk.org/site/Cxx/Modelling/Delaunay3D/ and https://vtk.org/files/examples/Delaunay3D/index.html). However, I think it will probably have the same hole issue.

- Currently the atom selection is based on index. I think users will want to select all waters molecules and etc.

ChiCheng45 (approved):

ChiCheng45:
Looking good. I made this nice trace of a water molecule and colored the oxygen and hydrogens differently. You can see the rotations of the water molecule as it moves.

image

image

The PR is good to go in. I think the next thing would be to make those surfaces smoothed out somehow.

PR #662, opened 2025-02-10
Add ruff format to gui

oerc0122:
**Description of work**
- Add ruff formatting instead of black.
- Run ruff format on changed files.

**Fixes**
Formatting only.

**To test**
Should be no changes, tests should pass.


MBartkowiakSTFC (approved):
The changes did not have any side effects, as far as I can tell. Looks good.

PR #663, opened 2025-02-11
Simplify PDB reader and add docs

oerc0122:
**Description of work**
- Clean up `MinimalPDBReader`. 
- Turn `MinimalPDBReader.find_atoms` into a generator.

**Fixes**
General cleanup of `MinimalPDBReader`

**To test**
Should be no functional changes.

### DIFF
--- a/MDANSE/Src/MDANSE/IO/MinimalPDBReader.py
+++ b/MDANSE/Src/MDANSE/IO/MinimalPDBReader.py
@@ -14,56 +14,50 @@
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+from collections.abc import Generator, Iterable
 from typing import List
 
 import numpy as np
 from ase.io import read as ase_read
-
-from MDANSE.MLogging import LOG
-from MDANSE.Framework.Units import measure
 from MDANSE.Chemistry import ATOMS_DATABASE
 from MDANSE.Chemistry.ChemicalSystem import ChemicalSystem
-from MDANSE.MolecularDynamics.Configuration import (
-    RealConfiguration,
-    PeriodicRealConfiguration,
-)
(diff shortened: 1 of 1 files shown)


MDANSE/Src/MDANSE/IO/MinimalPDBReader.py
```diff
@@ -74,113 +67,123 @@ def __init__(self, filename: str):
...
-                element_list.append(backup_element2)
-            elif backup_element3 in ATOMS_DATABASE.atoms:
-                element_list.append(backup_element3)
-            elif chemical_element in ATOMS_DATABASE.atoms:
-                element_list.append(chemical_element)
-            elif processed_atom_name in ATOMS_DATABASE.atoms:
-                element_list.append(processed_atom_name)
+                backup_elements.append(atom_name[0])
+
+            for trial in (chemical_element, processed_atom_name, *backup_elements):
```
MBartkowiakSTFC: Somewhere around here (but possibly a few lines before) the logic of the code got changed, and if I convert a Gromacs trajectory of a protein using this branch, it ends up containing helium and mercury where it was supposed to be hydrogen.

Should we add a test for it? I will share some example files with you.

oerc0122: I changed the order of `backup_elements` to go at the end (as one might think they ought) but I can stick them back at the front again. Try this:
```suggestion
            for trial in (*backup_elements, chemical_element, processed_atom_name):
```

MBartkowiakSTFC (approved):
After this one change it is working the same as the main branch.

oerc0122:
> After this one change it is working the same as the main branch.

Will add the test to ensure it keeps working before merging though

oerc0122:
Gromacs test added in #671

PR #668, opened 2025-02-11
Update the CP2K configurator and widget so that the velocity file is optional

### LINKED ISSUE #640: [BUG] CP2K velocity file widget is not optional
**Description of the error**
The CP2K converter execution is blocked unless the user inputs a velocity file, even though it should be enough to input the positions and unit cell files.

The velocity file should be optional.

Unless a valid file name has been specified, the GUI makes it impossible to start the converter.

**Suggested fix**
Optional input file widgets already exist in other converters, such as the MDTraj converter. A similar approach should be implemented.

**Additional details**
In principle, we should also have a slot for an optional input file containing forces per atom, since our trajectories can store these too.

ChiCheng45:
**Description of work**
Added an optional XYZ configurator and updated the cp2k converter so velocity files are now optional.

**Fixes**
Fixes #640

**To test**
Run the cp2k conversion with and without a velocity file. Both should convert without issues. Check that velocities have been saved into the MDA file when a velocity file has been used.


ChiCheng45:
@MBartkowiakSTFC In #640 you suggested adding forces to the CP2K converter, I think we can do this easily, we can add this feature to the PR if you think it's a good idea. Do you have force XYZ files already?

MBartkowiakSTFC:
> @MBartkowiakSTFC In #640 you suggested adding forces to the CP2K converter, I think we can do this easily, we can add this feature to the PR if you think it's a good idea. Do you have force XYZ files already?

Please go ahead and add forces input to the CP2K converter. I will share a suitable trajectory with you.

MBartkowiakSTFC:
From the point of view of the GUI, everything works fine.

Do we think we need to allow the case where we have positions and forces, but no velocities? I think our converter always assumes that if forces are present, then velocities must be there too. At the moment a run with positions+forces fails.

ChiCheng45:
> From the point of view of the GUI, everything works fine.
> 
> Do we think we need to allow the case where we have positions and forces, but no velocities? I think our converter always assumes that if forces are present, then velocities must be there too. At the moment a run with positions+forces fails.

This should work now so that you can use positions and forces only. I've added some tests for this use case. There was a typo with the force file closing.

MBartkowiakSTFC (approved):
All combinations work for me now. Looks good.

PR #671, opened 2025-02-14
Simplify test converter add gromacs test

oerc0122:
**Description of work**
- Refactor `test_converter` to reduce code duplication through `pytest.mark.parametrize`. 
- Replace `os.path` with `pathlib.Path`
- Add `gromacs-nvt` test
Instead implementing `_converter_test` to handle standard format testing.

**Fixes**
N/A

**To test**
Tests should pass.


MBartkowiakSTFC:
One big advantage of the new version is that it is much shorter, and this for me counts already as 'easier to read'.

Would you like to replace the occurrences of the word `typ` with something clearer to a reader who expects it to be written in English?

MBartkowiakSTFC (approved):
I think this can be merged now. Looks good.

Issue #286, opened 2024-01-25
[ENHANCEMENT] Add bonding and atom information to MDANSE


ChiCheng45:
**Is your feature request related to a problem? Please describe.**
At the moment there is no bonding and atom information in MDANSE. It might be useful to be able to distinguish between different bonding types (single, double, triple, ...) and atoms (aromatic, aliphatic, ...)in the future. This might be important if we want to allow the user to make more complex atom selections.

**Describe the solution you'd like**
The way bonding in MDANSE needs to be refactored and a bonding type needs to be added to the chemical entities.

ChiCheng45:
Bonding types could be determined with rdkit see #350.

oerc0122:
Is this done now?

ChiCheng45:
> Is this done now?

No, this isn't finished yet because all bond types in MDANSE are unspecified. The SMARTS pattern matching doesn't work with some patterns that specify bond and atom types, etc.

PR #672, opened 2025-02-14
New atom selection mechanism

MBartkowiakSTFC:
**Description of work**
The intention behind this PR is to implement a mechanism of selecting atoms which is:
1. Not attached to a specific trajectory file,
2. Transferable,
3. Easy to extend.

We can still discuss if the implementation in this PR meets the requirements.

This screenshot shows the carbon dioxide trajectory in which two out of three atoms have been selected in the molecules from the first half of the index range. This is not scientifically meaningful, but illustrates how the selection operations can be combined.
selection_helper

**Fixes**
1. New selection functions have been implemented to replace the classes we have been using so far.
2. The output of each selection function is a set of indices.
3. Selections are applied _sequentially_, and can be combined using set operations (union, intersection, difference).
4. ReusableSelection class stores a sequence of selection functions and uses JSON for saving and loading a sequence.
5. ReusableSelection.select_in_trajectory method takes a Trajectory instance as input, applies the selection and returns the resulting set of atom indices.
6. Analysis jobs save the JSON selection sequence as one of the job inputs, and an array of 0/1 values to mark the indices which got selected.

The JSON representation is saved as str:
saved_input
The indices are saved as a mask array:
saved_selection_array

**To test**
All tests must pass.


ChiCheng45:
Looks good. Just one thing about the SMARTS pattern matching. In the RDKit molecule object we use the Chem.rdchem.BondType.UNSPECIFIED for all of the bondtypes. This means that most SMARTS patterns you'd think should work won't and that is why the SMARTS patterns we use are a bit unusual.

The correct thing would be to set all the correct bond types and atom properties but I think this might get difficult to get working for a general molecule e.g. setting single/double/aromatic bonds for a benzene or cyclooctatetraene or some big protein.

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+        Set of all the atom indices
+    """
+    selection = set()
+    system = trajectory.chemical_system
+    element_list = system.atom_list
+    name_list = system.name_list
+    indices = set(range(len(element_list)))
+    index_list = function_parameters.get("index_list", None)
+    index_range = function_parameters.get("index_range", None)
+    index_slice = function_parameters.get("index_slice", None)
```
oerc0122: `None` is default, but happy to leave it for clarity. In fact, given you're subsequently doing intersections on these, may want to return empty `set`s.
```suggestion
    index_list = function_parameters.get("index_list")
    index_range = function_parameters.get("index_range")
    index_slice = function_parameters.get("index_slice")
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+    if index_list is not None:
+        selection = selection.union(indices.intersection(index_list))
```
oerc0122: If everything's a `set`.
```suggestion
        selection |= indices & index_list
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+    if index_range is not None:
+        selection = selection.union(
+            indices.intersection(range(index_range[0], index_range[1]))
+        )
```
oerc0122:
```suggestion
        selection |= indices & set(range(*index_range))
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+    if index_slice is not None:
+        selection = selection.union(
+            indices.intersection(range(index_slice[0], index_slice[1], index_slice[2]))
+        )
```
oerc0122:
```suggestion
        selection |= indices & set(range(*index_slice))
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+    atom_types = function_parameters.get("atom_types", None)
+    if atom_types:
+        new_indices = [index for index in indices if element_list[index] in atom_types]
+        selection = selection.union(new_indices)
```
oerc0122:
```suggestion
        new_indices = {index for index in indices if element_list[index] in atom_types}
        selection |= new_indices
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+        selection = selection.union(indices.intersection(index_list))
+    if index_range is not None:
+        selection = selection.union(
+            indices.intersection(range(index_range[0], index_range[1]))
+        )
+    if index_slice is not None:
+        selection = selection.union(
+            indices.intersection(range(index_slice[0], index_slice[1], index_slice[2]))
+        )
+    atom_types = function_parameters.get("atom_types", None)
```
oerc0122: Because you're checking `in`, `None` will raise:
```python
>>> 1 in None
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'NoneType' is not iterable
```

```suggestion
    atom_types = function_parameters.get("atom_types", ())
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+    if atom_types:
+        new_indices = [index for index in indices if element_list[index] in atom_types]
+        selection = selection.union(new_indices)
+    atom_names = function_parameters.get("atom_names", None)
```
oerc0122: As above
```suggestion
    atom_names = function_parameters.get("atom_names", ())
```

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,66 @@
...
+    if atom_names:
+        new_indices = [index for index in indices if name_list[index] in atom_names]
+        selection = selection.union(new_indices)
```
oerc0122:
```suggestion
        new_indices = {index for index in indices if name_list[index] in atom_names}
        selection |= new_indices
```

oerc0122: Could also merge the above two checks.
```python
atom_names = set(function_parameters.get("atom_names", ()))
atom_types = set(function_parameters.get("atom_types", ()))
new_indices = {index for index in indices
               if name_list[index] in atom_names or 
               element_list[index] in atom_types}
selection |= new_indices
```

MDANSE/Src/MDANSE/Framework/AtomSelector/general_selection.py
```diff
@@ -0,0 +1,65 @@
...
+    -------
+    Set[int]
+        An empty set.
+    """
+    return set()
+
+
+def invert_selection(trajectory: Trajectory, selection: Set[int]) -> Set[int]:
+    all_indices = select_all(trajectory)
+    inverted = all_indices.difference(selection)
```
oerc0122:
```suggestion
    inverted = all_indices - selection
```

MDANSE/Src/MDANSE/Framework/AtomSelector/group_selection.py
```diff
@@ -0,0 +1,72 @@
...
+from functools import reduce
+
+from MDANSE.Chemistry.ChemicalSystem import ChemicalSystem
+from MDANSE.MolecularDynamics.Trajectory import Trajectory
+
+
+def select_labels(
+    trajectory: Trajectory, **function_parameters: Dict[str, Any]
+) -> Set[int]:
+    """Selects all the atoms in the trajectory.
```
oerc0122: Not sure this docstring is right.

MDANSE/Src/MDANSE/Framework/AtomSelector/group_selection.py
```diff
@@ -0,0 +1,72 @@
...
+    -------
+    Set[int]
+        Set of all the atom indices
+    """
+    selection = set()
+    system = trajectory.chemical_system
+    atom_labels = function_parameters.get("atom_labels", None)
+    for label in atom_labels:
+        if label in system._labels:
+            selection = selection.union(system._labels[label])
```
oerc0122:
```suggestion
    system = trajectory.chemical_system
    atom_labels = function_parameters.get("atom_labels", ())
    selection = {system._labels[label] for label in atom_labels if label in system._labels}
```
Unless `system._labels` is a `dict[str, Sequence[T]]`.

MDANSE/Src/MDANSE/Framework/AtomSelector/group_selection.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Returns
+    -------
+    Set[int]
+        Set of all the atom indices
+    """
+    selection = set()
+    system = trajectory.chemical_system
+    pattern = function_parameters.get("rdkit_pattern", None)
+    if pattern:
```
oerc0122:
```suggestion
    if pattern := function_parameters.get("rdkit_pattern"):
```

MDANSE/Src/MDANSE/Framework/AtomSelector/molecule_selection.py
```diff
@@ -13,35 +13,37 @@
...
-    pattern = "[#8X2;H2](~[H])~[H]"
-    if check_exists:
-        return system.has_substructure_match(pattern)
-    else:
-        return system.get_substructure_matches(pattern)
+    molecule_names = function_parameters.get("molecule_names", None)
+    for molecule in molecule_names:
+        if molecule in system._clusters:
+            selection = selection.union(
+                reduce(list.__add__, system._clusters[molecule])
```
oerc0122: Any `reduce(?.__add__, ...)` is `sum`

MDANSE/Src/MDANSE/Framework/AtomSelector/molecule_selection.py
```diff
@@ -13,35 +13,37 @@
...
+        Set of all the atom indices
     """
+    selection = set()
     system = trajectory.chemical_system
-    pattern = "[#8X2;H2](~[H])~[H]"
-    if check_exists:
-        return system.has_substructure_match(pattern)
-    else:
-        return system.get_substructure_matches(pattern)
+    molecule_names = function_parameters.get("molecule_names", None)
```
oerc0122: Can't iterate over `None`

MDANSE/Src/MDANSE/Framework/AtomSelector/molecule_selection.py
```diff
@@ -13,35 +13,37 @@
...
+    for molecule in molecule_names:
+        if molecule in system._clusters:
+            selection = selection.union(
+                reduce(list.__add__, system._clusters[molecule])
+            )
```
oerc0122:
```suggestion
    selection = {cluster for cluster in system._clusters.get(molecule, ()) for molecule in molecule_names}
```

MDANSE/Src/MDANSE/Framework/AtomSelector/molecule_selection.py
```diff
@@ -13,35 +13,37 @@
...
 
 
-def select_water(
-    trajectory: Trajectory, check_exists: bool = False
-) -> Union[set[int], bool]:
-    """Selects the O and H atoms of all water molecules.
+def select_molecules(
+    trajectory: Trajectory, **function_parameters: Dict[str, Any]
+) -> Set[int]:
+    """Selects all the atoms in the trajectory.
```
oerc0122: Again, think these docstrings are copy-paste

MDANSE/Src/MDANSE/Framework/Configurators/AtomSelectionConfigurator.py
```diff
@@ -88,12 +95,8 @@ def get_natoms(self) -> dict[str, int]:
...
             A dictionary of the number of atom per element.
         """
-        nAtomsPerElement = {}
-        for v in self["names"]:
-            if v in nAtomsPerElement:
-                nAtomsPerElement[v] += 1
-            else:
-                nAtomsPerElement[v] = 1
+        names, counts = np.unique(self["names"], return_counts=True)
+        nAtomsPerElement = {names[n]: counts[n] for n in range(len(names))}
```
oerc0122: See: [`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter)

```
nAtomsPerElement = collections.Counter(self["names"])

File: MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py [hunk 18]
```diff
@@ -34,31 +34,30 @@ def __init__(self, trajectory: Trajectory) -> None:
...
         """
         if symbol not in ATOMS_DATABASE:
             raise ValueError(f"{symbol} not found in the atom database.")
 
-        self.selector.update_settings(selection_dict, reset_first=True)
-        for idx in self.selector.get_idxs():
+        self.selector.load_from_json(selection_string)
+        indices = self.selector.select_in_trajectory(self._current_trajectory)
+        for idx in indices:
             self._new_map[idx] = symbol
```
oerc0122: This will always just append to `_new_map`, is this desired or does `_new_map` need to be cleared?

oerc0122:
```python
    self._new_map.update(dict.fromkeys(indices, symbol))
```

MDANSE/Src/MDANSE/Framework/Configurators/PartialChargeConfigurator.py
```diff
@@ -58,8 +56,10 @@ def update_charges(
         charge: float
             The partial charge to map the selected atoms to.
         """
-        self.selector.update_settings(selection_dict, reset_first=True)
-        for idx in self.selector.get_idxs():
+        selector = ReusableSelection()
+        selector.load_from_json(selection_string)
+        indices = selector.select_in_trajectory(self._current_trajectory)
+        for idx in indices:
             self._new_map[idx] = charge
```
oerc0122, line 63: As above

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -127,7 +127,6 @@ def __getitem__(self, frame):
         :return: the configuration
         :rtype: dict of ndarray
         """
-
```
oerc0122: Should this be in this PR?

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -169,6 +169,22 @@ def initialize(self):
...
+                LOG.warning(
+                    "Job could not find total number of atoms in atom selection."
+                )
+            else:
+                valid_indices = self.configuration["atom_selection"]["flatten_indices"]
+                self._outputData.add(
+                    "selected_atoms",
+                    "LineOutputVariable",
+                    [index in valid_indices for index in range(array_length)],
+                )
```
```suggestion
        if selection := self.configuration.get("atom_selection"):
            try:
                array_length = selection["total_number_of_atoms"]
            except KeyError:
                LOG.warning(
                    "Job could not find total number of atoms in atom selection."
                )
            else:
                valid_indices = selection["flatten_indices"]
                self._outputData.add(
                    "selected_atoms",
                    "LineOutputVariable",
                    [index in valid_indices for index in range(array_length)],
                )
```

MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py
```diff
@@ -34,31 +34,30 @@ def __init__(self, trajectory: Trajectory) -> None:
...
         self._original_map = {}
         for number, element in enumerate(trajectory.chemical_system.atom_list):
             self._original_map[number] = element
         self._new_map = {}
+        self._current_trajectory = trajectory
 
-    def apply_transmutation(
-        self, selection_dict: dict[str, Union[bool, dict]], symbol: str
-    ) -> None:
+    def apply_transmutation(self, selection_string: str, symbol: str) -> None:
```
oerc0122, line 44: I realise this is existing API, but I'm not sure what "transmutation" is being applied here? It doesn't seem to be changing anything other than the selection? The docstring isn't helping me either.

MDANSE/Tests/UnitTests/test_reusable_selection.py
```diff
@@ -0,0 +1,204 @@
...
+    reusable_selection.set_selection(None, {'function_name': 'select_atoms', 'atom_types': ['Sb']})
+    sb_selection = reusable_selection.select_in_trajectory(traj_object.trajectory)
+    assert len(sb_selection) == 64
+    reusable_selection = ReusableSelection()
+    reusable_selection.set_selection(None, {'function_name': 'select_atoms', 'atom_types': ['S', 'Sb', 'Cu']})
+    json_string = reusable_selection.convert_to_json()
+    another_selection = ReusableSelection()
+    another_selection.load_from_json(json_string)
+    cusbs_selection = another_selection.select_in_trajectory(traj_object.trajectory)
+    assert len(cusbs_selection) == 480
```
oerc0122: Would consider making these tests fixture/parametrize combos as best as possible for all the tests here.

E.g. for this one (picked because that's where my eyes were):
```python
@pytest.fixture
def trajectory(request):
   return HDFTrajectoryInputData(request.param)

@pytest.mark.parametrize("trajectory", [short_traj], indirect=True)
@pytest.mark.parametrize("element, expected", (
    (["Cu"], 208),
    (["S"], 208),
    (["Sb"], 64),
    (["S", "Sb", "Cu"], 480)
))
def test_select_atoms_selects_by_element(trajectory, element, expected):
    reusable_selection = ReusableSelection()
    reusable_selection.set_selection(None, {"function_name": "select_atoms",
                                           "atom_types": element}
    selection = reusable_selection.select_in_trajectory(short_traj.trajectory)
    assert len(selection) == expected
```
N.B. if the `fixture` creation takes a while, it may be best to create 3 `"module"` scoped fixtures for each trajectory and pass those in directly as needed or through `parametrize`.

MDANSE/Tests/UnitTests/test_selection.py
```diff
@@ -0,0 +1,109 @@
...
+    traj_object = HDFTrajectoryInputData(traj_2vb1)
+    protein_selecton = select_molecules(traj_object.trajectory, molecule_names = ['C613 H959 N193 O185 S10'])
+    assert len(protein_selecton) == 613 + 959 + 193 + 185 + 10
+
+
+def test_select_molecules_inverted_selects_ions():
+    traj_object = HDFTrajectoryInputData(traj_2vb1)
+    all_molecules_selection = select_molecules(traj_object.trajectory, molecule_names = ['C613 H959 N193 O185 S10', 'H2 O1'])
+    non_molecules_selection = invert_selection(traj_object.trajectory, all_molecules_selection)
+    assert all([traj_object.chemical_system.atom_list[index] in ['Na', 'Cl'] for index in non_molecules_selection])
```
oerc0122:
```suggestion
    assert all(traj_object.chemical_system.atom_list[index] in ['Na', 'Cl'] for index in non_molecules_selection)
```
`all` handles iterators

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -13,25 +13,99 @@
...
+    IndexSelection,
+    MoleculeSelection,
+    PatternSelection,
+    LabelSelection,
+)
+
+
+VALID_SELECTION = "Valid selection"
+USELESS_SELECTION = "Selection did not change. This operation is not needed."
+MALFORMED_SELECTION = "This is not a valid JSON string."
```
oerc0122: Consider making these an `Enum`

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -13,25 +13,99 @@
...
+            else:
+                return USELESS_SELECTION
+
+    def current_selection(self, last_operation: str = "") -> Set[int]:
+        self.rebuild_selection(last_operation)
+        return self._selection.select_in_trajectory(self._trajectory)
+
+    def current_steps(self) -> str:
+        result = {}
+        for row in range(self.rowCount()):
```
oerc0122, line 140: Can you not iterate over the children directly?

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -89,10 +139,13 @@ def __init__(
...
-        self.selector = selector
+        self.trajectory = traj_data[1].trajectory
+        self.system = self.trajectory.chemical_system
+        self.selection_model = SelectionModel(self.trajectory)
         self._field = field
-        self.settings = self.selector.settings
-        self.atm_full_names = self.selector.system.name_list
+        self.atm_full_names = self.system.name_list
+        self.molecule_names = [str(x) for x in self.system._clusters.keys()]
+        self.labels = [str(x) for x in self.system._labels.keys()]
```
oerc0122: Could be:
```suggestion
        self.molecule_names = list(map(str, self.system._clusters))
        self.labels = list(map(str, self.system._labels))
```

oerc0122: I will note you keep externally accessing these variables which are marked `_private`, suggest making them public.

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -0,0 +1,274 @@
...
+        layout.addWidget(self.commit_button)
+        self.commit_button.clicked.connect(self.create_selection)
+
+    def get_mode(self) -> str:
+        if self.mode_box.currentIndex() == 2:
+            return "difference"
+        elif self.mode_box.currentIndex() == 1:
+            return "intersection"
+        else:
+            return "union"
```
oerc0122: Might be better as an `Enum`

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -0,0 +1,274 @@
...
+        self.selection_type_combo.currentTextChanged.connect(self.switch_mode)
+
+    @Slot(str)
+    def switch_mode(self, new_mode: str):
+        self.selection_field.setText("")
+        if new_mode == "list":
+            self.selection_field.setPlaceholderText("0,1,2")
+            self.selection_keyword = "index_list"
+            self.selection_separator = ","
+        if new_mode == "range":
```
oerc0122: `elif`?

oerc0122: Also probably `Enum`

MDANSE/Src/MDANSE/Framework/AtomSelector/molecule_selection.py
MBartkowiakSTFC: I probably misunderstood how to replace this. My simple test shows:
```
In [4]: b = [[1,2,3], [4,6], [9,11]]

In [5]: reduce(list.__add__, b)
Out[5]: [1, 2, 3, 4, 6, 9, 11]

In [6]: sum(b)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 sum(b)

TypeError: unsupported operand type(s) for +: 'int' and 'list'
```

MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py
MBartkowiakSTFC: The 'transmutation' idea in MDANSE is meant to use properties of another atom type than the one stored in the trajectory. As in, take a trajectory of water and then calculate the structure factor using the scattering length of deuterium instead. The '_new_map' ends up used in a different part of the code.

MDANSE/Src/MDANSE/Framework/Configurators/PartialChargeConfigurator.py
MBartkowiakSTFC: We definitely want to accumulate the changes without clearing the _new_map.

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
MBartkowiakSTFC: One complication here is that these values have to be JSON encoded and decoded (and remain human-readable when saved as a text string). I am sure that it can be done. Do you think it is worth the effort?

oerc0122: I was actually referring to the `1`, `2`...
If you had:
```python
class Modes(Enum):
    DIFFERENCE = auto()
    INTERSECTION = auto()
    UNION = auto()
...

        if self.mode_box.currentIndex() == Modes.DIFFERENCE:
            return "difference"
        elif self.mode_box.currentIndex() == Modes.INTERSECTION:
            return "intersection"
        elif self.mode_box.currentIndex() == Modes.UNION:
            return "union"
```
It removes magic numbers and ensures internal consistency a bit better if we ever add other operations.

As an aside, `Enum`s can act as [strings](https://docs.python.org/3/library/enum.html#enum.StrEnum)

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,71 @@
...
+#
+
+from typing import Union, Dict, Any, Set
+
+from MDANSE.Chemistry.ChemicalSystem import ChemicalSystem
+from MDANSE.MolecularDynamics.Trajectory import Trajectory
+
+
+def select_atoms(
+    trajectory: Trajectory, **function_parameters: Dict[str, Any]
```
oerc0122: It's currently recommended that for `**kwargs` you document the types of any variables that they might contain.

https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

MDANSE/Src/MDANSE/Framework/AtomSelector/general_selection.py
```diff
@@ -0,0 +1,76 @@
...
+    ----------
+    trajectory : Trajectory
+        A trajectory instance to which the selection is applied
+
+    Returns
+    -------
+    Set[int]
+        Set of all the atom indices
+    """
+    return set(range(len(trajectory.chemical_system.atom_list)))
```
oerc0122: I'm wondering if we want a utility function for `set(range(len(...)))` or if there's a nicer way of writing it.

MDANSE/Src/MDANSE/Framework/AtomSelector/general_selection.py
```diff
@@ -0,0 +1,76 @@
...
+#    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+from typing import Union, Dict, Any, Set
+from MDANSE.Chemistry.ChemicalSystem import ChemicalSystem
+from MDANSE.MolecularDynamics.Trajectory import Trajectory
+
+
+def select_all(
+    trajectory: Trajectory, **function_parameters: Dict[str, Any]
```
oerc0122: For cases where the arguments are not used internally, it's recommended that they are prefixed with an `_`

MDANSE/Src/MDANSE/Framework/AtomSelector/molecule_selection.py
```diff
@@ -13,35 +13,38 @@
...
     Parameters
     ----------
-    system : ChemicalSystem
-        The MDANSE chemical system.
-    check_exists : bool, optional
-        Check if a match exists.
+    trajectory : Trajectory
+        A trajectory instance to which the selection is applied
+    function_parameters : Dict[str, Any]
+        should include a list of str molecule names under key "atom_labels"
```
oerc0122: If there are sensible options, they should be in the function signature, i.e.
```python
def select_molecules(
    trajectory: Trajectory, *, 
    molecule_names: Sequence[str] | None = None, 
    **_function_parameters: typing.Never
) -> Set[int]:
```
(The lone `*` is to enforce `molecule_names` should only be passed via kwargs).

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+        self.reset()
+
+    def reset(self):
+        self.system = None
+        self.trajectory = None
+        self.all_idxs = set()
+        self.operations = {}
+
+    def set_selection(
+        self, number: Union[int, None] = None, function_parameters: Dict[str, Any] = {}
```
oerc0122: Should function parameters be `**` here?

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+    ):
+        if number is None:
+            number = len(self.operations)
+        else:
+            try:
+                number = int(number)
+            except TypeError:
+                number = len(self.operations)
```
oerc0122: I'm not sure if the best option on failure here is to add a new operation.

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+            True if the selection adds atoms, False otherwise
         """
-        idxs = set([])
-
-        for k, v in self.settings.items():
-            if isinstance(v, dict):
-                args = [{self._kwarg_keys[k]: i} for i in v.keys()]
-                switches = v.values()
+        function_parameters = json.loads(json_string)
+        if len(self.operations) == 0:
```
oerc0122:
```suggestion
        if not self.operations:
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+        if (
+            len(selection.difference(current_selection)) > 0
+            and operation_type == "union"
+        ):
+            return True
+        elif (
+            len(current_selection.difference(selection)) > 0
+            and operation_type != "union"
+        ):
+            return True
```
oerc0122: These reduce to:
```python
if selection.difference(current_selection):
    return True
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+        return False
+
+    def select_in_trajectory(self, trajectory: Trajectory) -> Set[int]:
```
oerc0122: Docstring

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+        selection = set()
+        self.all_idxs = set(range(len(trajectory.chemical_system.atom_list)))
+        sequence = sorted([int(x) for x in self.operations.keys()])
```
oerc0122:
```suggestion
        sequence = sorted(map(int, self.operations))
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+        if len(sequence) == 0:
```
oerc0122:
```suggestion
        if not sequence:
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
+                elif operation_type == "difference":
+                    selection = selection.difference(temp_selection)
                 else:
-                    new_settings[k] = True
-
-        for idx in idxs - added:
-            new_settings["index"][idx] = True
-
-        self.settings = new_settings
+                    selection = temp_selection
```
oerc0122: Seems like this set of operations is repeated. Might be worth making it a function?

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
-                        sub_list.append(k1)
-                if sub_list:
-                    minimal_dict[k0] = sorted(sub_list)
-        return json.dumps(minimal_dict)
+        return json.dumps(self.operations)
 
-    def json_to_settings(self, json_string: str) -> dict[str, Union[bool, dict]]:
-        """Loads the json string and converts to a settings.
+    def load_from_json(self, json_string: str):
+        """_summary_
```
oerc0122: Docstring?

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,170 @@
...
-        -------
-        bool
-            True if settings are valid.
-        """
-        try:
-            settings = self.json_to_settings(json_string)
-        except ValueError:
-            return False
-        return self.check_valid_setting(settings)
+            if isinstance(v0, dict):
```
oerc0122: If it's not a `dict` it's silently ignored? Are we sure that's the behaviour we want?

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Parameters
+    ----------
+    trajectory : Trajectory
+        A trajectory instance to which the selection is applied
+    function_parameters : Dict[str, Any]
+        may include
+        "frame_number" : int
+        "position_minimum" : np.ndarray[float]
+        "position_maximum" : np.ndarray[float]
```
oerc0122: If you're going to have it like this (c.f. comment above about sensible options), please document the defaults.

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Returns
+    -------
+    Set[int]
+        Set of all the atom indices
```
oerc0122: Docstring

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,81 @@
...
+    function_parameters : Dict[str, Any]
+        may include
+        "frame_number" : int
+        "sphere_centre" : np.ndarray[float]
+        "sphere_radius" : float
+
+    Returns
+    -------
+    Set[int]
+        Set of all the atom indices
```
oerc0122: Docstring

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    coordinates = trajectory.coordinates(function_parameters.get("frame_number", 0))
+    sphere_centre = np.array(function_parameters.get("sphere_centre", 3 * [0.0]))
+    sphere_radius = function_parameters.get("sphere_radius", 1.0)
```
oerc0122: I'm not convinced we want defaults here?

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,81 @@
...
+    -------
+    Set[int]
+        Set of all the atom indices
+    """
+    coordinates = trajectory.coordinates(function_parameters.get("frame_number", 0))
+    lower_limits = np.array(function_parameters.get("position_minimum", 3 * [-np.inf]))
+    upper_limits = np.array(function_parameters.get("position_maximum", 3 * [np.inf]))
+    mask1 = np.all(coordinates > lower_limits.reshape((1, 3)), axis=1)
+    mask2 = np.all(coordinates < upper_limits.reshape((1, 3)), axis=1)
+    return set(np.where(np.logical_and(mask1, mask2))[0])
```
oerc0122: `numpy` in general does sensible things:
```python
>>> q = np.arange(12).reshape(4, 3)
>>> lower_limits = np.array([2, 1, 7])
>>> upper_limits = np.array([8, 7, np.inf])
>>> q > lower_limits
array([[False, False, False],
       [ True,  True, False],
       [ True,  True,  True],
       [ True,  True,  True]])
>>> (q > lower_limits) & (q < upper_limits)
array([[False, False, False],
       [ True,  True, False],
       [ True,  True,  True],
       [False, False,  True]])
>>> ((q > lower_limits) & (q < upper_limits)).all(axis=1)
array([False, False,  True, False])
```

```suggestion
    valid = np.where(((coordinates > lower_limts) & 
                      (coordinates < upper_limits)).all(axis=1))
    return set(valid)
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomTransmutationWidget.py
```diff
@@ -137,9 +135,8 @@ def update_transmutation_textbox(self) -> None:
         map = self.transmuter.get_setting()
```
oerc0122: Not you, but `map` here is shadowing the python function.

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -0,0 +1,445 @@
...
+            position of the cursor in the text input field
+
+        Returns
+        -------
+        Tuple[int,str]
+            a tuple of (validator state, input string, cursor position)
+        """
+        state = QValidator.State.Intermediate
+        comma_count = input_string.count(",")
+        if len(input_string) > 0:
```
```suggestion
        if input_string:
```

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -0,0 +1,445 @@
...
+        ----------
+        input_string : str
+            current contents of a text input field
+        position : int
+            position of the cursor in the text input field
+
+        Returns
+        -------
+        Tuple[int,str]
+            a tuple of (validator state, input string, cursor position)
```
oerc0122: The type annotation doesn't match. Of note, Numpy doc allows for documenting multiple returns.
```suggestion
        int
            Validator state.
        str
            Original input string.
        int
            Cursor position.
```

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -0,0 +1,445 @@
...
+        """
+        state = QValidator.State.Intermediate
+        comma_count = input_string.count(",")
+        if len(input_string) > 0:
+            try:
+                values = [float(x) for x in input_string.split(",")]
+            except (TypeError, ValueError):
+                if input_string[-1] == "," and comma_count < 3:
```
```suggestion
                if input_string.endswith(",") and comma_count < 3:
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
oerc0122: Ignore me, I missed the order.

MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -0,0 +1,76 @@
...
+
+from MDANSE.MolecularDynamics.Trajectory import Trajectory
+
+
+def select_atoms(
+    trajectory: Trajectory,
+    *,
+    index_list: Sequence[int] = None,
+    index_range: Sequence[int] = None,
+    index_slice: Sequence[int] = None,
```
oerc0122: Structly pep-484 doesn't allow implicit `None`.

C.f. the end of https://peps.python.org/pep-0484/#union-types

MDANSE/Src/MDANSE/Framework/AtomSelector/group_selection.py
```diff
@@ -0,0 +1,72 @@
...
+        a SMARTS string to be matched
+
+    Returns
+    -------
+    Set[int]
+        Set of atom indices matched by rdkit
+    """
+    selection = set()
+    system = trajectory.chemical_system
+    if rdkit_pattern:
```
oerc0122: Does it make sense to have a default here?

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,194 @@
...
+            a trajectory instance for which current_selection is defined
+        current_selection : Set[int]
+            set of currently selected atom indices
 
         Returns
         -------
-        set[int]
-            The atoms indices.
+        bool
+            True if the selection adds atoms, False otherwise
```
oerc0122:
```suggestion
            True if the operation changes selection, False otherwise
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,194 @@
...
+        else:
+            operation_type = function_parameters.get("operation_type", "union")
+            function = function_lookup[function_name]
+            temp_selection = function(trajectory, **function_parameters)
+            if operation_type == "union":
+                selection = selection.union(temp_selection)
+            elif operation_type == "intersection":
+                selection = selection.intersection(temp_selection)
+            elif operation_type == "difference":
+                selection = selection.difference(temp_selection)
```
oerc0122: Could be either 
```suggestion
            if operation_type == "union":
                selection |= temp_selection
            elif operation_type == "intersection":
                selection &= temp_selection
            elif operation_type == "difference":
                selection -= temp_selection
```
Or the shorter, but slightly more esoteric
```suggestion
            if operation_type in ("union", "intersection", "difference"):
                selection = getattr(selection, operation_type)(temp_selection)
```

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,97 @@
...
+        trajectory frame at which to check the coordinates, by default 0
+    position_minimum : Sequence[float], optional
+        (x, y, z) lower limits of coordinates to be selected, by default None
+    position_maximum : Sequence[float], optional
+        (x, y, z) upper limits of coordinates to be selected, by default None
+
+    Returns
+    -------
+    Set[int]
+        _description_
```
oerc0122: Docstring

MDANSE/Src/MDANSE/Framework/Configurators/AtomSelectionConfigurator.py
```diff
@@ -132,14 +134,11 @@ def get_information(self) -> str:
 
         return "\n".join(info) + "\n"
 
-    def get_selector(self) -> Selector:
+    def get_selector(self) -> "ReusableSelection":
```
oerc0122: Am I wrong in thinking this is imported?
```suggestion
    def get_selector(self) -> ReusableSelection:
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -13,25 +13,103 @@
...
+    LabelSelection,
+    PositionSelection,
+    SphereSelection,
+)
+
+
+class SelectionValidity(StrEnum):
+    VALID_SELECTION = "Valid selection"
+    USELESS_SELECTION = "Selection did not change. This operation is not needed."
+    MALFORMED_SELECTION = "This is not a valid JSON string."
```
oerc0122:
```suggestion
    MALFORMED_SELECTION = "This is not a valid selection string."
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,234 @@
...
-            if isinstance(self.settings[k0], dict):
-                for k1, v1 in v0.items():
-                    self.settings[k0][k1] = v1
-            else:
-                self.settings[k0] = v0
+        """
+        if number is None:
+            number = len(self.operations)
+        else:
+            number = int(number)
```
oerc0122: If it's not too long an option is:
```suggestion
        number = int(number) if number is not None else len(self.operations)
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,234 @@
...
+        return bool(
+            (
+                len(selection.difference(current_selection)) > 0
+                and operation_type == "union"
+            )
+            or (
+                len(current_selection.difference(selection)) > 0
+                and operation_type != "union"
+            )
+        )
```
oerc0122: Give `if len(Collection) > 0` is equivalent to `if Collection` we can simplify this to:
```suggestion
        return ((selection - current_selection) and operation_type == "union") or
                (current_selection - selection) and operation_type != "union"))
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -13,362 +13,234 @@
...
-            settings = self.json_to_settings(json_string)
-        except ValueError:
-            return False
-        return self.check_valid_setting(settings)
+        json_setting = json.loads(json_string)
+        for k0, v0 in json_setting.items():
+            if isinstance(v0, dict):
+                self.set_selection(number=k0, function_parameters=v0)
+            else:
+                raise TypeError(f"Selection {v0} is not a dictionary.")
```
oerc0122:
```suggestion
            if not isinstance(v0, dict):
                raise TypeError(f"Selection {v0} is not a dictionary.")
            self.set_selection(number=k0, function_parameters=v0)
```

MDANSE/Src/MDANSE/Framework/AtomSelector/spatial_selection.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    coordinates = trajectory.coordinates(frame_number)
+    if position_minimum is None:
+        lower_limits = np.array(3 * [-np.inf])
+    else:
+        lower_limits = np.array(position_minimum)
+    if position_maximum is None:
+        upper_limits = np.array(3 * [np.inf])
+    else:
+        upper_limits = np.array(position_maximum)
```
oerc0122:
```suggestion
    lower_limits = (np.array(position_minimum)
                    if position_minimum is not None else
                    np.array([-np.inf] * 3))
    upper_limits = (np.array(position_maximum)
                    if position_maximum is not None else
                    np.array([np.inf] * 3))
```

MDANSE/Src/MDANSE/Framework/Configurators/AtomSelectionConfigurator.py
```diff
@@ -78,35 +88,39 @@ def configure(self, value: str) -> None:
...
+    def get_indices(self) -> dict[str, list[int]]:
+        """Group atom indices per chemical element.
+
+        Returns
+        -------
+        dict[str, list[int]]
+            For each atom type, a list of indices of selected atoms
+
+        """
         indicesPerElement = {}
```
oerc0122: Could be simplified to:
```python
indicesPerElement = defaultdict(list)
for i, v in enumerate(self["names"]):
    indicesPerElement[v].extend(self["indices"][i])
```

oerc0122 (approved):
Looks good to me if it does what it's meant to, just a couple of possible stylistic suggestions. Feel free to ignore if you feel they remove clarity.

ChiCheng45:
Some issues I've found and comments.

1. The default selection string in the gui `{"all": true}`, this causes the job to crash.
2. Can the invert button be on its own line? It's slightly confusing as it's in the ALL atoms section.
3. Selecting or deselecting in the 3D view doesn't update the selection setting.
4. Odd behaviours with the atom transmutation and the reset.

Animation64

5. Invalid SMARTS causes the job to fail.
6. The setting dragging causes the setting to disappear.

Animation65

7. The setting `{"0": {"function_name": "invert_selection"}}` is ambiguous; at the moment, it's equivalent to selecting all atoms. I guess this is fine but maybe it should be an invalid setting similarly to something like `{"0": {"function_name": "select_atoms", "atom_types": ["H"], "operation_type": "difference"}}`.

MBartkowiakSTFC:
> Some issues I've found and comments.
> 
> 1. The default selection string in the gui `{"all": true}`, this causes the job to crash.

Fixed.

> 2. Can the invert button be on its own line? It's slightly confusing as it's in the ALL atoms section.

I could argue that it is an operation on all atoms, since `inverted selection = ALL atom indices - current selection`
But this is mainly because I think it would take a lot of space in the layout, will be different to all the other widgets (since it is just one button) and it may still be hard to find after all this.

> 3. Selecting or deselecting in the 3D view doesn't update the selection setting.

Now that the selection is a sequence of operations, it is less clear how the clicks should be handled. Two options I can think of:
- each click creates its own selection operation on a single atom
- while you are clicking, indices are added to a single operation. If you do something else in the meantime, a new selection operation will be added next time you click.
Suggestions welcome.

> 4. Odd behaviours with the atom transmutation and the reset.

This *is* weird and I have not found the reason for it.

> 5. Invalid SMARTS causes the job to fail.

Fixed.

> 6. The setting dragging causes the setting to disappear.

Most likely I will need to subclass QListView to fix this. For now I will disable dragging.

> 7. The setting `{"0": {"function_name": "invert_selection"}}` is ambiguous; at the moment, it's equivalent to selecting all atoms. I guess this is fine but maybe it should be an invalid setting similarly to something like `{"0": {"function_name": "select_atoms", "atom_types": ["H"], "operation_type": "difference"}}`.

I could make sure that the selection widget always starts with an explicit "select all" instead of being empty. Because at the moment empty selection is assumed to mean "select all", but the first selection replaces it.

ChiCheng45:
Since we are letting the user enter SMARTS patterns for matching, I think we need to try to tackle #286 at some point. Currently, we don't have bonding information, so users can only use "C\~C" for two carbon atoms bonded using any bond ("\~") and things like "c:c" (aromatic carbons joined by an aromatic bond) and "c-c" (aromatic carbons joined by a single bond (e.g. biphenyl)) currently do not work.

MBartkowiakSTFC:
> Some issues I've found and comments.
> 
> 1. The default selection string in the gui `{"all": true}`, this causes the job to crash.
> 2. Can the invert button be on its own line? It's slightly confusing as it's in the ALL atoms section.
> 3. Selecting or deselecting in the 3D view doesn't update the selection setting.
> 4. Odd behaviours with the atom transmutation and the reset.
> 5. Invalid SMARTS causes the job to fail.
> 6. The setting dragging causes the setting to disappear.
> 7. The setting `{"0": {"function_name": "invert_selection"}}` is ambiguous; at the moment, it's equivalent to selecting all atoms. I guess this is fine but maybe it should be an invalid setting similarly to something like `{"0": {"function_name": "select_atoms", "atom_types": ["H"], "operation_type": "difference"}}`.

OK, I think that I managed to improve matters regarding points 1, 4, 5, 6 and 7.
I am happy to ignore point 2 for now, just because it is not urgent. We can agree on a solution and fix it.
Point 3 will require more work. I think I will open a new issue.

PR #674, opened 2025-02-18
Add Python 3.13 to the workflow matrix

### LINKED ISSUE #614: [BUG] Python3.13 stability
**Description of the error**
As of today, VTK 9.4.0 became available, which includes wheels for Python 3.13.
rdkit for python 3.13 is still missing.

It is possible to install and run MDANSE by removing rdkit from the code. This affects atom selection options and automatic molecule naming.

Ideally, MDANSE should run on Python 3.13 same as on 3.9-3.12.

Problems observed so far:
1. The MDAnalysis converter crashed with a segmentation fault when converting a Gromacs trajectory from SlimMD (CCPBioSim). This was **not reproducible** and did not happen at the second attempt.
2. On exiting GUI, an error typically appears in the text console:
```
Exception ignored in atexit callback <function shutdown at 0x147aa5940>:
  File "/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/logging/__init__.py", line 2242, in shutdown
    if getattr(h, 'flushOnClose', True):
RuntimeError: wrapped C/C++ object of type JobEntry has been deleted
```

**Suggested fix**
Some initial errors could be caused by files created by the older versions, which might not be fully compatible. This includes the Qt settings, database files, configuration files and instrument database. It has to be investigated if switching between Python versions makes the problems reappear.
Qt object deletion can be investigated and improved.

**Additional details**
More problems will be found as the testing continues. They should be mentioned here.

MBartkowiakSTFC:
**Description of work**
Python 3.13 seems to provide all our dependencies now. This PR adds Python 3.13 to the workflows.

closes #614

**Fixes**
Added 3.13 to the list of Python versions on which we run the workflows.

**To test**
All tests should pass.

### DIFF
--- a/.github/workflows/test_pip.yml
+++ b/.github/workflows/test_pip.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     if: |
       !contains( github.ref, 'legacy' )
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         # os: [ubuntu-20.04, windows-2019, macOS-11, ubuntu-latest, windows-latest, macOS-latest]
     if: |
       !contains( github.ref, 'legacy' )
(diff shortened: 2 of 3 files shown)


MBartkowiakSTFC:
@oerc0122 For now I just extended the existing lists of Python versions. Does it make sense to keep multiple versions in `wheels.yml` now that we don't have Cython extensions?

oerc0122:
If it's pure Python, you shouldn't need variant wheels. 

I recommend keeping the CIBW style around as at some point there is a suggestion of C++/Rust acceleration, but I suspect we can get away without multiple builds.

The question is, do we have tests for the PyPI pushed versions, and do we want some? Euphonic has some [examples](https://github.com/pace-neutrons/Euphonic/blob/master/.github/workflows/test_release_pypi.yml) we might ~steal~ be inspired by.

oerc0122 (approved):
Good for now, let's get this in and tested. 

I think further discussion on "more efficient" wheels can be dealt with separately.

PR #677, opened 2025-02-21
Save the MSD results with equal weights to GDISF output

### LINKED ISSUE #660: [ENHANCEMENT] Save MSD result in the GDISF calculations
Since the MSD is calculated in the GDISF we might as well save this into the results file. This will save the user from having to calculate the MSD again.

ChiCheng45:
**Description of work**
Saved the MSD results to the GDISF output. This is saved with equal weights since GDISF ~ exp( - q * MSD / 6) and the MSD isn't scaled by the weights.

**Fixes**
Fixes #660

**To test**
Run the GDISF calculation and check that the MSD results are saved to the outputs. The MSD results should be the same as the results from the MSD calculation with equal weights.


MBartkowiakSTFC (approved):
I get the same MSD, in all the runs, independent of the weights chosen for the GDISF. Looks good.

PR #678, opened 2025-02-21
Use pytest for test_analysis and fix test

oerc0122:
**Description of work**
- Migrate from `unittest` to `pytest`
- Parametrise tests to add capacity for further testing
- Merge `test_molecular_dynamics` with `test_analysis`

**Fixes**
- Fix bad test which claims to run `radius_of_gyration` with root, but actually runs `mean_square_fluctuation`

**To test**
Standard tests


MDANSE/Tests/UnitTests/test_analysis.py
```diff
@@ -1,67 +1,57 @@
 import unittest
```
MBartkowiakSTFC: This is probably not needed after changing to pytest.

oerc0122:
```suggestion
```

MBartkowiakSTFC (approved):
Looks good, I don't see any problems.

PR #681, opened 2025-02-26
Use pytest for `test_units`

oerc0122:
**Description of work**
- Refactor units testing to use pytest
- Refactor _Unit class to use dunders
- Replace `Equivalences` dict with `defaultdict`

**Fixes**
N/A

**To test**
Standard tests

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Units.py
+++ b/MDANSE/Src/MDANSE/Framework/Units.py
@@ -80,7 +80,7 @@ def get_trailing_digits(string: str) -> Tuple[str, int]:
     Parameters
     ----------
     string : str
-        FIXME: Add docs.
+        String to parse.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


MDANSE/Src/MDANSE/Framework/Units.py
```diff
@@ -364,6 +364,76 @@ def __rsub__(self, other):
 
         return other.__sub__(self)
 
+    def __ceil__(self):
+        """Ceil of a _Unit value in canonical units.
+
+        >>> print(measure(10.2, 'm/s').ceiling())
+        10.0000 m / s
```
MBartkowiakSTFC: I know that these docstrings were already written like this before you made changes, but I think that the examples given here are wrong.

The docstrings below for `__floor__` and `__round__` should be checked too.

Overall, I would expect 3.6 m/s to be 12.96 km/h, and the ceil, round, floor to give 13, 13 and 12, respectively.

oerc0122: Reworked the docstrings throughout and made sure it was valid under doctest.

MBartkowiakSTFC (approved):
I have not found any problems. Looks good.

PR #684, opened 2025-02-27
Use pytest for `test_configurable`

oerc0122:
**Description of work**
- Refactor configurable testing to use `pytest`

**Fixes**
N/A

**To test**
Standard tests.


MBartkowiakSTFC (approved):
Looks cleaner now. Also, the test is more complete now with the keyword arguments being used too.

Issue #673, opened 2025-02-18
[BUG] Unit tests time out randomly on ubuntu-latest


MBartkowiakSTFC:
**Description of the error**
Our unit test workflow seems to get stuck on the `ubuntu-latest` runner. We have 4 workflows on `ubuntu-latest`, for different versions of Python. Sometimes one out of four (or, rarely, two out of four https://github.com/ISISNeutronMuon/MDANSE/actions/runs/13350554804 ) ends up marked as failed, with the error:
`The job running on runner GitHub Actions X has exceeded the maximum execution time of 360 minutes.`

**Describe the expected result**
Most of the time, if the tests all passed on other platform, they should also pass on Linux.

**Describe the actual result**
Usually, the test all complete without problems. Sometimes, the unit tests stop running after the first file. The output is:
`Analysis/test_average_structure.py ................                                           [  2%]`
and the next test in 
`Analysis/test_dynamics.py`
either never starts, or never completes.

In the end the entire workflow times out.

**Suggested fix**
Since it is not reproducible, and seems only to affect one platform, it is difficult to recognise if the error is on our side, or is a problem with the runner. However, if the test fails, it seems to fail always at the same point in the workflow, so maybe the unit tests need to be changed somehow.

From the different point of view, we could also lower the time limit for the workflows, so they fail sooner. We don't normally need more than 20 minutes, so a limit of 30 minutes would be more than enough.

**Additional details**
N/A

MBartkowiakSTFC:
It seems that the tests always get stuck on the Ubuntu runner, but not always on the same test after all. Last failure:
```
Run cd MDANSE/Tests/UnitTests
============================= test session starts ==============================
platform linux -- Python 3.12.9, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/runner/work/MDANSE/MDANSE/MDANSE
configfile: pyproject.toml
collected 702 items

Analysis/test_average_structure.py ................                      [  2%]
Analysis/test_dynamics.py .............................................. [  8%]
Error: The operation was canceled.
```

MBartkowiakSTFC:
And another one, also on Ubuntu:
```
Run cd MDANSE/Tests/UnitTests
============================= test session starts ==============================
platform linux -- Python 3.12.9, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/runner/work/MDANSE/MDANSE/MDANSE
configfile: pyproject.toml
collected 702 items

[...]

PR #685, opened 2025-02-27
Add timeout to MDANSE unit tests

MBartkowiakSTFC:
**Description of work**

Following the discussion of #673 we add timeout to unit tests.

**Fixes**
Added pytest-timeout to the workflow.
Set tests timeout in pyproject.toml.

**To test**
Tests should pass same as before.


oerc0122 (approved):
The problem, of course, being that this is an intermittent failure not limited to any particular Python version. 

Doesn't seem to have happened, but at worst this should be a slight preventative measure which should allow us to find out what's actually going on.

PR #689, opened 2025-03-03
Add new adjusted weight scheme to work with DCSF and CCF jobs

### LINKED ISSUE #683: [BUG] DCSF terms are not scaled similarly to the coherent results in the NDTSF calculations
**Description of the error**
Here I calculated the DCSF and NDTSF of an argon trajectory but I set 1536 to NewAr36 and 512 to Ar36. Both Ar36 and NewAr36 have the same scattering lengths so the the incoherent scattering is zero and the total scattering will be equal to the coherent.

The issue looks like is due to the $1/(N_iN_j)^{0.5}$ extra scaling in the DCSF calculation, it should be $1/N_iN_j$ so it cancels with the $N_iN_j$ in the weighting parameters. Alternatively, the weighting parameters should be changed.

All DCSF and NDTSF coherent terms should be equal to each other up to a scaling factor.

Image

Image

ChiCheng45:
**Description of work**
Some prefactors involving two atom types were not compatible with the MDANSE weight scheme, so the weighted total was not equal to (or equal to within a scaling factor) the proper total result.

The following changes have been made.

- Currently in the DCSF 
```math
F_{\text{coh},\alpha\beta}{(\mathbf{q},t) =  \frac{1}{N\sqrt{c_{\alpha}c_{\beta}}}}{\sum\limits_{i \in \alpha}{\sum\limits_{j \in \beta }\left\langle {\exp\left\lbrack {{- i}\mathbf{q}\cdot\mathbf{r}_{i}\left( 0 \right)} \right\rbrack\exp\left\lbrack {i\mathbf{q}\cdot\mathbf{r}_{j}\left( t \right)} \right\rbrack} \right\rangle}}.
```
A new adjusted weight scheme has been created
```math
W_{\alpha\beta} = \frac{\sqrt{c_{\alpha} c_{\beta}}  b_{\alpha} b_{\beta} }{\sum_{\gamma\delta} c_{\gamma} c_{\delta} b_{\gamma} b_{\delta}}
```
to be used with DCSF and CCF. With this weight scheme, the DCSF total is calculated properly 
```math
F_{\text{coh},\alpha\beta}(\mathbf{q},t) =\frac{1}{\sum_{\gamma\delta}c_{\gamma}c_{\delta}b_{\delta}b_{\gamma}} \sum_{\alpha\beta} \sqrt{c_{\alpha}c_{\beta}}b_{\alpha}b_{\beta}F_{\text{coh},\alpha\beta}(\mathbf{q},t)
=\frac{1}{\sum_{\gamma\delta}c_{\gamma}c_{\delta}b_{\gamma}b_{\delta}}  \frac{1}{N} \sum_{ij} b_{i}b_{j}\left\langle {\exp\left\lbrack {{- i}\mathbf{q}\cdot\mathbf{r}_{i}\left( 0 \right)} \right\rbrack\exp\left\lbrack {i\mathbf{q}\cdot\mathbf{r}_{j}\left( t \right)} \right\rbrack} \right\rangle
```
therefore
```math
F_{\text{coh},\alpha\beta}(\mathbf{q},t) =\frac{1}{\sum_{\gamma\delta}c_{\gamma}c_{\delta}b_{\gamma}b_{\delta}} F_{\text{coh},\alpha\beta}^{\mathrm{NTDSF}}(\mathbf{q},t) 
```
so the results from the DCSF are just a scaled version of the NTDSF results.

- CCF has been corrected so that the prefactor goes ~ $1 / N \sqrt{c_{\alpha}c_{\beta}}$ so that it is the same as the DCSF.
- The diagonal part of the distinct part of the van Hove function goes  ~ $1 / \rho N c_{\alpha}c_{\alpha}$; previously, it was ~ $1 / \rho c_{\alpha} (N c_{\alpha} - 1)$. They are the same for large $N$, but the change is made for consistency with the weight scheme and everything else.
- Similar changes made for the van Hove function were done for the SSF, PDF, X-ray, and coordination number jobs.
- Fix X-Ray and SSF intra+inter partial results so that they can be plotted with or without the weights applied.
- Added an NTDSF test.

This makes the MDANSE calculations consistent with itself. Results from the DCSF and DISF are consistent with the NTDSF results (they are the same up to some scaling factor).  Here, I plot the $F_{\text{coh}}(q,0)$ results from the DCSF and NTDSF calculation using an Argon trajectory where I've randomly assigned the atom types to Test1 and Test2 which have the same b_inc and b_coh but with concentrations of 0.25 and 0.75.

image

Note that I plot the weighted partials $W_{\alpha\beta} F_{\text{coh},\alpha\beta}(q,0)$ from DCSF and $\sqrt{c_{\alpha}c_{\beta}}b_{\alpha} b_{\beta} F_{\text{coh},\alpha\beta}(q,0)$ from NTDSF so that you can see they are the same as the NTDSF except for some scaling factor. I have updated NTDSF results file so that you can also plot $F_{\text{coh},\alpha\beta}(q,t)$ by deselecting the "apply weight?" option in the plotter as with the DCSF results.

I also did the test for the SSF and DCSF for this system.

image

The blue plots are from the (unweighted) partials SSF results, and the orange plots  are the partial intermediate scattering function from the DCSF for t=0 converted to the F-Z formalism using

```math
S^{\text{A-L}}_{\alpha\beta}(Q) = \delta_{\alpha\beta} + \sqrt{c_{\alpha}c_{\beta}} (S^{\text{F-Z}}_{\alpha\beta}(Q) -1)
```

**Fixes**
Fixes #683

**To test**
Run the DCSF, DISF and NTDSF calculations. Compare the DCSF results with the coherent results in NTDSF they should be equivalent to some scaling factor.


MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
```diff
@@ -416,12 +417,12 @@ def finalize(self):
             ][:]
 
         # Compute incoherent functions and structure factor
-        for element in nAtomsPerElement:
+        for element, number in list(nAtomsPerElement.items()):
```
oerc0122: No need to cast to list.
```suggestion
        for element, number in nAtomsPerElement.items():
```

gonzalezma:
I see a potential issue with this redefinition. The original definition (A-L scheme) is reported in the literature and the partials have a well defined limit, as F(Q --> infinity, t=0) = delta_alpha,beta. The new definition does not correspond to neither the F-Z nor the A-L definitions, so the users should be warned about how to proceed if they want to use the partials.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -259,14 +259,13 @@ def finalize(self):
         assign_weights(self._outputData, weight_dict, "s(q,f)_%s")
         for element, number in list(nAtomsPerElement.items()):
             extra_scaling = 1.0 / number
-            self._outputData[f"f(q,t)_{element}"].scaling_factor *= extra_scaling
+            self._outputData[f"f(q,t)_{element}"] *= extra_scaling
```
oerc0122, line 262: Curious about this being reverted here.

MDANSE/Tests/UnitTests/Analysis/test_scattering.py
```diff
@@ -52,8 +52,8 @@ def dcsf():
     parameters["instrument_resolution"] = ("Ideal", {})
     parameters["output_files"] = (temp_name, ("MDAFormat",), "INFO")
     parameters["q_vectors"] = (
-        "SphericalLatticeQVectors",
-        {"seed": 0, "shells": (5.0, 36, 10.0), "n_vectors": 10, "width": 9.0},
+        "GridQVectors",
```
oerc0122, line 55: Why the change to `GridQ`?

oerc0122 (commented):
Seems fairly sensible. A couple of questions just for my own understanding of the changes.

MDANSE/Tests/UnitTests/Analysis/test_scattering.py
ChiCheng45: I updated NDTSF so that it is tested against reference data which uses the GridQVectors. It has to be switched because sphericallattiveqvectors are randomly generated while the GridQvectors are not.

oerc0122: Do we need seed fixing? Not for this PR, just in general?

ChiCheng45:
> I see a potential issue with this redefinition. The original definition (A-L scheme) is reported in the literature and the partials have a well defined limit, as F(Q --> infinity, t=0) = delta_alpha,beta. The new definition does not correspond to neither the F-Z nor the A-L definitions, so the users should be warned about how to proceed if they want to use the partials.

I had a think about this, we can change it so that MDANSE will have two weight schemes for properties that depend on two atom types. X-Ray, van Hove, SSF, coordination number, and PDF will use the current one, while DCSF and CCF will use a new one. DCSF and CCF will follow A-L definitions as in 1.5, but the weight scheme will be adjusted so the total is corrected.

The new weight scheme will be

```math
W_{\alpha\beta} = \frac{\sqrt{c_{\alpha} c_{\beta}}  b_{\alpha} b_{\beta} }{\sum_{\gamma\delta} c_{\gamma} c_{\delta} b_{\gamma} b_{\delta}}
```

where I think it's probably best to leave the denominator the same across the two weight schemes.

@gonzalezma is this preferable?

gonzalezma:
I think so, but we will need to check the results. In particular, I don't know how the CCF is calculated now in MDANSE and I think that I have only seen the definition of the CCF for a monatomic liquid, so I am not completely sure of the best way to handle this in a polyatomic system.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
ChiCheng45: This was changed during the NDTSF update but this extra_scaling should have been left in the partial terms so that it has the prefactor 1 / N c_i (where N c_i = N_i). This means that the partial terms are not dependent on the number of atoms.

MDANSE/Tests/UnitTests/Analysis/test_scattering.py
ChiCheng45: You can fix the seed, but I think you will get different q vectors dependent on the machine or library versions, etc which would cause the tests to fail.

MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
```diff
@@ -400,34 +400,53 @@ def finalize(self):
...
+                )
             else:
-                self._outputData[f"f(q,t)_coh_{pair_str}"] *= 2 * bi * bj * norm_natoms
-                self._outputData[f"s(q,f)_coh_{pair_str}"] *= 2 * bi * bj * norm_natoms
+                self._outputData[f"f(q,t)_coh_{pair_str}"].scaling_factor *= (
+                    2 * bi * bj * sqrt_cij
+                )
+                self._outputData[f"s(q,f)_coh_{pair_str}"].scaling_factor *= (
+                    2 * bi * bj * sqrt_cij
+                )
```
oerc0122: I do think if we're rewriting this it might be slightly clearer to have:
```python
pre_fac = 2 if pair[0] == pair[1] else 1
self._outputData[f"f(q,t)_coh_{pair_str}"].scaling_factor *= pre_fac * bi * bj * sqrt_cij
self._outputData[f"s(q,f)_coh_{pair_str}"].scaling_factor *= pre_fac * bi * bj * sqrt_cij
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -19,7 +19,9 @@
 import numpy as np
 
 
-def get_weights(props: Dict[str, float], contents: Dict[str, int], dim: int):
+def get_weights(
+    props: Dict[str, float], contents: Dict[str, int], dim: int, c_exp: float = 1.0
```
oerc0122: Not sure I like the name `c_exp` reminds me of the function `cexp(x)` which is `e**(i*x)`. It is, however, documented clearly, so I'm happy to keep it if there's not a better name.

oerc0122 (commented):
All looks good functionally. 

Just to double check, are there comparison plots of these new factors?

ChiCheng45:
> All looks good functionally.
> 
> Just to double check, are there comparison plots of these new factors?

The plots will be the weighted versions, so they are scaled by the scaling factor. The recent changes allow you to plot the unweighed ones as well; previously, you could only plot the weighted ones, as the scaling factor was set to 1.

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
ChiCheng45: Good point, I've changed it to conc_exp.

MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
ChiCheng45: Yes this is clearer I've changed this now.

oerc0122 (approved):
Happy with this if it's working as it should. LGTM.

PR #687, opened 2025-02-28
Add test helpers

oerc0122:
**Description of work**
- Add test_helpers 
  - Add `paths` test helper
  - Add `compare_mdt` to compare MDANSE trajectories.
  - Add `get_deep_atr` to extract attributes from nested objects.
- Refactor `test_converter`, `test_editor` and the `Analysis` folders to use new routines as well as `pytest`.

**Fixes**
Fixes #676 

**To test**
Standard tests.

PR #710, opened 2025-03-06
Update tests to use assert_allclose

ChiCheng45:
**Description of work**
Update array test to use assert_allclose. A lot of tests used assert_array_almost_equal, which only checks if `abs(desired-actual) < 1.5 * 10**(-decimal)` where `decimal=6` this is not good when checking values that are small.

**To test**
Test must pass


oerc0122:
Suggest we rebase this onto #687 ?

ChiCheng45:
> Suggest we rebase this onto #687 ?

Yeah could do. Maybe we should fix this branch first?

oerc0122:
Included in #687

PR #712, opened 2025-03-09
Documentation clean up and update

### LINKED ISSUE #375: [BUG] When "weights" is used is the calculated result really a total?
**Description of the error**
According to the documentation https://mdanse.readthedocs.io/en/protos/pages/parameters.html#weights the "total" property of a result is calculated using the following equations,

image

the first equation is actually the formula for a weighted arithmetic mean https://en.wikipedia.org/wiki/Weighted_arithmetic_mean#Mathematical_definition.

Therefore properties which are calculated with the "weights" setting are actually some sort of weighted arithmetic mean over the different atoms types.

**Suggested fix**
Update the documentation and code so that "total" results calculated with "weights" are referred to as something else.

### LINKED ISSUE #411: Update DOS documentation
**Description of the error**
Update the documentation to explain the meaning of some example weight settings that could be used.

**Additional details**
Documentation:
image

ChiCheng45:
**Description of work**
Update to the documentation.

- Remove some old rst files which were no longer used.
- Removed unused images.
- Changed Correlation Functions section to Time Series, it now includes a section on resolution function used in MDANSE.
- Added Weight Scheme section which describes how the total are summed up from the partials.
- Added q-Vector Generation section.
- Updated all analysis job sections. In particular, all equations have been rewritten so that they follow the same notation throughout. Analysis job that are no longer available or are still in development have been marked.
- Added Symbols and Notation section.
- Removed a number of How-To Guide sections.
- Removed Dynamic Coherent Structure Factor tutorial.

Feel free to make edits to this branch.

**Fixes**
Closes #411
Closes #375


MBartkowiakSTFC (approved):
Thanks for making the changes! The weights section helps a lot.

I think this can be merged.

PR #714, opened 2025-03-10
Treat all fields in ElementsDatabaseEditor as text fields

### LINKED ISSUE #713: [BUG] Chemical element editor does not handle small numbers
**Description of the error**
When editing the atom properties in the MDANSE_GUI, a new element can be edited freely. However, for the existing elements the table creates spin boxes which only have 1e-2 precision.

Any number should be allowed in the numerical inputs

Small numbers are just rounded to 0.

**Suggested fix**
The delegate widget for the element property table should always be a line edit and not a spinbox.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Changes to the chemical element editor have been implemented to prevent small numbers from being rounded to 0. Color data is updated using the color picker.

**Fixes**
Display role of QStandardItem has been removed, preventing QSpinBox delegate from appearing.

**To test**
Start the element editor dialog. Try changing numbers, especially for b_coherent and b_incoherent.


ChiCheng45:
I've also been working on the atom editor (#748) as tutorial 4 https://github.com/ISISNeutronMuon/MDANSE-Examples/pull/17 needs the user to make their own atom types by copying another one. @MBartkowiakSTFC Can we change this PR so it merges into #748, or if you prefer, we can do the opposite.

MBartkowiakSTFC:
I changed the base to #748 for now. If you prefer, you could also just merge this branch into your branch and resolve the conflicts then, and we would close this PR afterwards. Otherwise I will fix this one later this week.

ChiCheng45:
> I changed the base to #748 for now. If you prefer, you could also just merge this branch into your branch and resolve the conflicts then, and we would close this PR afterwards. Otherwise I will fix this one later this week.

Great, I made some updates so that the color strings are set with a default color and can be updated with a color picker then merged this into #748. #748 should be ready for review now.

PR #716, opened 2025-03-10
Fix to images, links and equations in documentation 

ChiCheng45:
**Description of work**
Fixed image names in https://mdanse.readthedocs.io/en/protos/pages/weights.html and fixed broken link in index page.


oerc0122 (approved):
All looks sensible.

PR #719, opened 2025-03-13
Improve plotter code

### LINKED ISSUE #715: [BUG] PlottingContext allocates too much RAM to multidimensional data
**Description of the error**
PlottingContext creates a separate 1D array for every row of a multidimensional array.

It should only process the rows that are being plotted.

Memory use increases rapidly into GB range.

**Suggested fix**
The `SingleDataset.curves_vs_axis()` method should be lazy. It can be passed a set of indices to use.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
This PR moves unit conversion from the plotter code to the SingleDataset class. Also, SingleDataset now only creates 1D arrays for plotting when it is needed, reducing the plotter's memory requirements.

closes #715 

**Fixes**
1. Unit conversion is now done inside SingleDataset.
2. `SingleDataset.curves_vs_axis` now only returns a limited number of curves.
3. Legend labels of 1D subplots now contain information about the curve's position on the other axes.
4. Docstrings have been added to most methods of the plotter code.

**To test**
All tests should pass.
Load an analysis result with 2D data arrays (e.g. van Hove function, dynamic coherent structure factor, etc.). Plot it using all plotters (Single, Grid, Heatmap, Text). Axis label and plot legend should contain information about the position of the individual curves ("at q=1.5 1/nm"). The labels should update when the preferred unit is changed.


ChiCheng45 (commented):
Looks good just a few things.

- Single plot label with grid has an "at" at the end.

image

- Plotting failure with molecular trace and the heatmap plot.

MBartkowiakSTFC:
Plots should now have correct labels for the following cases where we had problems before:
1. The lone "at" should not appear in Grid plots,
2. Heatmap plots should work for molecular trace,
3. In the case of Heatmap plots for multiple datasets, only the first dataset is plotted, and the other datasets do not affect the axis labels.

ChiCheng45 (approved):
Look good. I think this is ready to go in.

PR #722, opened 2025-03-17
[RFC] Qvectors rework

### LINKED ISSUE #446: [ENHANCEMENT] Q vector generation: path in reciprocal space

### LINKED ISSUE #458: [ENHANCEMENT] Q vector generation: load a user-provided array
Some users dealing with crystalline systems have measured inelastic neutron scattering data at specific values of H, K, L, dE. Instead of using a q-vector generator, they would prefer MDANSE just to load a list of HKL values which they already prepared elsewhere. 

A mock generator could be implemented, which takes as an input parameter the filename from which to load the vectors.

Going back to the way MDANSE used to work, we could then also save our own vector definitions as files and load them for analysis runs.

**Additional context**
The important thing is to save the q vectors together with the other input files, and not in a global definition directory.
The format of the q vector file remains to be specified.

### LINKED ISSUE #656: [ENHANCEMENT] Add a lattice q-vector generator that does not depend on the width option
The current lattice q-vector generator does not allow width=0. Setting a small width with the current lattice q-vector generators means that I might get zero valid q-vectors for a specific value of |q|,

Add a new lattice q-vector generator which generates all valid lattice q-vectors within a specific range. Unlike the current lattice q-vector generators, the q-vector step won't be constant. We might need an option to set the maximum number of q-vectors used for a given shell. Adding a preview of the q-vector shells that will be created with this generator in the GUI might also be useful.

### LINKED ISSUE #705: [ENHANCEMENT] Q-Vectors rework
Q-Vectors are currently somewhat limited in scope apparently to $\|Q\|$ calculations using a class hierarchy.

These generators will need to have the capacity to be passed various arguments as needed.

The proposal is to rework the Q-vector generation system as actual generators to allow infinite and finite Q-Vector generators with paths (#446), explicit values (#458), $\|Q\|$ functionality.

<details>
<summary> Current Q Vector generators </summary>

- ApproximateDispersionQVectors.py 
- CircularQVectors.py           
- GridQVectors.py  
- LatticeQVectors.py        
- LinearQVectors.py         
- SphericalLatticeQVectors.py 
- CircularLatticeQVectors.py        
- DispersionLatticeQVectors.py  
- IQVectors.py     
- LinearLatticeQVectors.py  
- MillerIndicesQVectors.py  
- SphericalQVectors.py

</details>

<details>
<summary> Proposed New Q Vector generators </summary>

- Q-vectors from generator function
- Q-vectors from list/array
- Q-vectors from paths

</details>

N/A

**Additional context**
N/A

oerc0122:
**Description of work**
Implement independent new scheme for handling Q-Vectors without shells.

**Fixes**
Fixes #656 
Fixes #458 
Fixes #446 
Fixes #705 

**To test**
New tests in UnitTests/NewQVectors


MDANSE/Src/MDANSE/Core/SubclassFactory.py
```diff
@@ -201,3 +203,75 @@ def indirect_subclass_dictionary(cls):
             dict(str:type) -- a dictionary of the subclasses of this class
         """
         return recursive_dict(cls)
+
+class RegisterFactory(ABC, Generic[T]):
```
oerc0122: This can be broken out into a separate PR if we prefer.

MBartkowiakSTFC:
Would you agree that the main reason for the RegisterFactory class is to register the subclasses explicitly and not automatically?

oerc0122:
> Would you agree that the main reason for the RegisterFactory class is to register the subclasses explicitly and not automatically?

Yeah, it also means you can have multiple aliases (if desired) and you can have more than one class per file allowing sensible grouping.

Scanning files is not terribly safe and also means that it's picking up temporary files from my editor.

MBartkowiakSTFC:
Overall, the suggested changes are looking good. Especially with generators being generators, and the clear examples in the docstrings.

It looks promising in the context of issues #446 and #458 so definitely we should go ahead with these changes.

PR #728, opened 2025-03-25
Fix saving paths in IJob.save

### LINKED ISSUE #726: [BUG] Broken scripts when saving from GUI
**Description of the error**
Scripts saved from the GUI are not valid, and fail with
```
$ python run_gromacs_disf.py 
  File "run_gromacs_disf.py", line 23, in <module>
    'trajectory': PosixPath('/complexPR/gromacsTraj1.mdt'),  # trajectory
                  ^^^^^^^^^
NameError: name 'PosixPath' is not defined
```

The GUI should have saved the path as string.

The relevant Path class is never imported in the script, so it will always fail.

**Suggested fix**
Make sure that the code which generates the script convert all the paths to string.

**Additional details**
N/A

oerc0122:
**Description of work**
- Fix saving paths as part of `IJob.save`
- Change docstring to Numpydoc

**Fixes**
`pathlib.Path`s should save correctly as `str`ings in saved files.

could potentially add `from pathlib import Path`, and force paths relative to save directory, but this may lead to further confusion.

Closes #726 

**To test**
Check GUI can save scripts.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -13,29 +13,29 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
-from multiprocessing import Queue
-from logging import FileHandler
-from logging.handlers import QueueHandler, QueueListener
-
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
It fixed the problem, and the code looks better.

PR #735, opened 2025-04-02
Fix PDF calculation for non-orthogonal unit cells

### LINKED ISSUE #686: [BUG] Confusing traceback on invalid file
**Description of the error**
On providing an invalid file to any `IJob`, the traceback refers to an `KeyError` in `selection_length` rather than stating the file doesn't exist. While this is reported in the logs, so MDANSE apparantly knows.

A `FileNotFoundError` raised with a useful message.

<details>
<summary>Log</summary>

```
Could not use xxxx as input trajectory
IJob did not find 'write_logs' in output_files
Job failed with traceback: Traceback (most recent call last):
  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py", line 395, in run
    self.initialize()
  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/VelocityAutoCorrelationFunction.py", line 114, in initialize
KeyError: 'selection_length'

  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py", line 395, in run
    self.initialize()
  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/VelocityAutoCorrelationFunction.py", line 114, in initialize
KeyError: 'selection_length'

During handling of the above exception, another exception occurred:

  File "/home/jacob/MDANSE/MDANSE/Tests/UnitTests/Analysis/tmp.py", line 15, in <module>
    vacf.run(parameters, status=True)
  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py", line 418, in run
    raise JobError(self, tb)
MDANSE.Framework.Jobs.IJob.JobError: Traceback (most recent call last):
  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py", line 395, in run
    self.initialize()
  File "/home/jacob/MDANSE/MDANSE/Src/MDANSE/Framework/Jobs/VelocityAutoCorrelationFunction.py", line 114, in initialize
KeyError: 'selection_length'
```

</details>

**Suggested fix**
If trying to load a necessary file, `raise` early. 

**Additional details**
<details>
 <summary>MWE</summary>

```python
from MDANSE.Framework.Jobs.IJob import IJob

temp_name = "output"

parameters = {
    "frames": (0, 10, 1, 5),
    "interpolation_order": 1,
    "output_files": (temp_name, ("MDAFormat",), "INFO"),
    "running_mode": ("single-core",),
    "normalize": True,
    "trajectory": "xxxx",
}

vacf = IJob.create("VelocityAutoCorrelationFunction")
vacf.run(parameters, status=True)
```

</details>

### LINKED ISSUE #734: [BUG] Wrong PDF counts for non-orthogonal unit cells
**Description of the error**
If a pair distribution function calculation is run on a system with a non-orthogonal unit cell, some counts appear in the low-r range (<0.5 Angstrom), while it is clear that these distances do not appear in the trajectory.

This effect disappears if the atoms close to the edge of the simulation box are excluded, which indicates that the periodic boundary conditions are not handled correctly.

The counts at low r should be 0, and all the interatomic distances should be larger than 1 Angstrom.

Small but significant number of counts at r < 0.5 Angstroms.

**Suggested fix**
Investigate the UnitCell direct and inverse arrays.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Following a recent bug report, this PR changes the way the pair distribution function is calculated for non-orthogonal unit cells.

Additionally, an option of _not_ writing an output file and storing the analysis results in memory is added. It is only usable in scripts and not in the GUI. In this PR, the new PDF unit test is using the in-memory output.

Also, IJob has been extended to check all the configuration inputs on startup. This way all the errors from different configurators are gathered in a single RuntimeError exception, while the GUI can try to configure inputs multiple times without raising anything, same as before.

closes #686
closes #734

**Fixes**
- use normal arrays in UnitCell instead of transposed arrays,
- add FileInMemory output format,
- add check_status method to IJob. It throws an exception if a job is run that had configuration errors.

**To test**
All tests must pass.

The in-memory output file relies on h5py 3.13.0. The version requirement has been added to pyproject.toml, but if you are using an editable install, you may have to update the h5py version manually.


MDANSE/Src/MDANSE/Framework/Jobs/DistanceHistogram.py
```diff
@@ -207,14 +210,17 @@ def run_step(self, index):
...
-        :Parameters:
-            #. index (int): The index of the step.\n
-            #. x (any): The returned result(s) of run_step
-        """
+    def combine(self, _index, x):
+        """Add the results of run_step to the output arrays.
 
+        Parameters
+        ----------
+        index : _type_
```
oerc0122: Missing type?

MDANSE/Src/MDANSE/Framework/Jobs/DistanceHistogram.py
```diff
@@ -156,32 +158,33 @@ def detailed_unit_cell_error(self):
...
+
+        Parameters
+        ----------
+        index : int
+            number of the simulation frame for which to calculate the distances.
 
+        Returns
+        -------
+        int, tuple[float, np.ndarray, np.ndarray]
+            repeated input index and the analysis results
```
oerc0122: Numpydoc spec allows this spec to be:
```suggestion
        int
            Repeated input index.
        tuple[float, ~numpy.ndarray, ~numpy.ndarray]
            The analysis results.
```

MDANSE/Src/MDANSE/Framework/Jobs/DistanceHistogram.py
```diff
@@ -148,6 +149,7 @@ def initialize(self):
         self.indices_intra = find_index_groups(self.indexToMolecule, self.indexToSymbol)
 
     def detailed_unit_cell_error(self):
```
oerc0122: I realise that this isn't introduced in this PR, but since we're cleaning up this file already.

This messes with the trackback for no real benefit. It also doesn't use `self`.
```python
    @staticmethod
    def detailed_unit_cell_error() -> ValueError:
        """Return an error for if a valid unit cell is not present in the trajectory."""
        return ValueError(
            "This analysis job requires a unit cell (simulation box) to be defined. "
            "The box will be used for calculating density in the analysis. "
            "You can add a simulation box to the trajectory using the TrajectoryEditor job. "
            "Be careful adding the simulation box, as the wrong dimensions can render the results meaningless."
        )
```
Then use as:
```python
if should_fail:
    raise self.detailed_unit_cell_error()
```
Probably more sensible still is remove the function all together.
```python
    DETAILED_CELL_MESSAGE = (
        "This analysis job requires a unit cell (simulation box) to be defined. "
        "The box will be used for calculating density in the analysis. "
        "You can add a simulation box to the trajectory using the TrajectoryEditor job. "
        "Be careful adding the simulation box, as the wrong dimensions can render the results meaningless."
    )

    if should_fail:
        raise ValueError(self.DETAILED_CELL_MESSAGE)
```
Which has the benefit of documenting the error raised in-place.

MDANSE/Src/MDANSE/Framework/Configurable.py
```diff
@@ -124,6 +123,41 @@ def configuration(self):
...
+        """
+        errors = {}
+        warnings = {}
+        for name, conf in list(self._configuration.items()):
+            if not conf.valid:
+                if conf.optional:
+                    warnings[name] = conf.error_status
+                else:
+                    errors[name] = conf.error_status
+        if len(warnings):
```
oerc0122:
```suggestion
        if warnings:
```

MDANSE/Src/MDANSE/Framework/Configurable.py
```diff
@@ -124,6 +123,41 @@ def configuration(self):
...
+            LOG.warning(
+                "\n".join(
+                    ["Optional configuration entries were not valid:"]
+                    + [f"{entry}: {error}" for entry, error in warnings.items()]
+                )
+            )
+        if len(errors):
```
oerc0122:
```suggestion
        if errors:
```

MDANSE/Src/MDANSE/Framework/Formats/FileInMemory.py
```diff
@@ -0,0 +1,68 @@
...
+        cls,
+        filename: str,
+        data: dict[str, "IOutputVariable"],
+        header: str = "",
+        run_instance: "IJob" = None,
+        extension: str = extensions[0],
+    ) -> h5py.File:
+        """Write the MDA data structure to an in-memory HDF object.
+
+        Attributes
```
oerc0122:
```suggestion
        Parameters
```

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
```diff
@@ -47,11 +47,13 @@ class HDFFormat(IFormat):
     def write(
         cls,
         filename: Union[Path, str],
-        data: Dict[str, "IOutputVariable"],
+        data: dict[str, "IOutputVariable"],
```
oerc0122, line 50: Given we're >=3.9 and don't have `from __future__ import annotations` this may not work

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -156,6 +157,11 @@ def configuration(self):
     def finalize(self):
         if self._log_filename is not None:
             self.remove_log_file_handler()
+        if hasattr(self._outputData, "data_object"):
+            self._in_memory_result = self._outputData.data_object
```
oerc0122: Up to you, but could also do:
```suggestion
        self._in_memory_result = getattr(self._outputData, "data_object", None)
```
This will assign a value to `self._in_memory_result`.

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -156,6 +157,11 @@ def configuration(self):
...
+
+    def get_results(self):
```
oerc0122: Any reason not to be an `@property`?

MDANSE/Src/MDANSE/Framework/Jobs/VanHoveFunctionSelf.py
```diff
@@ -119,7 +119,7 @@ class VanHoveFunctionSelf(IJob):
     )
     settings["output_files"] = (
         "OutputFilesConfigurator",
-        {"formats": ["MDAFormat", "TextFormat"]},
+        {"formats": ["MDAFormat", "TextFormat", "FileInMemory"]},
```
oerc0122: Do these just want to be defined as default in `OutputFilesConfigurator` and only overridden if different?

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -497,16 +497,10 @@ def read_atomic_trajectory(
...
-                    for nf in range(first, last, step)
-                ]
+                [self.unit_cell(nf).direct for nf in range(first, last, step)]
             )
             inverse_cells = np.array(
-                [
-                    self.unit_cell(nf).transposed_inverse
-                    for nf in range(first, last, step)
-                ]
+                [self.unit_cell(nf).inverse for nf in range(first, last, step)]
```
oerc0122, line 501: How come these are calls (`()`), but below they're getitems (`[]`)? If they can be `[]` do we just want
```python
np.array(self.unit_cell[first:last:step])
```

MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
```diff
@@ -330,7 +330,7 @@ def to_real_coordinates(self, box_coordinates, first, last, step):
             real_coordinates = np.empty(box_coordinates.shape, dtype=np.float64)
             comp = 0
             for i in range(first, last, step):
-                direct_cell = self._unit_cells[i].transposed_direct
+                direct_cell = self._unit_cells[i].direct
                 real_coordinates[comp, :] = np.matmul(
                     direct_cell, box_coordinates[comp, :]
                 )
```
oerc0122: `@` is a shortcut for `matmul`

> The matmul function implements the semantics of the @ operator introduced in Python 3.5 following [PEP 465](https://peps.python.org/pep-0465/).

```suggestion
                real_coordinates[comp, :] = direct_cell @ box_coordinates[comp, :]
```

MDANSE/Src/MDANSE/Framework/Configurators/OutputFilesConfigurator.py
```diff
@@ -15,38 +15,46 @@
...
     resulting from an analysis.
 
-    Once configured, this configurator will provide a list of files built by joining the given output directory, the
+    Once configured, this configurator will provide a list of files
+    built by joining the given output directory, the
     basename and the extensions corresponding to the input file formats.
 
-    For analysis, MDANSE currently supports only the HDF and Text formats. To define a new output file format
-    for an analysis, you must inherit from MDANSE.Framework.Formats.IFormat.IFormat interface.
+    For analysis, MDANSE currently supports only the HDF and Text formats.
```
oerc0122: No longer true with this addition.

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
MBartkowiakSTFC: I am pretty sure this should already work in 3.9, following this
https://docs.python.org/3.9/library/typing.html#corresponding-to-built-in-types

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
MBartkowiakSTFC: H5MDTrajectory is a wrapper class which is supposed to make an H5MD file look like an .MDT file. The idea behind a `unit_cell(self, index)` method was to make it easier to change the code if we ever decide to change the way we construct an MDANSE UnitCell out of the data in the H5MD file.

Having said that, it is entirely possible that this abstraction is no longer necessary.

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
oerc0122: You are correct, we should update these globally. What's not allowed until 3.10 is the `|` Unions.

MDANSE/Src/MDANSE/Framework/Jobs/VanHoveFunctionDistinct.py
```diff
@@ -411,8 +404,8 @@ def run_step(self, time: int) -> tuple[int, tuple[np.ndarray, np.ndarray]]:
                 frame_index_t1
             )
             coords_t1 = conf_t1["coordinates"][self._indices]
-            direct_cell = conf_t1.unit_cell.transposed_direct
-            inverse_cell = conf_t1.unit_cell.transposed_inverse
+            direct_cell = conf_t1.unit_cell.direct
+            inverse_cell = conf_t1.unit_cell.inverse
 
             scaleconfig_t0 = coords_t0 @ inverse_cell
             scaleconfig_t1 = coords_t1 @ inverse_cell
```
ChiCheng45, line 412: I'm a bit confused about this. In UnitCell.py we have 

https://github.com/ISISNeutronMuon/MDANSE/blob/270c6d71d2976ec86057f793dc029a50353bdd5e/MDANSE/Src/MDANSE/MolecularDynamics/UnitCell.py#L52-L79

so the direct matrix is

```math
\begin{bmatrix} 
a_x & a_y & a_z \\
b_x & b_y & b_z \\
c_x & c_y & c_z
\end{bmatrix}
```

where $a_x$, $a_y$, and $a_z$ are the $x$, $y$, and $z$ coordinates of the unit cell vector $a$. If this is true, then the original code with the transpose should be correct. Could it be that in #734, the trajectory conversion did not correctly store the direct matrix following the definition above?

ChiCheng45: I just had a think and 

```
inverse_cell = conf_t1.unit_cell.inverse
scaleconfig_t0 = coords_t0 @ inverse_cell
scaleconfig_t1 = coords_t1 @ inverse_cell
```

would be the correct way to do it. The confusion comes from all the transpositions we are doing. Anyway, in textbooks we'd usually write something like this

```math
\begin{bmatrix} 
r_x  & \\
r_y  & \cdots \\
r_z  &
\end{bmatrix} = \begin{bmatrix} 
a_x & b_x & c_x \\
a_y & b_y & c_y \\
a_z & b_z & c_z
\end{bmatrix}
\begin{bmatrix} 
\rho_a   &  \\
\rho_b   &  \cdots \\
\rho_c &
\end{bmatrix}
```
to go from fractional to Cartesian. In our numpy arrays, we have them ordered with `x, y, z` in each row. So we need to transpose both sides of the equation.

```math
\begin{bmatrix} 
r_x  &
r_y  &
r_z  \\
 &
\vdots &
\end{bmatrix} = \begin{bmatrix} 
\rho_a  &
\rho_b  &
\rho_c \\
 &
\vdots &

\end{bmatrix} \begin{bmatrix} 
a_x & a_y & a_z \\
b_x & b_y & b_z \\
c_x & c_y & c_z
\end{bmatrix}
```

I think we need to double-check that we use the bottom equation everywhere where the transformation matrix here is direct in MDANSE.

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -497,16 +497,10 @@ def read_atomic_trajectory(
...
             )
             atomic_traj = atomic_trajectory(
                 coords, direct_cells, inverse_cells, box_coordinates
```
ChiCheng45, line 504: I think we also need to update atomic_trajectory since that function expects the transpose versions of direct and inverse.

ChiCheng45:
I've updated to code so that transpose_inverse and transpose_direct are never used, these properties are removed. This should make the code less confusing. No more of this nonsense `direct = unit_cell.transpose_direct`. Basically, from now onwards, anyone who develops on MDANSE should just use the following.

`cartesian_coordinates = fractional_coordinates @ direct`
`fractional_coordinates = cartesian_coordinates @ inverse`

New developers will only see this, so now if they just copy and paste code, they will always get the correct one.

ChiCheng45:
I've just created an argon trajectory with the same particle density but with different unit cells. I used LAMMPS to do this, and I converted the trajectory using #736 as the current LAMMPS parser doesn't work with triclinic unit cells.

image
image

The results were similar for a bunch of calculations, slight differences occurred as I didn't use a massive trajectory. Apparently, the trajectory with the non-orthogonal cell was 1.5K hotter for some reason. The PDF distance cutoff needs fixing for the non-orthogonal unit cell.

image

image

image

image

image

image

image

image

image


Here is the PDF calculation from protos for comparison.

image

ChiCheng45:
I've corrected the distance cutoff for non-orthogonal unit cells.

image

ChiCheng45:
Previously I used NVE, I changed to NVT and the temperatures are the same now.

image

MSD are now much more similar, again differences will be the short trajectory I'm using.

image

DISF
image

SSF
image

oerc0122 (approved):
Looks reasonable to me if the answers check out.

PR #736, opened 2025-04-03
Rework lammps parser

### LINKED ISSUE #720: [BUG] LAMMPS converter does not support ix, iy, iz fields
**Description of the error**
At the moment, the LAMMPS converter can recognise some fields in the trajectory file ('x', 'xs') but fails for 'xsu' and 'ix'.

It should either read the files correctly, or fail with an error.

It reads the coordinates, but does not unwrap them, leading to wrong results.

**Suggested fix**
Try to support all the possible cases in the converter, or find another converter that works better.

**Additional details**
N/A

oerc0122:
**Description of work**
Rework lammps parser to be more effective and useful.

**Fixes**
- Now parses all cell types
- Handles different coordinate types
- Extracts as much information as useful from Config
- Makes some things optional.
- Uses image indices to correctly position atoms (Fixes #720)
- Shifts atoms relative to the cell origin
- Fix for VMD style dumps
- Fix for ASE style dumps

**To test**
- New tests added for unwrapping coordinates
- Updated tests to reflect image shifts.
- Add new test of lammps_CaOMgOSiOSiO2FeO system


MDANSE/Src/MDANSE/Framework/Converters/LAMMPS.py
```diff
@@ -13,319 +13,510 @@
...
                     try:
-                        temp_index = int(temp[0])
-                    except ValueError:
-                        self._rankToName[i] = name
-                    else:
-                        self._rankToName[temp_index - 1] = name
+                        atom_type = temp.get("type")
+                        if atom_type is None:
+                            atom_type = config["atom_types"][i]
+                        atom_type -= 1
```
MBartkowiakSTFC: It seems to me that, for the trajectory without the atom ID column (with element names in the trajectory file), the conversion will only work if we have
```
else:
    atom_type -= 1
```

MBartkowiakSTFC (approved):
I tried it out on all the trajectories I could find. It worked for all of them, and recognised the non-orthogonal unit cells correctly. Looks good.

PR #738, opened 2025-04-04
Split distinct van Hove calculation into segments

### LINKED ISSUE #776: [ENHANCEMENT] Remove intra and intermolecular results if the system does not contain molecules
In the PDF, SSF, the distinct part of the van Hove function, the results are split into inter- and intra-molecular components, even if there are no molecules in the system. Currently, if there are no molecules, then the intramolecular components are zero, so that the intermolecular and total results will be the same.

Only the total result should be saved if there aren't no molecules in the system.

MBartkowiakSTFC:
**Description of work**
This PR is an extension of #735 
It should improve the memory efficiency of PDF calculations. You are welcome to suggest improvements to speed up the code.

The correction to the coordinate unfolding should bring a small improvement to the calculation accuracy.

The performance of this code is still likely to be an issue. The last part of the `van_hove_distinct` function which sorts the distances into the `intra` and `inter` parts is pure Python, and could be improved.

This PR should prevent the memory requirements of PDF and distinct van Hove calculations from increasing indefinitely (and quadratically) with the number of atoms in the system.

**Fixes**
1. For calculations of van Hove function, the coordinates at t0, t1, are folded using the unit cell dimensions from t0, t1 respectively. (Before it was using cell from t0 for both coordinate arrays.
2. The `van_hove_distinct` function now has a `size_limit = 1024` optional keyword argument. If the coordinate array is larger that size_limit, it will be split into smaller arrays to prevent the creation of large `N^2` distance arrays.
3. The distances are calculated using `scipy.spatial.KDTree`.
4. Closes #776

**To test**
All tests must pass.
Try running a PDF calculation for a very large trajectory; it should not allocate multiple gigabytes of RAM per process.


ChiCheng45:
Great I think the idea seems good. I will do some performance testing soon.

Just one thing, now that you want to try to fix things for NPT. We need to think about the how the density is handled with PDF and the van Hove functions.

image

Note that the equations assume that the density is fixed which is not the case with NPT.

MBartkowiakSTFC:
This branch, apart from introducing an upper limit on memory requirements of the calculation, adds also a small speedup. Our unit tests in Analysis/test_dynamics.py::test_dynamics_analysis for VanHoveDistinct had the following total run time on my computer:

- current version (protos): 86.7 s
- this PR: 76 s

ChiCheng45:
I have tested this a few times, and it looked good. I think it's probably as optimised as it can be with numpy. Also, the idea looks good and makes sense. Can put me down to put a review in if it's ready.

ChiCheng45:
- I've found an issue that the inter- and intramolecular parts are not split up correctly.

Here is the plot using at DFTB trajectory.

image

Here is the same on protos.
image

- I also find that the calculations are much slower than on protos for slightly larger systems. I'm using an argon trajectory with 256 atoms and 1001 frames and its taking over 2 hrs for the van hove function calculation with default parameters. Did you see the same? Here are some timings for the PDF calculation with 256 atoms and 1001 frames. Protos: 7s. This branch: 1min 5s.

MBartkowiakSTFC:
@ChiCheng45 You were right about the poor performance. The current version should produce correct results. I expect it to be up to 20% slower than the protos version, but it will be able to deal with large systems without running out of RAM. Also, the jobs based on DistanceHistogram should only try to split the results into the "inter" and "intra" parts if molecules are defined.

Also, I added tests to test_structure.py that run on a trajectory with molecules, as this has not been the case so far.

ChiCheng45 (approved):
Ace, this is much better now. I got a massive speedup for this 16384-atom 1 frame trajectory, protos 12 mins vs this branch 38s. On protos, it was using all my RAM and started writing to disk, causing my whole computer to slow down. Results are the same as protos. I think this is good to go in.

PR #740, opened 2025-04-04
Extend MDANSE to support complex neutron scattering lengths

### LINKED ISSUE #718: [BUG] MDANSE does not work with complex scattering lengths
**Description of the error**
In the atoms database, scattering lengths with complex values have been given values of zero; see, for example, Lithium-6. 

We need to fix MDANSE so that complex scattering lengths can be used.

MBartkowiakSTFC:
**Description of work**
Users reported problems trying to analyse MD simulations of systems containing 6Li. This PR is meant to enable calculations on atoms with a non-negligible imaginary component of the scattering length.

closes #718

**Fixes**
1. Added `complex` to supported variable types in atom database.
2. Replaced missing scattering lengths with values from `periodictable` where possible, and from the NIST table in other cases (typically for the incoherent scattering length).
3. Removed the `.real` part from the correlation calculations.
4. Changed the unit of scattering length to fm.
5. Added unit information to the atom database.
6. Modified the code of the `get_weights` function to use the complex conjugate when multiplying b_i*b_j.
7. Removed b_incoherent2 from the database.
8. Updated the TrajectoryWriter to write the new scattering lengths correctly.
9. Extended the plotter, which now separates a dataset into two parts (real and imaginary) if it detects a non-zero imaginary component in it.
10. Renamed the atomic mass unit from "uma" to "Da".
11. Switched from `rfft` to `fft` in the jobs that calculate spectra (PPS, CCF, DOS, IR).
12. Modified the `compare_hdf5` function in the tests to compare only the real part of the results.
13. Performance improvements for reading atom properties from trajectory (by caching).

**Note:** while FFT is calculated instead of RFFT, the negative side of the results is still discarded in the code.

**To test**
All tests must pass.
Convert a trajectory using the protos branch. Convert the same trajectory using this branch. Run DCSF, DISF and DOS analysis for both trajectories. The results should be plottable and nearly identical.


MDANSE/Src/MDANSE/Chemistry/atoms.json
```diff
@@ -1348,19 +1357,18 @@
...
             "ionization_energy": 2932.98564,
             "xs_absorption": 2.1e-09,
             "xs_scattering": 0.0,
-            "b_incoherent": 0.0,
+            "b_incoherent": "0.0",
             "electronegativity": 2.6,
             "group": 18,
             "electron_affinity": 0.0,
-            "b_coherent": 0.0,
+            "b_coherent": "(nan-0.005839822024471635j)",
```
ChiCheng45, line 1371: Is this supposed to be nan?

ChiCheng45: There are quite a few nan and also some None values in dotted across the atoms.json for the scattering lengths

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -13,15 +13,104 @@
...
+            dtype=complex,
+        )
+
+        factor = atom_conc_product**conc_exp * property_product
+        # E.g. for property b_coh, 5 Cu atoms, 100 total atoms, and dim=2
+        # factor = (5*5/(100*100))**conc_exp * b_coh(Cu)*b_coh(Cu)
+
+        weights[elements] = np.copy(factor)
+        norm_factor += atom_conc_product * property_product
+    return weights, norm_factor
```
ChiCheng45: Do we have to have the ND version if it will never be used?

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -13,15 +13,104 @@
...
+    for el1, el2 in cartesianProduct:
+        concentration = contents[el1] * contents[el2] / n_atms**2
+        if isinstance(props[el1], np.ndarray):
+            prop1 = np.conjugate(props[el1].astype(complex))
+        else:
+            prop1 = complex(props[el1]).conjugate()
+        if isinstance(props[el2], np.ndarray):
+            prop2 = props[el2].astype(complex)
+        else:
+            prop2 = complex(props[el2])
```
ChiCheng45, line 81: Just curious, when will the properties be a instance of ndarray?

MDANSE/Src/MDANSE/Chemistry/atoms.json
MBartkowiakSTFC: I agree that we should aim for more consistency when it comes to "nan" and "None". In any case, these elements and isotopes really are missing scattering length values in the tables (both the NIST table and periodictable package).

`periodictable` seems to give the imaginary part of b that has been recalculated from the absorption cross section. I tried to replace all our b values with those from `periodictable`. Then I compared it to the NIST tables, and replaced b_incoherent with the NIST values for those elements where NIST explicitly includes the imaginary component in b_incoherent. Some b values were not provided by either of them, and that's where we have the odd entries.

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
MBartkowiakSTFC: This happens for the X-ray Static Structure Factor.

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
MBartkowiakSTFC: I'm happy to remove it.

ChiCheng45: Cool, my concern was that ND with dim=2 wasn't the same as 2D because ND doesn't have the conjugation. Also it isn't clear how the conjugation should be dealt with for 3D and etc. Maybe we should figure this out when something needs 3D or higher dimensions.

MDANSE/Src/MDANSE/Framework/Jobs/DensityOfStates.py
```diff
@@ -175,6 +181,7 @@ def initialize(self):
                 (instrResolution["n_romegas"],),
                 axis="romega",
                 units="au",
+                dtype=np.complex128,
```
ChiCheng45: I don't think we need to store the DOS (also PPS and IR) as complex.

The DOS is a FT of a symmetrised version of the VACF signal so that the VACF is both real and symmetric. The DOS will be a FT of a real and symmetric function so that it will always be real. The only time it wouldn't be real would be if we calculated the VACF for times -t/2 to t/2 instead of calculating it from 0 to t and symmetrising to get a VACF from -t to t.

Or maybe you storing them as complex here because we are allowing the weights to take complex values?

MDANSE/Src/MDANSE/Chemistry/atoms.json
ChiCheng45: Right I see. I think we should do something about them though.

If I use scattering length with None values then the jobs fail. If a nan value exist then the jobs continue to run and saves an output plot there are no results in the plot as the values are nan.

ChiCheng45:
We will need to update the atom database editor as you can now enter any string for their scattering lengths.

Screenshot from 2025-05-15 17-26-41

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
```diff
@@ -321,7 +327,7 @@ def combine(self, index: int, x: np.ndarray):
                 corr = correlate(x[pair[0]], x[pair[1]][:n_configs], mode="valid").T[
                     0
                 ] / (n_configs * x[pair[0]].shape[1])
-                self._outputData[f"f(q,t)_{pair_str}"][index, :] += corr.real
+                self._outputData[f"f(q,t)_{pair_str}"][index, :] += corr
```
ChiCheng45: In protos we have

```math
F_{AB}(q,t) = \frac{1}{N\sqrt{c_{A}c_{B}}}(\beta_{A}\beta_{B} +  \beta_{B}\beta_{A})C_{AB}(q,t)
```

where

```math
C_{AB}(q,t) = \sum_{j \in A}\sum_{k \in B} \langle  \exp[-iq \cdot r_j(0)] \exp[iq \cdot r_k(t)]\rangle 
```

This branch now has

```math
F_{AB}(q,t) = \frac{1}{N\sqrt{c_{A}c_{B}}} (\beta_{A}^{*}\beta_{B} +  \beta_{B}^{*}\beta_{A}) C_{AB}(q,t)
```

I wonder if we should change it so that we are using 

```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)+ \beta_{B}^{*}\beta_{A}C_{BA}(q,t) \right]
```

so that we avoid making any assumptions about $C_{AB}(q,t) = C_{BA}(q,t)$ and etc. Also, I think the thing in the bracket should be real so that we don't need to save the results as a complex number in the h5py file. Also similar comments for the CCF.

MBartkowiakSTFC: Yes, I had some doubt about this when preparing this PR. I think you are right.

ChiCheng45: I've had a look to see how $C_{AB}(q,t)$ and $C_{BA}(q,t)$ are related. In the theory of simple liquids, equation 7.1.9 for a generation correlation function $C_{AB}(t)$

```math
```
where $\epsilon_{A}$ and $\epsilon_{B}$ are the time-reversal signatures of $A$ and can be +1 or -1. So that for our case

```math
C_{AB}(q,t) = C_{AB}(q,-t) = C_{BA}^{*}(q,t)
```

since $A$ and $B$ will have the same sign for their time-reversal signatures and therefore

```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)+ \beta_{B}^{*}\beta_{A}C_{AB}^{*}(q,t) \right]
```

```math
F_{AB}(q,t) =  \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left[\beta_{A}^{*}\beta_{B} C_{AB}(q,t) \right]
```

In other words, it looks like the results for DCSF and also for the CCF should be real, and we need not store the complex part since it should be small. I think it still might be preferable to calculate

```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)+ \beta_{B}^{*}\beta_{A}C_{BA}(q,t) \right]
```

since you might get slightly better statistics for not much more computational effort.

MBartkowiakSTFC:
@ChiCheng45 I added a ComplexValidator to the atom property table, and replaced all the `None` values with `nan`. If #780 gets merged, we could try to add at least a warning in the GUI if the atoms in the trajectory have unknown scattering lengths.

The output data are `np.float64` again. Also, `assign_weights` now takes a complex conjugate of every other pair of atom types, which I thought would be an implementation of the equations you presented above, giving us a $2Re[(\beta_A^{*}\beta_B)]$ prefactor for $F_{AB}$ but please have another look and check if you agree with it.

MDANSE/Src/MDANSE/Framework/Jobs/DensityOfStates.py
```diff
@@ -255,17 +261,21 @@ def finalize(self):
...
+            )[-self._n_omegas :]
             if self.add_ideal_results:
                 self._outputData[f"dos_ideal_{element}"][:] = get_spectrum(
                     self._outputData[f"vacf_{element}"],
                     None,
                     self.configuration["instrument_resolution"]["time_step"],
-                    fft="rfft",
-                )
+                    fft="fft",
+                )[-self._n_omegas :]
```
ChiCheng45: Should be able to revert this and use rfft since vacf is real.

MDANSE/Src/MDANSE/Framework/Jobs/Infrared.py
```diff
@@ -221,15 +224,15 @@ def finalize(self):
...
+        )[-self._n_omegas :]
         if self.add_ideal_results:
             self._outputData["ir_ideal"][:] = get_spectrum(
                 self._outputData["ddacf"],
                 None,
                 self.configuration["instrument_resolution"]["time_step"],
-                fft="rfft",
-            )
+                fft="fft",
+            )[-self._n_omegas :]
```
ChiCheng45: Same as previous comments.

MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
```diff
@@ -390,42 +400,42 @@ def finalize(self):
...
+            bi: complex = self.configuration["trajectory"][
+                "instance"
+            ].get_atom_property(element, "b_incoherent")
             self._outputData[f"f(q,t)_inc_{element}"].scaling_factor *= (
-                bi * number * norm_natoms
+                bi.conjugate() * bi * number * norm_natoms
             )
             self._outputData[f"s(q,f)_inc_{element}"].scaling_factor *= (
-                bi * number * norm_natoms
+                bi.conjugate() * bi * number * norm_natoms
```
ChiCheng45, line 438: I think we need to cast `bi.conjugate() * bi * number * norm_natoms` to real.

MDANSE/Src/MDANSE/Framework/Jobs/PositionPowerSpectrum.py
```diff
@@ -234,15 +238,15 @@ def finalize(self):
...
+            )[-self._n_omegas :]
             if self.add_ideal_results:
                 self._outputData[f"pps_ideal_{element}"][:] = get_spectrum(
                     self._outputData[f"pacf_{element}"],
                     None,
                     self.configuration["instrument_resolution"]["time_step"],
-                    fft="rfft",
-                )
+                    fft="fft",
+                )[-self._n_omegas :]
```
ChiCheng45: See previous comments.

ChiCheng45:
> The output data are `np.float64` again. Also, `assign_weights` now takes a complex conjugate of every other pair of atom types, which I thought would be an implementation of the equations you presented above, giving us a 2 R e [ ( β A ∗ β B ) ] prefactor for F A B but please have another look and check if you agree with it.

Not exactly, the equations were 

```math
F_{AB}(q,t) =  \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left[\beta_{A}^{*}\beta_{B} C_{AB}(q,t) \right]
```

I'm not sure that the imaginary part of $C_{AB}(q,t)$  is always small.  $C_{AB}(q,t) + C_{BA}(q,t)$ on the other hand should be real.

MDANSE/Src/MDANSE/Framework/Jobs/CurrentCorrelationFunction.py
```diff
@@ -502,30 +513,30 @@ def finalize(self):
...
                     None,
                     self.configuration["instrument_resolution"]["time_step"],
                     axis=1,
-                    fft="rfft",
-                )
+                    fft="fft",
+                )[:, -self._nOmegas :]
 
         weights = self.configuration["weights"].get_weights()
         weight_dict = get_weights(weights, nAtomsPerElement, 2, conc_exp=0.5)
```
ChiCheng45: I think we need to update correct the CCF in the same way that the DCSF needs be corrected for imaginary weights in case someone uses them with CCF.

ChiCheng45:
> The output data are `np.float64` again. Also, `assign_weights` now takes a complex conjugate of every other pair of atom types, which I thought would be an implementation of the equations you presented above, giving us a 2 R e [ ( β A ∗ β B ) ] prefactor for F A B but please have another look and check if you agree with it.


I had another think about this, and I think this should be fine as long as specific q-vector generators are used. For example, if we used the spherical q-vectors with sufficient sampling, then in our averaging, we should have something like this,

```math
C_{AB}(\vert q \vert, t) = \frac{1}{N_{q}} [\cdots + C_{AB}(-q, t) +  C_{AB}(q, t) + \cdots ]
```

since 

```math
C_{AB}(-q, t) = C_{AB}^{*}(q, t) = C_{BA}(q, t)
```

$C_{AB}(\vert q \vert, t)$ is real and 

```math
F_{AB}(\vert q \vert,t) =  \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left[\beta_{A}^{*}\beta_{B} C_{AB}(\vert q \vert,t) \right] = \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left[\beta_{A}^{*}\beta_{B} \right] C_{AB}(\vert q \vert,t) 
```

The only problem is that the above equation might not be true for other q-vector generation schemes since the sampling may not have generated $q$ and $-q$.

ChiCheng45:
I did some more thinking. This is related to the above comment. Since,

```math
2 \mathrm{Re} [C_{AB}(q, t)] = C_{AB}(q, t) + C_{AB}(-q, t)
```

so in this branch, we do the following,

```math
 \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} \right\}  \mathrm{Re} \left\{ C_{AB}(q,t) \right\}=   \frac{1}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} [C_{AB}(q,t) + C_{AB}(-q,t) ]  \right\}  = \frac{1}{2}[F_{AB}(q,t) + F_{AB}(-q,t)]
```

in other words when we use any q-vector $q$ we actually average over the values of $q$ and $-q$. I guess there are a few things we could do, update the spherical lattice q-vectors so it always includes $q$ and $-q$, update it so it only includes $q$ or just leave it and rely on the sampling to generate enough $q$ and $-q$.

MBartkowiakSTFC:
This has also been solved by #963 instead.

PR #751, opened 2025-04-22
Fix issue when MultiInput... has non-extant files

oerc0122:
**Description of work**
`raise` fails due to passing `Path` to `str.join` which requires `str`ing.

**Fixes**
- Forces `Path`s to `str`ings.
- Simplify loop to comprehension with filter.

**To test**
Supply non-extant files to a multi-input configurator.


MBartkowiakSTFC (approved):
Looks good (apart from the linter complaining about one line). I tried it out, and I got the right error message (file does not exist), which I was not getting before.

PR #620, opened 2024-12-10
Molecule preview button and fast plotting

raunakkbanerjee:
**Description of work**

Added molecule preview button which, on clicking, creates a window with a 3d view of the selected molecule.
Added fast plotting functionality, on double clicking the file under plot creator tab creates a new plot under plot holder tab. The dataset with main tag is plotted with solid line and dataset with partial and main tag is plotted with dashed line.

**To test**

Molecule preview can tested by selecting a job like angular correlation under the actions tab after using a trajectory with multiple molecules
Fast plotting can be tested by selecting a .mda file for plotting under the plot creator file and double clicking the file creates a new plot under plot holder tab.

PR #756, opened 2025-04-24
Molecule preview using rdkit

MBartkowiakSTFC:
**Description of work**
An alternative version of PR #620, using rdkit for drawing molecules instead of Qt3D.

PR #620 works on most platforms, but has been found to cause crashes on modern MacOS combined with older Python interpreters. This seems to be the problem with the external libraries or Python interpreter itself, since for Python 3.13 the problem disappears.

In the future we should be able to use Qt3D, but for now this PR is meant to offer a working solution.

**Fixes**
1. Adds a molecule preview dialog,
2. Enables fast plotting of results based on in-file tags ("main", "partial").

**To test**
1. Load two trajectories, with and without molecules. Pick AngularCorrelation analysis and try to visualise different molecules. Check if the molecule view button becomes inactive if no molecules are present.
3. Load any recent analysis results, preferably one with partial results in 1D (e.g. Density of States). In the plot creator tab, double-click the file name. A plot should be created instantly with the breakdown of the results.


ChiCheng45 (commented):
I've tried running this and after I load the DFTB water trajectory and click the molecule preview on the angular correlation job, MDANSE hangs and then crashes.

@MBartkowiakSTFC Do you get this issue on Mac?

MBartkowiakSTFC:
Screenshot 2025-04-30 at 14 03 18
I get this window, which looks correct to me.

MBartkowiakSTFC:
I was getting the crash with the other branch, from #620.

ChiCheng45:
> Screenshot 2025-04-30 at 14 03 18 I get this window, which looks correct to me.

Yeah not sure what's going on, works fine on my Ubuntu VM but crashes on Windows. Maybe an issue with rdkit?

Animation72

MBartkowiakSTFC:
Maybe try upgrading rdkit on your Windows machine? Also, what Python version are you using?

ChiCheng45:
> Maybe try upgrading rdkit on your Windows machine? Also, what Python version are you using?

Yea I'm going to try some different stuff out to figure out what it might be. 

The version from above is Python 3.12.4 and rdkit 2025.03.2.
I also find a crash with Python 3.11.11 and rdkit 2024.09.6.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculePreviewWidget.py
```diff
@@ -0,0 +1,141 @@
...
+        allchem.Compute2DCoords(submolecule)
+        return submolecule
+
+    def show_formula(self):
+        """Draw the molecule."""
+        submolecule = self.prepare_rdkit_molecule()
+        draw_options = draw.MolDrawOptions()
+        draw_options.addAtomIndices = True
+        pil_image = draw.MolToImage(submolecule, size=(600, 600), options=draw_options)
+        qt_image = ImageQt(pil_image)
```
ChiCheng45:
```suggestion
        qt_image = ImageQt(pil_image).copy()
```
This fixes the crash on Windows.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculePreviewWidget.py
```diff
@@ -0,0 +1,141 @@
...
+        draw_options.addAtomIndices = True
+        driver.SetDrawOptions(draw_options)
+        draw.rdMolDraw2D.PrepareAndDrawMolecule(
+            driver,
+            submolecule,
+            highlightAtoms=selected_atoms,
+            highlightBonds=optional_bonds,
+        )
+        png_data = driver.GetDrawingText()
+        image = QImage.fromData(png_data)
```
ChiCheng45:
```suggestion
        image = QImage.fromData(png_data).copy()
```
Suggested change since the copy was required to fix the crash above.

MBartkowiakSTFC:
Thanks for the corrections. On my machine it still works the same after the changes.

ChiCheng45:
Great, everything seems to work fine. I've tested this on Windows and Ubuntu. 

I had a suggestion in PR #620 (https://github.com/ISISNeutronMuon/MDANSE/pull/620#pullrequestreview-2508546857) that maybe the fast plotting should be done by right-clicking the file for plotting the main results. I think you could right-click the individual results to plot them as well. The double click is a bit clunky since it also opens the drop-down. I'm happy to approve this PR and put my suggestions as an enhancement request if you prefer.

MBartkowiakSTFC:
I would prefer to merge it now and add improvements later. (I saw some of the users work with the GUI and they used double click a lot, expecting something to happen.)

ChiCheng45 (commented):
Just one more thing. I've found that for the static structure factor, pair distribution function and both van Hove function results, double-clicking the file makes an empty plot.

MBartkowiakSTFC:
I added a few changes:
1. PDF, SSF and both van Hove calculations now have the fast plotting tags included in the output data. The files produced with the current version will produce output with fast plotting.
2. Double click on the file name in tree view will not show the child nodes. It is still possible to unfold the tree structure by clicking on the bullet/triangle symbol left of the name.

ChiCheng45:
I've gone through all the jobs, and this should be the rest of the results that don't have a working double click.

RootMeanSquareDeviation
CoordinationNumber
XrayStructureFactor

MBartkowiakSTFC:
Thanks for pointing these out. They were all failing for different reasons.

They should be fine now.

ChiCheng45 (approved):
Looks good to me.

PR #758, opened 2025-04-25
Re-use the job instance in Action.py

MBartkowiakSTFC:
**Description of work**
This is meant to reduce the amount of times we re-load the trajectory in the GUI.

**Fixes**
1. Only re-create the job instance in the GUI if the job type OR trajectory file have changed.
2. Only re-load the trajectory in the trajectory configurator if the path/filename do not match the current one.

**To test**
Load a big trajectory with many atoms and molecules. Pick a job, and start editing the input fields. While the initial setting up of the job panel is still likely to be slow, editing the individual input fields should be faster.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
@@ -138,6 +138,7 @@ class Action(QWidget):
         self._trajectory_configurator = None
         self._settings = None
         self._job_name = None
+        self._job_instance = IJob()
         self._use_preview = use_preview
         self._current_instrument = None
         self._has_been_initialised = False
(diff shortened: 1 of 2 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
```diff
@@ -199,26 +204,32 @@ def update_panel(self, job_name: str) -> None:
...
-        job_instance.build_configuration()
-        settings = job_instance.settings
-        self._job_instance = job_instance
+            self._job_name = job_name
+            if self._default_path is None or Path(self._default_path).samefile(Path()):
+                self._default_path = str(Path(self._parent_tab.get_path(job_name)))
+            try:
+                job_instance = IJob.create(job_name)
+            except ValueError as e:
+                LOG.debug(
```
oerc0122, line 222: This probably wants to be a `LOG.error` if it's a failure.

oerc0122 (approved):
Looks good if it reduces load times (which it seemed to). 

The comment is about extant code so can be ignored for this PR if we don't want to add too many changes.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
MBartkowiakSTFC: I think that we would need to make changes on the other end, i.e. ensure that whatever sends us the `job_name` can only send a valid name that is expected to work, before we make it an error. So, I would prefer not to make changes here right now.

PR #761, opened 2025-04-29
Add data normalisation to the plotter

MBartkowiakSTFC:
**Description of work**
As a temporary measure (before a more complete reworking of the workflow as suggested in #550) this PR implements a basic interface of normalising the curves in the plotter.

**Fixes**
1. Added an extra line of widgets in the plotting view.
2. Single, Grid and Heatmap plots can normalise different rows of data by dividing the entire row by a selected subset of values.

**To test**
1. Load analysis results with 1D curves. Plot several curves together in Single mode. Switch normalisation on and off. The curves should be re-scaled when normalisation is applied, and then return to original values when normalisation is switched off.
2. Load analysis results with a 2D array. Curves should be normalised the same as in point 1. In Heatmap mode, data should be normalised per row, which means that the relative intensity of the rows will change.
3. Load analysis results with a 3D array (molecular trace). Check the normalisation also works for Heatmap for a 3D array.


ChiCheng45:
Looks great, I'm still testing it. But I think with this change, we can now delete the normalize options from the VACF, GACF, and PACF?

MBartkowiakSTFC:
> Looks great, I'm still testing it. But I think with this change, we can now delete the normalize options from the VACF, GACF, and PACF?

I don't really see any good reason to keep the "normalize" option. But will we need to enable this (plotter-based) normalisation also for the text output?

ChiCheng45:
> > Looks great, I'm still testing it. But I think with this change, we can now delete the normalize options from the VACF, GACF, and PACF?
> 
> I don't really see any good reason to keep the "normalize" option. But will we need to enable this (plotter-based) normalisation also for the text output?

Ah yes I guess maybe. I think it is fine for the text output just to save the raw results. If they want to normalize it, it should be easy enough for them to do it. Maybe we should have an option to save plot results to text after the user has manipulated the data, this could be a different PR.

MBartkowiakSTFC:
For the moment, I followed the suggestion and removed "normalize" from VACF, GACF and PACF.

ChiCheng45:
Looks good. Another thing I noticed is that the indices the normalization uses allow negative values, but it seems a bit odd.

Animation73

Here you can see a big jump from (0 to 50) to (-1 to 50) with sum.

MBartkowiakSTFC:
> Looks good. Another thing I noticed is that the indices the normalization uses allow negative values, but it seems a bit odd.
> 
> Here you can see a big jump from (0 to 50) to (-1 to 50) with sum.

Well, these are Python slice indices. I thought it would be good to have an easy way of selecting the last point of the curve using the -1 index. But I guess that strange things will happen if the users mix positive and negative indices, as it won't be clear what really happened.

Should we give users more information about which indices were used to normalise the curve? Or block negative indices for now?

ChiCheng45:
> > Looks good. Another thing I noticed is that the indices the normalization uses allow negative values, but it seems a bit odd.
> > Here you can see a big jump from (0 to 50) to (-1 to 50) with sum.
> 
> Well, these are Python slice indices. I thought it would be good to have an easy way of selecting the last point of the curve using the -1 index. But I guess that strange things will happen if the users mix positive and negative indices, as it won't be clear what really happened.
> 
> Should we give users more information about which indices were used to normalise the curve? Or block negative indices for now?

I think the easiest would be to block negative indices, and also the from index should also be blocked from being larger than to index.

MBartkowiakSTFC:
Two changes:
1. The spin boxes are now locked. If their values are N, M, then the allowed range for the lower one is [0, M-1], and for the higher one it is [N+1, N_POINTS-1].
2. If normalisation cannot be performed (e.g. scaling factor is 0), the normalisation widgets will change colour to indicate an error, and will show more error details in the tooltip. This is similar to the way the job input widgets highlight errors.

ChiCheng45 (approved):
Looks good. Seems to work well.

PR #762, opened 2025-04-29
Update os env setting to make numpy use one thread when the GUI is used

ChiCheng45:
**Description of work**
Update the os envs setting in the GUI main.py. The following environment variables are set.

```
os.environ["OMP_NUM_THREADS"] = "1"
os.environ["OPENBLAS_NUM_THREADS"] = "1"
os.environ["MKL_NUM_THREADS"] = "1"
os.environ["VECLIB_MAXIMUM_THREADS"] = "1"
os.environ["NUMEXPR_NUM_THREADS"] = "1"
```

**To test**
Run a job with MDANSE_GUI on single or multicore and check that all Python processes do not use more than 100 %CPU.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/__init__.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/__init__.py
@@ -13,6 +13,14 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
+import os
+
+os.environ["OMP_NUM_THREADS"] = "1"
+os.environ["OPENBLAS_NUM_THREADS"] = "1"
+os.environ["MKL_NUM_THREADS"] = "1"
+os.environ["VECLIB_MAXIMUM_THREADS"] = "1"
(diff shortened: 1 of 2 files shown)


MBartkowiakSTFC:
Some people in this thread:
https://stackoverflow.com/questions/30791550/limit-number-of-threads-in-numpy
suggest that the environment variables must be set _before_ numpy is imported for it to work.

I thought that this would not matter since the jobs we start are new processes, but you could try setting the environment variables before importing anything from MDANSE, just to be on the safe side.

ChiCheng45:
> Some people in this thread: https://stackoverflow.com/questions/30791550/limit-number-of-threads-in-numpy suggest that the environment variables must be set _before_ numpy is imported for it to work.
> 
> I thought that this would not matter since the jobs we start are new processes, but you could try setting the environment variables before importing anything from MDANSE, just to be on the safe side.

Good point, I have moved them into the `__init__.py` should ensure they get set before anything else gets imported.

MBartkowiakSTFC (approved):
Looks good. I tried it out using the same test as yesterday, and during five minutes of the run I did not see any extra threads.

PR #766, opened 2025-05-02
Fixed the structure factor from scattering function job.

### LINKED ISSUE #711: [BUG] Fix static structure factor from scattering function
**Description of the error**
The static structure factor from scattering function job doesn't appear to do anything useful. The job should take in the DCSF results and return the SSF. In #689 I calculated SSF from $F(q,t=0)$ to test the results using

```math
S^{\text{A-L}}_{\alpha\beta}(Q) = \delta_{\alpha\beta} + \sqrt{c_{\alpha}c_{\beta}} (S^{\text{F-Z}}_{\alpha\beta}(Q) -1)
```

we just need to make the job apply the above equation to the DCSF results taking care to remove/apply the weights.

ChiCheng45:
**Description of work**
The StructureFactorFromScatteringFunction job has now been fixed. This takes a DCSF calculation result as input and calculates the static structure factor from it. SFFSF uses the same weights that were used in the DCSF calculation. The SSF is calculated so that it is the same as the SSF job in MDANSE. The following conversion to the Faber-Ziman definition is done.

```math
F_{\mathrm{coh}}(Q,t=0) = S^{\text{A-L}}_{\alpha\beta}(Q) = \delta_{\alpha\beta} + \sqrt{c_{\alpha}c_{\beta}} (S^{\text{F-Z}}_{\alpha\beta}(Q) -1)
```

image

image

Here, I ran the SSF and SFFSF job on an Argon trajectory where I transmuted atom types to two different ones.

closes #711

Also removed the multicore setting from the NDTSF job since it is not used.

**To test**
Run the SSF and DCSF calculations using the same range for the qvectors. Using the DCSF results run the SFFSF job and compare the results from the SSF calculation with those from the SFFSF job.


MBartkowiakSTFC (approved):
I tried it out on a few trajectories. It works fine, as far as the convergence of DCSF allows it. Also, the updated documentation looks better now.

PR #772, opened 2025-05-07
Threaded GUI

MBartkowiakSTFC:
**Description of work**
Following user feedback, this PR is meant to meant the GUI more responsive by performing some operations in background threads.

The main problem is that Qt forbids threads other than the main thread to create widgets, and will shut down the GUI if this rule is broken. At the moment this can still happen if Action.py receives a signal about the trajectory that finished loading. **Until this has been fixed, this PR has to be considered a draft.**

**Fixes**
1. Each tab has its own QThread with a separate event loop.
2. Trajectory loading is done in threads. Items in the model are placeholders, which then inform other widgets that the loading has finished and trigger an update.
3. The GUI creates a Trajectory instance for each loaded trajectory, and tries to re-use it in different tabs, instead of loading again.

**To test**
All GUI features must be tested.


MBartkowiakSTFC:
This will be done differently, and later.

PR #773, opened 2025-05-08
Add ideal results when using the instrument resolution 

ChiCheng45:
**Description of work**
In many cases, an instrument resolution is applied so that results appear more similar to those obtained from an experiment. However, it is always a good idea to check the results without it applied so that the user can see the quality of their MD simulation and the results derived from it. In many cases, the instrument resolution can hide bad MD analysis results, which can result in misinterpreted data.

In this PR, the ideal instrument resolution is used and saved to the results file whenever a non-ideal instrument resolution is used. This saves the user from having to run two calculations with ideal and non-ideal instrument resolutions.

image

In this plot, we use `instrument_resolution = ('gaussian', {'mu': 0.0, 'sigma': 0.1129})` with a short trajectory. Any peaks seen above other than the one at 0 meV are probably not real. Looking at the ideal results shows that they are probably just noise and a longer trajectory is needed so more correlation frames can be used.

image

Also, MSD has been removed from the main results of GDISF.

**To test**
Run a few of the following with an instrument resolution applied: DOS, CCF, DCSF, DISF, GDISF, PPS, and IR. Check the results, there should be a set of `ideal` results. You may need to use larger sigma values to see a difference, depending on your trajectory.  Run some calculations with ideal resolution, there should not be any extra `ideal` results.


MBartkowiakSTFC (approved):
I tried out all the jobs, and the results looked fine. Also, it's great that the ideal curves are not output twice if the resolution was already ideal.

PR #779, opened 2025-05-13
Add the final progress update to IJob._run_multicore

MBartkowiakSTFC:
**Description of work**
Job progress bars have been showing 100% in single core, but <100% in multicore runs.

**Fixes**
1. Final progress update has been added in `IJob._run_multicore`, once the loop over steps has ended.

**To test**
Run several analysis jobs in multicore mode. They should show 100% completion once finished.


ChiCheng45 (approved):
I've tested this for a few trajectories with multicore and the progress bar seems to be fixed.

PR #780, opened 2025-05-15
GUI warning for trajectories without a fixed cell

### LINKED ISSUE #682: [ENHANCEMENT] Highlight jobs which produced warnings
At the moment, the users have to click on each job in the GUI to view its log. Since we have are gradually adding meaningful warnings to different jobs, it would be good to attract the user's attention to the jobs which produced warnings.

We could either create a new job status for "finished, but with warnings", to differentiate it from those that just finished, or highlight them in the table.

We could also mark the logger tab for the user when a warning pops up, but this is probably not specific enough and the users will start ignoring it quickly.

**Additional context**
N/A

### LINKED ISSUE #764: [ENHANCEMENT] A visible GUI warning for job inputs
Some inputs don't stop the users from running a calculation, but the results are not likely to be useful. We can point this out to users and highlight the entries that are likely to cause problems.

The main point of the warning mechanism is to inform the users that trajectories with variable unit cell dimensions are not supported by the code, and the results are likely to be wrong. The mechanism should be similar to the existing input error mechanism.

We could block the execution completely for the questionable inputs, but this may stop users from running legitimate analysis.

**Additional context**
This is related to the PR #634 which tries to reduce the noise in the results for NPT trajectories. While the changes there make the results slightly better, this is not the same as actually handling variable-cell trajectories correctly. 

The long-term goal to support these trajectories correctly is described in #763

MBartkowiakSTFC:
**Description of work**
The GUI should detect trajectories which do not have a fixed unit cell, and warn the user that the analysis results may be incorrect.

closes #764
closes #682 

**Fixes**
1. Added a warning string to the configurators.
2. WidgetBase has a method for marking warnings.
3. Both the input widget and the run button are highlighted in case of a warning.

**To test**
Load a fixed-cell trajectory, an NPT trajectory, and (optionally) a trajectory without a unit cell, or with a unit cell set to 0.
When an analysis is picked in the GUI, the trajectory widget should be highlighted for all the trajectories except the one with a fixed cell, and the tooltip text should show a warning relevant to each trajectory.


oerc0122 (changes requested):
Honestly, I'm not sure I like the infrastructure of adding new class variables. Particularly separate `unit_cell_warning` and `warning_status`.

Python has the [`warnings`](https://docs.python.org/3/library/warnings.html) module. Which has the capability of catching the warnings.
```python
with warnings.catch_warnings(record=True) as w:
```

By using that, we can unify the warnings infrastructure between the scripting and the GUI interfaces. 

In the GUI handler, just wrapping the configurator in a `catch_warnings` block when calling the `configure` and running with that might be a nicer way of setting this up and allowing us to use warnings more generally.

Happy to be persuaded otherwise though.

MBartkowiakSTFC:
I will rewrite it to use `warnings` instead. There are some pre-defined warning types there. What warning category should it be if the unit cell is missing? Runtime, User, or do we need a new one?

oerc0122:
Given we have a lot of custom exceptions I see no real reason to not use some custom warnings if there's no relevant warning.

MBartkowiakSTFC:
At the moment it is not guaranteed that the `Configurator.configure` method will be called by the widget. For trajectories, we try to load them as little as possible, and so we will be re-using an `HDFTrajectoryConfigurator` instance. For this reason, it still makes sense to store the warning status as an attribute in the `Configurator`. The mechanism of catching a warning requires the widget to exist already when the `configure` is called.

The current version of this PR still uses the `warning_status` attribute. The setter of `warning_status` now produces also a Python warning and logs the warning, to make sure that the warning information is available also outside of the GUI.

Also, in an attempt to clean up the code, most `@property` decorators have been removed from the `Configurator` classes.

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -37,61 +46,58 @@ def default(self, obj):
...
-        :type message: str
-        :param configurator: the configurator in which the exception was raised
-        :type configurator: an instance or derived instance of a MDANSE.Framework.Configurators.Configurator object
-        """
+class ConfiguratorError(Error):
+    """Error raised by a job input parser."""
 
-        self._message = message
-        self._configurator = configurator
+    def __init__(self, message: str, configurator: IConfigurator = None):
```
```suggestion
    def __init__(self, message: str, configurator: IConfigurator | None = None):
```

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -130,206 +127,180 @@ def __init__(self, name, **kwargs):
...
 
-        self._label = kwargs.get(
+        self.label = kwargs.get(
             "label",
             (
-                self.__class__._label
-                if hasattr(self.__class__, "_label")
+                self.__class__.label
+                if hasattr(self.__class__, "label")
                 else " ".join(name.split("_")).strip()
```
```suggestion
                getattr(type(self), "label", name.replace("_", " ").strip())
```

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -130,206 +127,180 @@ def __init__(self, name, **kwargs):
             "_error_status",
         ]
 
-        self._configurable = kwargs.get("configurable", None)
+        self.configurable = kwargs.get("configurable")
```
oerc0122, line 130: Really, these shouldn't be in `**kwargs`, but could be kwargs after `*` to allow proper typing, defaults and inspection. Not for this PR though.

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -130,206 +127,180 @@ def __init__(self, name, **kwargs):
...
+
         """
         self._error_status = error_text
-        if len(self._error_status) > 2:
-            self._valid = False
+        if len(self._error_status) > ERROR_LENGTH_MIN:
+            self.valid = False
         else:
-            self._valid = True
+            self.valid = True
```
```suggestion
        self.valid = error_text == "OK"
```

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -130,206 +127,180 @@ def __init__(self, name, **kwargs):
...
-                if self._configurable._configuration[name].is_configured()
+                if self.configurable._configuration[name].is_configured()
             ]
 
-        for c in list(self._dependencies.values()):
-            if c not in configured:
-                return False
-
-        return True
+        return all(c in configured for c in list(self.dependencies.values()))
```
```suggestion
        return all(c in configured for c in self.dependencies.values())
```

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -130,206 +127,180 @@ def __init__(self, name, **kwargs):
...
+        """
         if configured is None:
-            names = [str(key) for key in self._configurable._configuration.keys()]
+            names = [str(key) for key in self.configurable._configuration]
             configured = [
                 name
                 for name in names
-                if self._configurable._configuration[name].is_configured()
+                if self.configurable._configuration[name].is_configured()
             ]
```
```suggestion
            configured = [
                str(name)
                for name, prop in self.configurable._configuration.values()
                if prop.is_configured()
            ]
```

MDANSE/Src/MDANSE/Framework/Configurators/MultipleChoicesConfigurator.py
```diff
@@ -56,15 +56,15 @@ def configure(self, value):
...
                 return
 
         indices = []
         for v in value:
             try:
-                indices.append(self._choices.index(v))
+                indices.append(self.choices.index(v))
             except ValueError:
                 self.error_status = f"{v} item is not a valid choice"
                 return
```
oerc0122:
```suggestion
        try:
            indices = [self._choices.index(v) for v in value]
        except ValueError:
            self.error_status = f"{', '.join(set(value) - set(self._choices))} are not valid choices"
            return

File: MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py [hunk 8]
```diff
@@ -253,6 +260,16 @@ def _load_unit_cells(self):
...
+            self.unit_cell_warning = NO_CELL
+        if not self.unit_cell_warning:
+            if abs(self._unit_cells[0].volume) < CELL_SIZE_LIMIT:
+                self.unit_cell_warning = BAD_CELL
+                return
+            reference_array = self._unit_cells[0].direct
+            for uc in self._unit_cells[1:]:
+                if not np.allclose(reference_array, uc.direct):
+                    self.unit_cell_warning = CHANGING_CELL
+                    return
```
oerc0122:
```suggestion
            if any(not np.allclose(reference_array, uc.direct) 
                   for uc in self._unit_cells[1:]):
                self.unit_cell_warning = CHANGING_CELL
                return
```

oerc0122 (approved):
Just a couple of minor stylistic things. Feel free to pick and choose.

PR #783, opened 2025-05-16
Reimplement grouping as analysis input

### LINKED ISSUE #644: [ENHANCEMENT] Combine atom selection and grouping
At the moment, atom selection allows the users to pick a subset of atoms on which to perform the analysis. Grouping allows to output the results not per element but per molecule (averaged or summed over atoms in the molecule). However, there are no checks for situations in which the users removes half of one molecule using the selection, which should not be allowed.

One solution would be to have the grouping level determine the valid selection options. For example, if molecule grouping is active, the selection options should only allow to select or remove entire molecules and not single atoms.

Any other ideas are welcome.

**Additional context**
N/A

### LINKED ISSUE #680: [BUG] Inconsistent MSD total when atom or molecule grouping level is used
**Description of the error**
Here I calculated the MSD with the cubane discover trajectory which C60 and cubane molecules.

Image

Image

Results1 is the one that uses atom grouping. I think the issue might be due to the weighting scheme since I'm not sure how it should work with molecule grouping.

MBartkowiakSTFC:
**Description of work**
Following user feedback, new grouping mechanism has been introduced. The main intention was to make it compatible with the new atom selection mechanism.

Affected analysis types:
Elastic Incoherent Structure Factor
Dynamic Incoherent Structure Factor
Gaussian Dynamic Incoherent Structure Factor
Position Autocorrelation Function
Mean Square Displacement
Root Mean Square Deviation

More work will be needed in the future to enable grouping in the Root Mean Square Fluctuation analysis.

closes #644
closes #680

**Fixes**
1. Apart from the default "atom", two new grouping options are "average over molecules" and "individual molecules".
2. GroupingTool class has been introduced to assign the correct results to their corresponding output datasets.
3. The atom-averaged results are always included. The molecule-grouped results are produced from the same calculations, in order to reduce repetition in the calculations.
4. The analysis jobs still iterate over individual atom indices, same as for the "atom" grouping.
5. Unit tests have been added for the analysis types that use grouping.
6. At the moment, molecules which are not fully within the selection are **omitted** from the calculation.

**To test**
Load a trajectory that contains molecules. Generate a Centre of Mass Trajectory out of it.
Run the jobs listed above on the COM trajectory (using "atom" grouping"), and on the normal trajectory using the molecule grouping. The results of the COM analysis and molecule-grouped analysis should be close to each other, but not the same.


MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -66,6 +67,10 @@ def initialise_atoms(self, element_list: List[str], name_list: List[str] = None)
         self._unique_elements.update(set(element_list))
         if name_list is not None:
             self._atom_names = [str(x) for x in name_list]
+        for element in self._unique_elements:
+            self.element_indices[element] = {
```
oerc0122, line 71: Bear in mind that `set` is unordered. Doubt this will cause problems here, though.

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        """Store the dictionary with weights per atom type.
+
+        Parameters
+        ----------
+        new_weights : dict[tuple[str],float]
+            dictionary of {atom_types: scaling_factor} pairs
+
+        """
+        self._weight_dictionary = {
+            "".join([*key]): value for key, value in new_weights.items()
```
oerc0122:
```suggestion
            "".join(key): value for key, value in new_weights.items()
```
If key is a sequence this will do.

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+
+        Parameters
+        ----------
+        text_key : str
+            name of the grouping level in GROUPING_LABELS
+
+        """
+        self._grouping = GROUPING_LABELS.get(text_key, GroupingLevel.ATOM)
+
+    def set_weight_dictionary(self, new_weights: dict[tuple[str], float]):
```
oerc0122, line 97:
```suggestion
    def set_weight_dictionary(self, new_weights: dict[tuple[str, ...], float]):
```
If it is an arbitrary sized tuple
```suggestion
    def set_weight_dictionary(self, new_weights: dict[tuple[str, str], float]):
```
If it is fixed size (in this case 2)

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+
+        Parameters
+        ----------
+        parameters : dict[str, Any]
+            dictionary of parameters needed by OutputData.write
+
+        """
+        for key in self._mandatory_keys:
+            if key not in parameters:
+                raise KeyError(f"Setting {key} is missing from data parameters.")
```
oerc0122:
```suggestion
        if missing := self._mandatory_keys - parameters.keys():
            raise KeyError(f"Settings ({', '.join(missing)}) are missing from data parameters.")
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        for key in self._mandatory_keys:
+            self._data_parameters[key] = parameters[key]
+        for key in self._extra_keys:
+            value = parameters.get(key)
+            if value is None:
+                value = BACKUP_DATA_PARAMETERS[key]
+            self._data_parameters[key] = value
```
oerc0122:
```suggestion
            self._data_parameters[key] = parameters.get(key, BACKUP_DATA_PARAMETERS[key])
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+    MOL_AVERAGE = auto()
+
+
+GROUPING_LABELS = {
+    "atom": GroupingLevel.ATOM,
+    "individual molecules": GroupingLevel.MOL_EACH,
+    "average over molecules": GroupingLevel.MOL_AVERAGE,
+}
+
+BACKUP_DATA_PARAMETERS = {
```
oerc0122: I would call these `DEFAULT` rather than `BACKUP`

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
oerc0122: In fact there's no need to have two loops here. You've guaranteed the mandatory keys are in the keys, so we can instead just make this:

```suggestion
        for key in self._mandatory_keys + self._extra_keys:
            self._data_parameters[key] = parameters.get(key, BACKUP_DATA_PARAMETERS.get(key))
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+            keywords to be used by this dataset, by default None
+        weight_key : Optional[str], optional
+            scaling factor of this dataset, by default None
+
+        """
+        self._debug_counter[name] += 1
+        pardict = {key: self._data_parameters[key] for key in self._extra_keys}
+        if override:
+            for key, value in override.items():
+                pardict[key] = value
```
oerc0122:
```suggestion
            pardict.update(override)
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+            key used by the OutputData dictionary
+        override : Optional[dict[str, Any]], optional
+            keywords to be used by this dataset, by default None
+        weight_key : Optional[str], optional
+            scaling factor of this dataset, by default None
+
+        """
+        self._debug_counter[name] += 1
+        pardict = {key: self._data_parameters[key] for key in self._extra_keys}
+        if override:
```
oerc0122: May want to check:
```suggestion
        if override is not None:
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        if self._data_parameters is None:
+            raise RuntimeError("Creating output data groups without parameters.")
+        if not self._current_selection:
+            raise RuntimeError("Trying to group an empty selection.")
+        self.create_atom_groups(name)
+        if self._grouping == GroupingLevel.MOL_AVERAGE:
+            self.create_averaged_molecule_groups(name)
+        elif self._grouping == GroupingLevel.MOL_EACH:
+            self.create_individual_molecule_groups(name)
```
oerc0122:
```suggestion
        if self._grouping is GroupingLevel.MOL_AVERAGE:
            self.create_averaged_molecule_groups(name)
        elif self._grouping is GroupingLevel.MOL_EACH:
            self.create_individual_molecule_groups(name)
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        name : str
+            root of the dataset names
+
+        """
+        for atom_type in self._cs._unique_elements:
+            indices = self._current_selection.intersection(
+                self._cs.element_indices[atom_type]
+            )
+            if indices:
+                dset_name = "_".join([name, str(atom_type)])
```
oerc0122:
```suggestion
                dset_name = f"{name}_{atom_type}"
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+                self.add_dataset(dset_name, weight_key=str(atom_type))
+                self._output_data[dset_name].atom_indices = list(indices)
+                self._indices_per_data_key[dset_name] = indices
+                self._plain_datasets.add(dset_name)
+        dset_name = "_".join([name, "total"])
```
oerc0122:
```suggestion
        dset_name = f"{name}_total"
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        for molecule in self._cs.unique_molecules():
+            all_indices = set()
+            for mindex, mol in enumerate(self._cs._clusters[molecule]):
+                trimmed_mol = self._current_selection.intersection(mol)
+                if set(mol) == trimmed_mol:
+                    all_indices.update(mol)
+                elif len(trimmed_mol):
+                    molecules_dropped.append(mindex)
+            if all_indices:
+                dset_name = "_".join([name, str(molecule), "all"])
```
oerc0122:
```suggestion
                dset_name = f"{name}_{molecule}_all"
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+            root of the dataset names
+
+        """
+        molecules_dropped = []
+        for molecule in self._cs.unique_molecules():
+            all_indices = set()
+            for mindex, mol in enumerate(self._cs._clusters[molecule]):
+                trimmed_mol = self._current_selection.intersection(mol)
+                if set(mol) == trimmed_mol:
+                    all_indices.update(mol)
```
oerc0122:
```suggestion
                    all_indices |= mol
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        name : str
+            root of the dataset names
+
+        """
+        molecules_dropped = []
+        for molecule in self._cs.unique_molecules():
+            for mindex, mol in enumerate(self._cs._clusters[molecule]):
+                trimmed_mol = self._current_selection.intersection(mol)
+                if set(mol) == trimmed_mol:
+                    dset_name = "_".join([name, str(molecule), str(mindex + 1)])
```
oerc0122:
```suggestion
                    dset_name = f"{name}_{molecule}_{mindex + 1}"
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+
+        """
+        molecules_dropped = []
+        for molecule in self._cs.unique_molecules():
+            all_indices = set()
+            for mindex, mol in enumerate(self._cs._clusters[molecule]):
+                trimmed_mol = self._current_selection.intersection(mol)
+                if set(mol) == trimmed_mol:
+                    all_indices.update(mol)
+                elif len(trimmed_mol):
```
oerc0122:
```suggestion
                elif trimmed_mol:
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        for molecule in self._cs.unique_molecules():
+            for mindex, mol in enumerate(self._cs._clusters[molecule]):
+                trimmed_mol = self._current_selection.intersection(mol)
+                if set(mol) == trimmed_mol:
+                    dset_name = "_".join([name, str(molecule), str(mindex + 1)])
+                    self.add_dataset(dset_name, override={"main_result": False})
+                    self._output_data[dset_name].atom_indices = list(trimmed_mol)
+                    self._indices_per_data_key[dset_name] = trimmed_mol
+                    self._molecule_datasets.add(dset_name)
+                elif len(trimmed_mol):
```
oerc0122:
```suggestion
                elif trimmed_mol:
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+            LOG.warning(
+                "Molecules %s were not fully selected and will be omitted",
+                molecules_dropped,
+            )
+
+    def assign_result(
+        self,
+        index: int,
+        result: np.ndarray,
+        normalise: bool = True,
```
oerc0122, line 308: Generally considered good practise to make `bool`s kw-only.

```suggestion
        *,
        normalise: bool = True,
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+            if True, results for each atom of the molecule will be scaled
+            by the atom mass
+
+        """
+        with self._mutex:
+            self.assign_result_molecular(index, result, centre_of_mass)
+            self.assign_result_atomic(index, result, normalise)
+
+    def assign_result_molecular(
+        self, index: int, result: np.ndarray, centre_of_mass: bool = True
```
oerc0122:
```suggestion
        self, index: int, result: np.ndarray, *, centre_of_mass: bool = True
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+                if centre_of_mass:
+                    atom_mass = self._mass_dictionary[atom_type]
+                    self._output_data[dset_name] += result * atom_mass
+                    self._molecule_mass[dset_name] += atom_mass
+                else:
+                    self._output_data[dset_name] += result
+                    self._molecule_mass[dset_name] += 1.0
+
+    def assign_result_atomic(
+        self, index: int, result: np.ndarray, normalise: bool = True
```
oerc0122:
```suggestion
        self, index: int, result: np.ndarray, *, normalise: bool = True
```

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
```diff
@@ -113,14 +113,18 @@ def write(
...
         # Loop over the OutputVariable instances to write.
 
-        for var in list(data.values()):
+        for key, var in data.items():
             varName = str(var.varname).strip().replace("/", "|")
 
             dset = outputFile.create_dataset(varName, data=var, shape=var.shape)
 
             # All the attributes stored in the OutputVariable instance are written to the HDF file.
             for k, v in list(vars(var).items()):
```
oerc0122:
```suggestion
            for k, v in vars(var).items():
```

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
```diff
@@ -113,14 +113,18 @@ def write(
...
-                dset.attrs[k] = v
+                try:
+                    dset.attrs[k] = v
+                except TypeError:
+                    print(f"type error: {k}: {v} in {key}")
+                    raise RuntimeError("Bad HDF5 write")
```
oerc0122:
```suggestion
                except TypeError as error:
                    log.ERROR("TypeError: %s: %s in %s", k, v, key)
                    raise RuntimeError("Bad HDF5 write.") from error
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -113,11 +108,67 @@ def initialize(self):
...
+
+        self._groupers = {}
+        self._groupers["fqt"] = GroupingTool(
+            self.configuration["trajectory"]["instance"].chemical_system,
+            self._outputData,
+        )
+        self._groupers["sqf"] = GroupingTool(
+            self.configuration["trajectory"]["instance"].chemical_system,
+            self._outputData,
+        )
```
oerc0122:
```suggestion
        self._groupers = {
            "fqt": GroupingTool(
                self.configuration["trajectory"]["instance"].chemical_system,
                self._outputData,
            ),
            "sqf": GroupingTool(
                self.configuration["trajectory"]["instance"].chemical_system,
                self._outputData,
            ),
        }
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -259,10 +260,10 @@ def combine(self, index, disf_per_q_shell):
             #. index (int): The index of the step.\n
             #. x (any): The returned result(s) of run_step
         """
-
-        element = self.configuration["atom_selection"]["names"][index]
-        for i, v in enumerate(disf_per_q_shell.values()):
-            self._outputData[f"f(q,t)_{element}"][i, :] += v
+        self._groupers["fqt"].assign_result(
+            self.configuration["atom_selection"]["flatten_indices"][index],
+            np.array([v for v in disf_per_q_shell.values()]),
```
oerc0122:
```suggestion
            np.array(list(disf_per_q_shell.values())),
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -271,50 +272,29 @@ def finalize(self):
...
-                axis=1,
-            )
-            if self.add_ideal_results:
-                self._outputData[f"s(q,f)_ideal_{element}"][:] = get_spectrum(
-                    self._outputData[f"f(q,t)_{element}"],
-                    None,
+        self._groupers["fqt"].finalise_centre_of_mass()
+
+        for inkey in self._outputData.keys():
+            if "f(q,t)" in inkey:
```
oerc0122:
```suggestion
        for inkey in self._outputData:
            if "f(q,t)" in inkey:
```
Could also write as:
```suggestion 
        for inkey in (key for key in self._outputData if "f(q,t)" in key):
```
And remove a layer of nesting

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -271,50 +272,29 @@ def finalize(self):
...
+                outkey = "_".join(["s(q,f)"] + inkey.split("_")[1:])
```
oerc0122:
```suggestion
                outkey = "s(q,f)_{inkey.split('_', 1)[1]}"
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -271,50 +272,29 @@ def finalize(self):
...
-
-        if self.add_ideal_results:
-            self._outputData["s(q,f)_ideal_total"][:] = weighted_sum(
-                self._outputData,
-                weight_dict,
-                "s(q,f)_ideal_%s",
-            )
-
+                if self.add_ideal_results:
+                    outkey = "_".join(["s(q,f)_ideal"] + inkey.split("_")[1:])
```
oerc0122:
```suggestion
                    outkey = "s(q,f)_ideal_{inkey.split('_', 1)[1]}"
```
Although I'd probably recommend making `inkey.split("_", 1)[1]` a named variable since you're using it twice

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -271,50 +272,29 @@ def finalize(self):
...
+                    self._outputData[outkey][:] = get_spectrum(
+                        self._outputData[inkey],
+                        None,
+                        self.configuration["instrument_resolution"]["time_step"],
+                        axis=1,
+                    )
+
+        for group in self._groupers:
+            if group != "fqt":
```
oerc0122: Can remove layer of nesting
```suggestion
        for group in filter(lambda x: x != "fqt", self._groupers):
```
Or perhaps more pythonic, this also returns the grouper directly meaning we don't need to `getitem` from `self._groupers`
```suggestion
        for grouper in (value for key, value in self._groupers.items() if key != "fqt"):
```

oerc0122: Can remove layer of nesting
```suggestion
        for group in filter(lambda x: x != "fqt", self._groupers):
```
Or perhaps more pythonic, this also returns the grouper directly meaning we don't need to `getitem` from `self._groupers`
```suggestion
        for grouper in (value for key, value in self._groupers.items() if key != "fqt"):
```

MDANSE/Src/MDANSE/Framework/Jobs/ElasticIncoherentStructureFactor.py
```diff
@@ -97,10 +92,26 @@ def initialize(self):
         Initialize the input parameters and analysis self variables
         """
         super().initialize()
+        self._nQShells = self.configuration["q_vectors"]["n_shells"]
 
         self.numberOfSteps = self.configuration["atom_selection"]["selection_length"]
```
oerc0122, line 97: Any reason for re-ordering this?

MDANSE/Src/MDANSE/Framework/Jobs/GaussianDynamicIncoherentStructureFactor.py
```diff
@@ -150,69 +145,63 @@ def initialize(self):
...
+        msd_dict = get_weights(
+            {atom: 1.0 for atom in nAtomsPerElement}, nAtomsPerElement, 1
+        )
+
+        self._groupers = {}
+        for key in ["fqt", "sqf", "msd"]:
+            self._groupers[key] = GroupingTool(
+                self.configuration["trajectory"]["instance"].chemical_system,
+                self._outputData,
             )
```
oerc0122:
```suggestion
        self._groupers = {
            key: GroupingTool(
                self.configuration["trajectory"]["instance"].chemical_system,
                self._outputData,
            ) 
            for key in ("fqt", "sqf", "msd")
        }
```

MDANSE/Src/MDANSE/Framework/Jobs/GaussianDynamicIncoherentStructureFactor.py
```diff
@@ -267,67 +256,43 @@ def combine(self, index: int, x: tuple[np.ndarray, np.ndarray]):
...
-            self._outputData[f"msd_{element}"][:] /= number
-            if self.add_ideal_results:
-                self._outputData[f"s(q,f)_ideal_{element}"][:] = get_spectrum(
-                    self._outputData[f"f(q,t)_{element}"],
-                    None,
+        self._groupers["fqt"].finalise_centre_of_mass()
+
+        for inkey in self._outputData.keys():
+            if "f(q,t)" in inkey:
+                outkey = "_".join(["s(q,f)"] + inkey.split("_")[1:])
```
oerc0122: See above

MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py
```diff
@@ -98,22 +109,21 @@ def initialize(self):
...
         self._atoms = self.configuration["trajectory"][
             "instance"
         ].chemical_system.atom_list
 
+        nAtomsPerElement = self.configuration["atom_selection"].get_natoms()
+
+        weights = self.configuration["weights"].get_weights()
+        weight_dict = get_weights(weights, nAtomsPerElement, 1)
+        self._grouper.set_weight_dictionary(weight_dict)
+        print(weight_dict)
```
oerc0122: Leftover print?

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareDeviation.py
```diff
@@ -87,29 +100,16 @@ def initialize(self):
...
-            axis="time",
-            units="nm",
-            main_result=True,
-        )
-
         self._atoms = self.configuration["trajectory"][
             "instance"
         ].chemical_system.atom_list
+        nAtomsPerElement = self.configuration["atom_selection"].get_natoms()
+        weight_dict = {key: 1.0 for key in nAtomsPerElement}
```
oerc0122:
```suggestion
        weight_dict = dict.fromkeys(nAtomsPerElement, 1.0)
```

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        for dset_name in self._molecule_datasets:
+            norm = self._molecule_mass[dset_name]
+            if np.isclose(norm, 0):
+                norm = 1.0
+            self._output_data[dset_name] /= norm
+            if not np.isclose(exponent, 1.0):
+                self._output_data[dset_name][:] = (
+                    self._output_data[dset_name] ** exponent
+                )
```
oerc0122, line 418: It should be noted that there may be optimisations in place for e.g. `sqrt`/`cbrt`, which cannot be done for generalised exponents.

MDANSE/Src/MDANSE/Framework/OutputVariables/IOutputVariable.py
```diff
@@ -61,6 +61,7 @@ def __new__(
         *,
         main_result: bool = False,
         partial_result: bool = False,
+        dtype: type = np.float64,
```
oerc0122: Strictly these should be of type [`np.dtype`](https://numpy.org/doc/2.2/reference/arrays.dtypes.html)

Not sure it matters too much.

MDANSE/Src/MDANSE/Framework/OutputVariables/IOutputVariable.py
```diff
@@ -87,9 +88,9 @@ def __new__(
             If dimensions of provided data do not align with those of object.
         """
         if isinstance(value, tuple):
-            value = np.zeros(value, dtype=np.float64)
+            value = np.zeros(value, dtype=dtype)
         else:
-            value = np.array(list(value), dtype=np.float64)
+            value = np.array(list(value), dtype=dtype)
```
oerc0122, line 93: This will not correctly convert nested data.

oerc0122 (commented):
A few things to consider.

MDANSE/Src/MDANSE/Framework/Jobs/ElasticIncoherentStructureFactor.py
MBartkowiakSTFC: The only reason was that it had to be set before the call to `self._grouper.set_dataset_parameters`. Any other position in the code before that line would have worked too.

MDANSE/Src/MDANSE/Framework/OutputVariables/IOutputVariable.py
MBartkowiakSTFC: Please tell me more about what scenario will make it fail.

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
MBartkowiakSTFC: You're right. Here we only use the sets to check which index belongs for which set, so it should be fine.

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
MBartkowiakSTFC: The way MDANSE works, in all the jobs that use the GroupingTool this will be a fixed-size tuple containing a single string.

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
MBartkowiakSTFC: Since 0.5 is what we really expect here, I added a check for it. For exponent 0.5 we will use `np.sqrt`.

MDANSE/Src/MDANSE/Chemistry/GroupingTool.py
oerc0122: In which case this is fine, just wanted to double check because that's an odd use of a tupe.

MDANSE/Src/MDANSE/Framework/OutputVariables/IOutputVariable.py
oerc0122: If you had a generator of generators, this will only convert the top level. Not sure if this is a likely case coming in.

ChiCheng45:
I've run a GDISF calculation using the DFTB water trajectory. I get a different result for `f(q,t)_total` and `f(q,t)_H2_O1_all`. 

image


I know that the grouping level is `average over molecules` but I don't think it make sense for grouping to work like that for this and other similar analysis jobs. The total intermediate scattering function is

```math
F_{\mathrm{inc}, \mathrm{tot}}(q, t) = \frac{1}{N} \sum_{j} w_j \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle 
```

when we use the atom grouping, the above sum is split up by atom.

```math
F_{\mathrm{inc}, \mathrm{tot}}(q, t) = \sum_{\alpha} W_{\alpha} F_{\mathrm{inc}, \alpha}(q, t) 
```

```math
F_{\mathrm{inc}, \alpha}(q, t) = \frac{1}{Nc_{\alpha}} \sum_{j \in \alpha} \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle 
```

For the molecular case, we could split the summation into atomic parts associated to each molecule. For a water and ethanol system, we would get 

```math
F_{\mathrm{inc}, \mathrm{tot}}(q, t) = F_{\mathrm{inc}, \mathrm{H2O}}(q, t) + F_{\mathrm{inc}, \mathrm{EtOH}}(q, t) 
```

where

```math
F_{\mathrm{inc}, \mathrm{H2O}}(q, t) = W^{\mathrm{H2O}}_{\mathrm{H}}F_{\mathrm{inc}, \mathrm{H}}^{\mathrm{H2O}}(q, t) + W^{\mathrm{H2O}}_{\mathrm{O}}F_{\mathrm{inc}, \mathrm{O}}^{\mathrm{H2O}}(q, t)
```

and

```math
F_{\mathrm{inc}, \mathrm{EtOH}}(q, t) = W^{\mathrm{EtOH}}_{\mathrm{H}}F_{\mathrm{inc}, \mathrm{H}}^{\mathrm{EtOH}}(q, t) + W^{\mathrm{EtOH}}_{\mathrm{C}}F_{\mathrm{inc}, \mathrm{C}}^{\mathrm{EtOH}}(q, t) + W^{\mathrm{EtOH}}_{\mathrm{O}}F_{\mathrm{inc}, \mathrm{O}}^{\mathrm{EtOH}}(q, t)
```

The intermediate scattering functions are

```math
F_{\mathrm{inc}, \mathrm{H}}^{\mathrm{H2O}}(q, t) = \frac{1}{Nc^{\mathrm{H2O}}_{\mathrm{H}}} \sum_{j \in (\mathrm{H2O} \cap \mathrm{H})} \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle
```
and 

```math
W^{\mathrm{H2O}}_{\mathrm{H}} = c_{\mathrm{H}}^{\mathrm{H2O}} w_{\mathrm{H}}
```

where $c_{\mathrm{H}}^{\mathrm{H2O}} = N_{\mathrm{H}}^{\mathrm{H2O}} / N$ and $N_{\mathrm{H}}^{\mathrm{H2O}}$ is the number of hydrogen atoms in water. 


For the water trajectory above my expectations would've been 

```math
F_{\mathrm{inc}, \mathrm{tot}}(q, t) = F_{\mathrm{inc}, \mathrm{H2O}}(q, t)
```

so that the plots above should be the same since I had the `Apply weights?` checked.

MBartkowiakSTFC:
Thanks for this comment. If we implemented grouping the way you described it would certainly make more sense to me too.

In the code of MDANSE prior to this PR, any grouping of the results other than "atom" would result in MDANSE taking several atoms, then calculating the trajectory of _their centre of mass_, and running the analysis on this. This, of course, makes me wonder why we needed grouping to start with, since it is possible to run CenterOfMassTrajectory first and then run the further analysis on it.

@sanghamitra-mukhopadhyay What is grouping supposed to do?
1. Split the total signal into proportional contributions from specific molecules or specific molecule types?
2. Group atoms into averaged objects with averaged dynamics, which can then be analysed?
3. Something else?
You are welcome to give us specific description for the two separate cases, grouping into individual molecules and averaging over molecules of each type.

MBartkowiakSTFC:
This has been replaced by #810

PR #787, opened 2025-05-19
Check for length-1 arrays in PlottingContext

### LINKED ISSUE #786: [BUG] Plot shows nan in the legend
**Description of the error**
Running a job like DISF with only a single q-shell causes nan values to appear in the plots. The labelling of the legend works workly when two or more q-shells are used.

**Additional details**

Image

Image

MBartkowiakSTFC:
**Description of work**
At the moment, labels for 1D cuts of a 2D array do not work correctly if the array is shaped (1, N). This PR adds a correction to generate a meaningful label in this case.

closes #786 

**Fixes**
A check for the axis array length has been added. A different formula is used to calculate rounding precision in that case.

**To test**
Load a 2D array with only one row, and plot the results. The plot legend should contain the value of the corresponding axis at its only index.


ChiCheng45 (approved):
Great, looks like it's fixed.

PR #790, opened 2025-05-20
Fix gui so that trajectory editor loads up correctly

### LINKED ISSUE #789: [BUG] Trajectory editor job does not load up in the GUI
**Description of the error**
Trajectory editor job does not load up properly in the GUI. Looks like this bug was introduced in #757.

**Additional details**
Image

ChiCheng45:
**Description of work**
Fixes a recent bug that was introduced. The trajectory editor job now loads up correctly.

**Fixes**
Fixes #789

**To test**
Load a trajectory and open the trajectory editor job. The job should now load up correctly. Run a trajectory editor job and check that a new trajectory is made.


MBartkowiakSTFC (approved):
It works, thanks.

PR #796, opened 2025-05-22
Fix to unit conversion

### LINKED ISSUE #793: [BUG] MDANSE measure doesn't convert some units correctly.
**Description of the error**
When I convert some units they are not correct.

```pyython
a = measure(1.0, "1/cm")
a.equivalent = True
a.toval("kcal/mol")
Out[5]: 2.8591539196940727e-06
```

Should be 0.0028591 (https://wild.life.nctu.edu.tw/class/common/energy-unit-conv-table.html), its out by a factor of 1000.

ChiCheng45:
**Description of work**
Corrected unit equivalence, corrected j_per_mole and cal_per_mole units, and updated unit tests. 

**Fixes**
Fixes #793

**To test**
Unit tests should pass.


MBartkowiakSTFC:
The changes are good, especially the additional unit tests.

One problem: if a user has been running a version of MDANSE with the old units, and created their own units, MDANSE will have saved a local copy of `units.json`. Users who did that will not see the changes from this PR, because their MDANSE will keep using the local copy of `units.json` with the old values.

MBartkowiakSTFC:
The thing I wrote above opens many new questions and there would also be several possible solutions, so I am happy not to deal with this problem here, but in some other PR later on. I don't really see any benefits of defining new units in MDANSE.

ChiCheng45:
> The changes are good, especially the additional unit tests.
> 
> One problem: if a user has been running a version of MDANSE with the old units, and created their own units, MDANSE will have saved a local copy of `units.json`. Users who did that will not see the changes from this PR, because their MDANSE will keep using the local copy of `units.json` with the old values.

I guess we could get MDANSE to check that the data in the cache is valid during the startup probably need to do it as soon as the json file is loaded. The same could probably be done for the atoms.json for the default atoms. Would you prefer it to be done in this PR or another?

MBartkowiakSTFC:
Let's do it in another PR.

Similar to what you did with the chemical element definitions, it would most likely make sense to disable changes to the existing units. And I am not sure if it makes sense to allow defining new units, since it is not likely that they will be used anywhere.

Another comment: in #740 I wanted to suggest creating a new file name for what is now `atoms.json`, since the new file is not compatible with the old one. MDANSE would log a warning to inform the user that their old local file will be ignored. A similar approach may not be the best idea here, but I just thought I would mention it.

MBartkowiakSTFC (approved):
Looks good, thanks!

ChiCheng45:
Thanks, I've created #797 for the json file checking suggestion.

ChiCheng45:
> Similar to what you did with the chemical element definitions, it would most likely make sense to disable changes to the existing units. And I am not sure if it makes sense to allow defining new units, since it is not likely that they will be used anywhere.

I think it can be useful for users to define their own units. I've noticed that the MDAnalysis converter might not get the units correct in some cases. If MDAnalysis doesn't know what the units are it just set them to the default. So it might be good for users to create their own units and have the option to tell MDANSE to convert the units from the MDAnalysis defaults to whatever it was supposed to be.

PR #799, opened 2025-05-23
Adjust the MDANSE GUI logo

ChiCheng45:
**Description of work**
Adjust the MDANSE logo so that it is more symmetric. The current MDANSE GUI logo is slightly asymmetric for some reason. Generally, symmetric logos are more visually appealing. I've also cleaned up the splash screen so that we use high resolution version of all the logos. Also, updated tutorial links in the documentation.

Before
image

After
image

**To test**
Run the MDANSE GUI and check that the splash screen looks OK on your computer.


ChiCheng45:
@MBartkowiakSTFC do you have the original files for the splash screen?

MBartkowiakSTFC (approved):
No problems, the new images look good.

PR #800, opened 2025-05-27
Use configurator defaults for VectorWidget

oerc0122:
**Description of work**
Since `VectorWidget` is always associated with a `Configurator`, use its defaults rather than `[0, 0, 1]`.

**Fixes**
Uses `Configurator` to set initial default.

**To test**
Ensure the GUI loads vector widgets with correct defaults.


MBartkowiakSTFC (approved):
While it is not really used in the main branch at the moment, I checked that it works.

PR #805, opened 2025-05-28
Add missing attributes when axis=index in plotters

### LINKED ISSUE #804: [BUG] Plotting fails for 1D array plotted against index
**Description of the error**
Recently, plotting stopped working for data arrays which do not have their own plotting axis and are plotted against 'index'. At the moment, no plot appears, and the following error is logged:

`2025-05-28 14:48:33,820 - ERROR - process[54655] - PlotTab 115 - Visualiser failed to plot data: 'SingleDataset' object has no attribute '_axes_scaling'
`

A curve should appear, each point plotted against the index of the data point (0, 1, ...)

Plot is empty, with the error message included above.

**Suggested fix**
SingleDataset needs to be corrected.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Create additional entries in SingleDataset to handle 1D arrays plotted against 'index'.

closes #804

**Fixes**
_axes_scaling and  _axes_order are assigned values also for the key 'index'.

**To test**
Load an .mda file of DOS results and plot 'omega' dataset. The points should be plotted as a straight line. (This currently fails in the main branch).

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
@@ -95,14 +95,18 @@ class SingleDataset:
                 self._scaling_factor = np.array(source[name].attrs["scaling_factor"])
         self._axes = {}
         self._axes_units = {}
+        self._axes_scaling = {}
+        self._current_units = {}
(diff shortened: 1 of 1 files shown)


ChiCheng45 (approved):
Great looks like it is fixed.

PR #808, opened 2025-05-30
Show file header in the GUI, remove unused code

### LINKED ISSUE #695: [ENHANCEMENT] Redundant Fortran output
A [source search for the term "fortran"](https://github.com/search?q=repo%3AISISNeutronMuon%2FMDANSE%20fortran&type=code) reveals that the `MDANSE.IO.FortranFormat` classes are entirely unused and should probably be removed.

Removal of redundant code.

N/A

**Additional context**
N/A

### LINKED ISSUE #730: [ENHANCEMENT] Native Python `TextFile` compression
`MDANSE.IO.TextFile` relies on external programs run through `os.popen`.

Python provides [`zipfile`](https://docs.python.org/3/library/zipfile.html#module-zipfile) and [`tarfile`](https://docs.python.org/3/library/tarfile.html#module-tarfile) for native handling

N/A

**Additional context**
N/A

### LINKED ISSUE #807: [RFC] Configurable._info is not being used
**Description of the error**
At the moment, each Configurable (and IJob) instance has an attribute `_info` which stores the output of `Configurator.get_information()` method of all the `Configurator` instances in the settings dictionary. Later, the jobs typically pass the `_info` to `OutputData.write` as **header** and the text contained there gets written into the output HDF5 file as an attribute. However:
- MDANSE_GUI does not access the contents of that attribute,
- the information from the individual `Configurator` instances is usually incomplete,
- the `Configurator.get_information()` method provides some information about its current state, but not about the input that the `Configurator` received,
- MDANSE output files now have a `metadata` dataset which stores the inputs. For jobs which use q vectors, the q vectors are also output as datasets.

**Suggested fix**
There are several options of how to improve the situation:
0. Improve the `get_information` output of each `Configurator` to make it complete and useful,
then either
1. Make it possible to view the data header contents in the GUI,
or
2. Combine this output with the `metadata` dataset.

OR we could remove the entire `get_information` mechanism.

### LINKED ISSUE #811: [BUG] Jobs that output a trajectory don't write metadata
**Description of the error**
Every MDANSE job should write a dataset called "metadata" into the output. It contains the full set of the input parameters used to create the output file.

The input parameters should be written as "metadata" both in .mda and .mdt files.

CenterOfMassesTrajectory and TrajectoryEditor write out trajectories that do not contain the "metadata" group.

**Suggested fix**
At the moment the `Converter` class has a `write_metadata` method which is used for writing metadata into a trajectory. This method should be moved outside of the `Converter` class, so it can be used by other jobs that write trajectories, even if these jobs are not converters.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
The `HDFTrajectoryInputData` class appeared to be an unnecessary layer between the `Trajectory` class and the rest of the code using trajectories. This PR removes it, together with some other classes that do not seem to be needed for anything.

closes #695 
closes #730 
closes #807
closes #811

**Fixes**
1. Removed all classes from `MDANSE.IO` except for `MinimalPDBReader`.
2. Removed all classes from `MDANSE.Framework.InputData`.
3. Replaced different `Configurator.get_information` implementations with a common `IConfigurator.__str__` implementation.
4. Adjusted the GUI code to use `Trajectory` instead of `HDFTrajectoryInputData`.
5. Introduced a common `check_metadata` function to `MDANSE.Framework.Formats.HDFFormat`, which is now used in the GUI for both trajectory and output data files.

**To test**
All the trajectory information displayed in the GUI should still be shown in the trajectory tab.
The plot selection tab should show additional information about the inputs of its analysis run. The 'metadata' section should now be followed by 'header' section.


MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -145,6 +154,16 @@ def __init__(self, name, **kwargs):
...
         self._original_input = ""
 
+    def __str__(self) -> str:
+        return "\n".join(
+            [""]
+            + [
+                f"{label}={str(getattr(self, label, 'Not set'))}"
+                for label in self._printable_attributes
+            ]
+            + [f"{key}={str(self.get(key, 'Not set'))}" for key in self.keys()]
```
oerc0122: Maybe a usecase for [`value_chain`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.value_chain) is this more readable?
```suggestion
        return "\n".join(
            value_chain(
                "",
                (
                    f"{label}={str(getattr(self, label, 'Not set'))}"
                    for label in self._printable_attributes
                ),
                (f"{key}={str(self.get(key, 'Not set'))}" for key in self),
            )
        )
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/TrajectoryInfo.py
```diff
@@ -20,10 +20,68 @@
...
+    metadata = check_metadata(traj.file)
+    for k, v in metadata.items():
+        val.append(f"{k}: {v}")
+
+    mol_types = {}
+    val.append("\nMolecular types found:")
+    for molname, mollist in traj.chemical_system._clusters.items():
+        val.append(f"Molecule: {molname}; Count: {len(mollist)}")
+
+    for k, v in mol_types.items():
```
oerc0122: `mol_types` is empty?

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
```diff
@@ -28,6 +29,56 @@
...
+
+    Returns
+    -------
+    dict[str, str]
+        dictionary of saved input information used to create the file
+
+    """
+    meta_dict = {}
+
+    def put_into_dict(name, obj):
```
oerc0122: Needs type-hints and a docstring. Not sure what this is trying to do.

oerc0122 (changes requested):
Looks like a huge improvement in legibility and maintenance. Just a couple of minor things

MBartkowiakSTFC:
I implemented the changes you recommended.

Also, I noticed that TrajectoryEditor and CentreOfMassesTrajectory were not writing the 'metadata' group into the output file, since that was normally done by a method of `Converter`. Following the reasoning of this PR, I moved `write_metadata` to `MDANSE.Framework.Formats.HDFFormat` so it can be used by converters and other jobs that write trajectories. (This will be relevant to the TrajectoryFilter as well, once merged.)

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -145,6 +159,18 @@ def __init__(self, name, **kwargs):
...
 
+    def __str__(self) -> str:
+        return "\n".join(
+            value_chain(
+                "",
+                [
+                    f"{label}={str(getattr(self, label, 'Not set'))}"
+                    for label in self._printable_attributes
+                ],
+                [f"{key}={str(self.get(key, 'Not set'))}" for key in self.keys()],
```
oerc0122: Shouldn't need to instantiate lists here, also, if this derives from `dict`, `keys` is the default iterable
```suggestion
                (
                    f"{label}={str(getattr(self, label, 'Not set'))}"
                    for label in self._printable_attributes
                ),
                (f"{key}={str(self.get(key, 'Not set'))}" for key in self),
```

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
```diff
@@ -28,6 +29,103 @@
...
+        return
+    else:
+        meta.visititems(put_into_dict)
+
+    meta_dict["<b>file header</b>"] = "\n" + hdf5_file.attrs.get("header", "no header")
+
+    return meta_dict
+
+
+def write_metadata(job: "IJob", output_file: h5py.File):
```
oerc0122, line 97: This seems like it might want to be a method on `IJob`, though I'm not 100% sure since it makes sense here too...

oerc0122 (approved):
Just a stylistic thing which may not be correct. 

Otherwise I'm happy.

Issue #438, opened 2024-05-21
[BUG] Unexpected normalisation of DOS results


MBartkowiakSTFC:
**Description of the error**
I have run the Density of States on a trajectory of SrTiO3. Here are the results, depending on the atom selection:
![SrTiO3_DOS_allSrAtoms](https://github.com/ISISNeutronMuon/MDANSE/assets/108934199/ac000eef-ec95-47b5-b784-ed1fb95a0e4c)
Each one of those curves has been calculated for ALL the Sr atoms. The selection was:
1. blue: all atoms (Sr, Ti, O)
2. orange: all Sr and Ti atoms
3. green: all Sr and 4 Ti atoms,
4. red: all Sr and 1 Ti atom.

Similarly, the Ti results are not what I expected:
![SrTiO3_DOS_differentTiAtoms](https://github.com/ISISNeutronMuon/MDANSE/assets/108934199/1fc59b4b-959a-4351-8f53-e8cd491d0ce7)
By deselecting the O atoms I made the Ti signal increase.

**Describe the expected result**
I would prefer if the result for Sr atoms did not change when the selection of other atoms is changed.

**Describe the actual result**
At the moment, changing anything in the atom selection affects all the results.

**Suggested fix**
In the first approximation, we could normalise the results to ALL the atoms in the system, and not just the selected ones.

**Additional details**
N/A

ChiCheng45:
I guess the weight scheme should be like this then

![image](https://github.com/ISISNeutronMuon/MDANSE/assets/14276033/67480d1a-255c-401c-b6b8-9a1d63c546c4)

where $a$ is the atom, $i$ is the index of an atom $a$ and $n_i$ is whether the atom with index $i$ is selected or not.

Currently I think its something like this

![image](https://github.com/ISISNeutronMuon/MDANSE/assets/14276033/48121c16-2d2a-483a-9da9-b8181087db43)

so the selection changes the normalization.

MBartkowiakSTFC:
Yes, this would be my preference. Probably not a big deal since we are working with arbitrary units here, but still it would make the results more consistent.

PR #810, opened 2025-05-30
Reimplement grouping as analysis input - 2

### LINKED ISSUE #644: [ENHANCEMENT] Combine atom selection and grouping

### LINKED ISSUE #680: [BUG] Inconsistent MSD total when atom or molecule grouping level is used

ChiCheng45:
**Description of work**
Grouping has now been implemented and will be applied to all jobs that use the weights configurator and other places where it makes sense to have grouping.

**Fixes**
closes #644
closes #680
Fixed intramolecular TCF result. See typo in protos below.
https://github.com/ISISNeutronMuon/MDANSE/blob/d9f00436b13464b70d6a73e1103eaab5efa9c801/MDANSE/Src/MDANSE/Framework/Jobs/PairDistributionFunction.py#L160-L172

**Explanation**
Here is a short explanation of how grouping works with the weight scheme. The total intermediate scattering function is

```math
F_{\mathrm{inc}, \mathrm{tot}}(q, t) = \frac{1}{N} \sum_{j} w_j \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle 
```

when we use the atom grouping, the above sum is split up by atom.

```math
F_{\mathrm{inc}, \mathrm{tot}}(q, t) = \sum_{\alpha} W_{\alpha} F_{\mathrm{inc}, \alpha}(q, t) 
```

```math
F_{\mathrm{inc}, \alpha}(q, t) = \frac{1}{Nc_{\alpha}} \sum_{j \in \alpha} \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle 
```

For the molecular case, we split the summation into atomic parts associated with each molecule. For a water and ethanol system, we would get 

```math
 F_{\mathrm{inc}, \mathrm{tot}}(q, t) = c_{\mathrm{H2O}}F_{\mathrm{inc}, \mathrm{H2O}}(q, t) +c_{\mathrm{EtOH}} F_{\mathrm{inc}, \mathrm{EtOH}}(q, t) 
```

where

```math
c_{\mathrm{H2O}} F_{\mathrm{inc}, \mathrm{H2O}}(q, t) = W^{\mathrm{H2O}}_{\mathrm{H}}F_{\mathrm{inc}, \mathrm{H}}^{\mathrm{H2O}}(q, t) + W^{\mathrm{H2O}}_{\mathrm{O}}F_{\mathrm{inc}, \mathrm{O}}^{\mathrm{H2O}}(q, t)
```

and

```math
c_{\mathrm{EtOH}}F_{\mathrm{inc}, \mathrm{EtOH}}(q, t) = W^{\mathrm{EtOH}}_{\mathrm{H}}F_{\mathrm{inc}, \mathrm{H}}^{\mathrm{EtOH}}(q, t) + W^{\mathrm{EtOH}}_{\mathrm{C}}F_{\mathrm{inc}, \mathrm{C}}^{\mathrm{EtOH}}(q, t) + W^{\mathrm{EtOH}}_{\mathrm{O}}F_{\mathrm{inc}, \mathrm{O}}^{\mathrm{EtOH}}(q, t)
```

The intermediate scattering functions are

```math
F_{\mathrm{inc}, \mathrm{H}}^{\mathrm{H2O}}(q, t) = \frac{1}{Nc^{\mathrm{H2O}}_{\mathrm{H}}} \sum_{j \in (\mathrm{H2O} \cap \mathrm{H})} \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle
```
and 

```math
W^{\mathrm{H2O}}_{\mathrm{H}} = c_{\mathrm{H}}^{\mathrm{H2O}} w_{\mathrm{H}}
```

where $c_{\mathrm{H2O}} = N_{\mathrm{H2O}} / N$, $c_{\mathrm{H}}^{\mathrm{H2O}} = N_{\mathrm{H}}^{\mathrm{H2O}} / N$, $N_{\mathrm{H2O}}$ is the number of atoms in water, and $N_{\mathrm{H}}^{\mathrm{H2O}}$ is the number of hydrogen atoms in water. 

**Jobs that deviate from the above mechanism**
 - RootMeanSquaredDeviation
RMSD doesn't follow the standard way in which the total results are summed up from individual components. The total results are not a sum or weighted sum of the individual components. RMSD results with grouping (e.g., `rmsd_[H2_O1]_all`) follow how the `rmsd_all` results are calculated, but with the atoms associated with a particular molecule. So `rmsd_[H2_O1]_all` will be the same as `rmsd_all` but with only `H2_O1` molecules selected.

- RootMeanSquaredFluctuation
Grouping for this job has been changed to `each atom` or `each molecule` just to make it clear that this job doesn't follow the same grouping mechanism explained above. The grouped results in this job are unchanged from what occurs on protos.

**Other jobs which could have a grouping**
- DensityProfile
I think grouping could be added to the DensityProfile job since it uses atom selection, atom transmutation, and weights. I have decided to leave it until #820 is resolved.
- Density
This job doesn't even split results into atomic parts yet, but I don't see why it couldn't be.

**All updated jobs**
- VanHoveFunctionDistinct
- DensityOfStates
- MeanSquaredDisplacement
- PositionAutoCorrelationFunction
- PositionPowerSpectrum
- VanHoveFunctionSelf
- VelocityAutoCorrelationFunction
- CurrentCorrelationFunction
- DynamicCoherentStructureFactor
- DynamicIncoherentStructureFactor
- ElasticIncoherentStructureFactor
- GaussianDynamicIncoherentStructureFactor
- StaticStructureFactorFromScatteringFunction
- NeutronTotalDynamicStructureFactor
- CoordinationNumber
- PairDistributionFunction
- StaticStructureFactor
- XRayStaticStructureFactor
- RootMeanSquaredDeviation
- RootMeanSquaredFluctuation

**Atom selection and transmutation**
Atom selection and transmutation work as before. With grouping atoms maintain their group after selection and transmutation so that if we have water molecules and select only the hydrogen atoms they will still be associated with the water molecule group. Similarly if we transmute the waters hydrogen atoms to deuterium or any other atoms, those atom will still be associated with the water molecule group.

**Example**
Here I have the DFTB water trajectory which I have transmuted some oxygen atoms to sulfur, so I have two different molecules, H2O (57 molecules) and H2S (43 molecules). 

image

Here I run DOS with molecule grouping. My total DOS result split into `dos_[H2_O1]_total` and  `dos_[H2_S1]_total` which get split into their atomic components, `dos_[H2_O1]_H` and `dos_[H2_O1]_O` for H2O and `dos_[H2_S1]_H` and `dos_[H2_S1]_S` for H2S. The total DOS is the sum of the total molecular DOS.

image

And the `dos_[H2_O1]_total` is a sum of its atomic components `dos_[H2_O1]_H` and `dos_[H2_O1]_O`.

image

Here is the DISF for `f(q,t)_total`,`f(q,t)_[H2_O1]_total`, and `f(q,t)_[H2_S1]_total`.

image

From the explanation we had

```math
 F_{\mathrm{inc}, \mathrm{tot}}(q, t) = c_{\mathrm{H2O}}F_{\mathrm{inc}, \mathrm{H2O}}(q, t) +c_{\mathrm{EtOH}} F_{\mathrm{inc}, \mathrm{EtOH}}(q, t) 
```

We can also plot $F_{\mathrm{inc}, \mathrm{H2O}}(q, t)$ and $F_{\mathrm{inc}, \mathrm{H2S}}(q, t)$ instead of $c_{\mathrm{H2O}}F_{\mathrm{inc}, \mathrm{H2O}}(q, t)$ and $c_{\mathrm{EtOH}} F_{\mathrm{inc}, \mathrm{EtOH}}(q, t)$ by deselecting the `apply weight?` in the plotter to get

image

which gives you a result similar to a calculation where you selected those molecules (they are not exactly the same because the selection can change the weights normalisation factor #438)

Here is an example from the DOS. In the orange plot, only water molecules were selected and atom grouping was used. The blue is the total H2O results where I had all atoms selected with molecule grouping.

image

Now, if I normalise the data using the plotter, you can see they are the same; it's just that MDANSE's weight depends on the selection (#438).

image

**To test**
Run a few jobs with atom selection, transmutation and grouping for trajectories with molecules and check that results follow expectations.


MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -72,73 +73,63 @@ def configure(self, value):
...
-                                [chemical_system.atom_list[x] for x in cluster]
-                            )
-                            names.append(f"{group_name}_num{counter}_in_{clustername}")
-                            masses.append([mass_lookup[x] for x in cluster])
-                            counter += 1
+                    for x in cluster:
+                        if x not in atomSelectionConfig["flatten_indices"]:
+                            continue
+                        indices.append([x])
+                        flatten_indices.append(x)
```
oerc0122: I think this can be simplified, `more_itertools` provides the [`collapse`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.collapse) iterator which should iterate over the flattened indices.

Since it is subsequently stored, you might just be able to do
```python
flattened_indices = list(collapse(indices))
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -72,18 +81,59 @@ def configure(self, value):
...
 
-        if value == "atom":
-            return
+        self["level"] = value
+
         trajConfig = self._configurable[self._dependencies["trajectory"]]
         atomSelectionConfig = self._configurable[self._dependencies["atom_selection"]]
         chemical_system = trajConfig["instance"].chemical_system
+
+        if value == "atom" or value == "each atom":
```
oerc0122:
```suggestion
        if value in {"atom", "each atom"}:
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -72,18 +81,59 @@ def configure(self, value):
...
+                    group_n_atms[mol_name] = n_atms
+
+            self["name_to_element"] = {}
+            for name, element in zip(names, elements):
+                self["name_to_element"][name] = element[0]
+            self["group_names"] = sorted(set(group_names))
+            self["group_elements"] = group_elements
+            self["group_n_atms"] = group_n_atms
+
+        if value == "each molecule":
```
oerc0122: If they're mutually exclusive:
```suggestion
        elif value == "each molecule":
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -72,18 +81,59 @@ def configure(self, value):
...
+        group_elements = {}
+        group_n_atms = {}
         mass_lookup = chemical_system.atom_property("atomic_weight")
 
         if value == "molecule":
+            if len(trajConfig["instance"].chemical_system.unique_molecules()) == 0:
+                self.error_status = "The trajectory does not contain molecules."
+                return
+
+            for mol_name in chemical_system._clusters.keys():
```
oerc0122: If a `dict` `keys` is default iterator.
```suggestion
            for mol_name in chemical_system._clusters:
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -72,18 +81,59 @@ def configure(self, value):
...
+                        masses.append([mass_lookup[x]])
+                        n_atms += 1
+                if mol_selected:
+                    group_names.append(mol_name)
+                    group_elements[mol_name] = group_name_elements
+                    group_n_atms[mol_name] = n_atms
+
+            self["name_to_element"] = {}
+            for name, element in zip(names, elements):
+                self["name_to_element"][name] = element[0]
```
oerc0122:
```suggestion
            self["name_to_element"] = {name: element[0]
                                       for name, element in zip(names, elements)}
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
-                            indices.append(list(molecule.intersection(residue)))
-                            elements.append(
-                                [chemical_system.atom_list[x] for x in cluster]
-                            )
-                            names.append(f"{group_name}_num{counter}_in_{clustername}")
-                            masses.append([mass_lookup[x] for x in cluster])
-                            counter += 1
 
         atomSelectionConfig["indices"] = indices
+        atomSelectionConfig["flatten_indices"] = collapse(indices)
```
oerc0122: Note this is an iterator and will exhaust on multiple uses. Probably want to still make this a list or tuple.

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
 
-        :note: this is a static method
+    def add_grouped_totals(
+        self,
+        output_data: dict[str, npt.NDArray],
+        result_name: str,
+        data_type: str,
+        dim: int = 1,
+        conc_exp: float = 1.0,
+        intra: bool = False,
```
oerc0122, line 182: `bool`s and large argument counts of extra arguments are usually best expressed as kwarg only. 
```suggestion
        *,
        intra: bool = False,
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
+                )
+                if scaling_factor:
+                    output_data[
+                        f"{result_name}_[{grp_i}][{grp_j}]_{post_label}"
+                    ].scaling_factor = conc
+        else:
+            raise NotImplementedError("Grouped total for dim > 2 not implemented.")
+
+    def pair_labels(
+        self, intra=False, all_pairs=False
```
oerc0122:
```suggestion
        self, *, intra: bool = False, all_pairs: bool = False
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
+            for ele_i, ele_j in sorted(pairs):
+                pair_label = f"[{grp_i}][{grp_j}]_{ele_i}{ele_j}"
+                label_i = f"[{grp_i}]_{ele_i}"
+                label_j = f"[{grp_j}]_{ele_j}"
+                labels.append((pair_label, (label_i, label_j)))
+        return labels
+
+    def update_pair_results(
+        self,
+        calc_func: Callable[[str, str], Iterator[tuple[str, bool, npt.NDArray]]],
```
oerc0122: To be precise, calc_func only needs to return an `Iterable` (can be used in a for-loop, i.e. has `__iter__`) rather than an `Iterator` (supports `next`)
```suggestion
        calc_func: Callable[[str, str], Iterable[tuple[str, bool, npt.NDArray]]],
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
+            return
+
+        for grp_i, grp_j in it.combinations_with_replacement(self["group_names"], 2):
+            eles_i = sorted(set(self["group_elements"][grp_i]))
+            eles_j = sorted(set(self["group_elements"][grp_j]))
+            if grp_i == grp_j and not all_pairs:
+                iterable = it.combinations_with_replacement(eles_i, 2)
+            else:
+                iterable = it.product(eles_i, eles_j)
+            for ele_i, ele_j in sorted(iterable):
```
oerc0122: In general the `product` or `combinations` of a sorted set will be sorted. This materialises the iterable and has to store it, which may be very large.

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
+                label_j = f"[{grp_j}]_{ele_j}"
+                for name, intra, result in calc_func(label_i, label_j):
+                    if intra and grp_i == grp_j:
+                        output_data[f"{name}_[{grp_i}]_{ele_i}{ele_j}"][...] = result
+                    elif intra and grp_i != grp_j:
+                        continue
+                    else:
+                        output_data[f"{name}_[{grp_i}][{grp_j}]_{ele_i}{ele_j}"][
+                            ...
+                        ] = result
```
oerc0122:
```suggestion
                    if intra and grp_i != grp_j:
                        continue

                    if intra and grp_i == grp_j:
                        output_data[f"{name}_[{grp_i}]_{ele_i}{ele_j}"][...] = result
                    else:
                        output_data[f"{name}_[{grp_i}][{grp_j}]_{ele_i}{ele_j}"][
                            ...
                        ] = result
```

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -92,49 +142,280 @@ def configure(self, value):
...
+
+    def label_pairs(self, labels: list[str], all_pairs) -> list[tuple[str, str]]:
```
oerc0122:
```suggestion
    def label_pairs(self, labels: Iterable[str], *, all_pairs: bool) -> list[tuple[str, str]]:
```

MDANSE/Src/MDANSE/Framework/Jobs/CoordinationNumber.py
```diff
@@ -169,68 +144,69 @@ def finalize(self):
...
 
-            idi = self.selectedElements.index(at1)
-            idj = self.selectedElements.index(at2)
-
-            if idi == idj:
-                nij = ni**2 / 2.0
-            else:
-                nij = ni * nj
+        # symmetrize the data
+        for i, j in sorted(
```
oerc0122: Again, if these are pre-sorted, combinations will be.

MBartkowiakSTFC:
I am not sure if this should be the result:
1. I take a trajectory with 4 molecules (2 types):
Screenshot 2025-06-09 at 11 48 41
2. I select a single molecule:
Screenshot 2025-06-09 at 11 48 53
3. The VanHoveFunctionDistinct with molecule grouping on this selection shows that inter- and intramolecular results are both non-zero.
Screenshot 2025-06-09 at 11 50 04

When the selection consists of a single molecule, I would expect the "inter" part to be 0.

ChiCheng45:
I think the above issue is not related to this PR. There is something wrong with the trajectory or how the inter- or intra-molecular components are worked out. I've used the above trajectory and calculated the PDF with only the C60 selected on protos (8947373572069c1f3baaedfa744afa09eb810c39 because the plotter doesn't work on the latest commit).

image

Running the PDF shows non-zero intra- and intermolecular parts.

image

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -125,20 +124,20 @@ def weighted_sum(
...
 
     Returns
     -------
     np.ndarray
-        total sum of all the component arrays scaled by their weights
+        Total sum of all the component arrays scaled by their weights
     """
-    weightedSum = 0.0
-    matches = {key % k for k in weights if k not in ["sum"]}
+    matches = [match_key % val[0] for val in match_labels]
```
oerc0122: Set contains should be faster.
```suggestion
    matches = {match_key % val[0] for val in match_labels}
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -77,7 +78,8 @@ def get_weights(
 def assign_weights(
     values: dict[str, np.ndarray],
     weights: dict[str, float],
-    key: str,
+    match_key: str,
+    match_labels: Iterable,
```
oerc0122:
```suggestion
    match_labels: Iterable[tuple[str, ?]],
```

MDANSE/Src/MDANSE/Framework/Jobs/XRayStaticStructureFactor.py
```diff
@@ -173,79 +227,79 @@ def finalize(self):
...
+                    ele[0],
                     self._outputData["q"],
                     self.configuration["trajectory"]["instance"],
                 ),
             )
-            for k in list(nAtomsPerElement.keys())
+            for name, ele in zip(
+                self.configuration["atom_selection"]["names"],
+                self.configuration["atom_selection"]["elements"],
+            )
```
oerc0122:
```python
    asf = {name: atomic_scattering_factor(
               ele[0],
               self._outputData["q"],
               self.configuration["trajectory"]["instance"],
           )
           for name, ele in zip(
                self.configuration["atom_selection"]["names"],
                self.configuration["atom_selection"]["elements"],
            )
        }

File: MDANSE/Src/MDANSE/Framework/Jobs/XRayStaticStructureFactor.py [hunk 17]
```diff
@@ -130,41 +140,85 @@ def finalize(self):
...
+                The element label.
+            label_j : str
+                The element label.
+
+            Yields
+            ------
+            tuple[str, bool, np.ndarray]
+                A tuple of the results name, a bool specifying whether
+                results correspond to intermolecular atom pairs, and
+                the results.
```
oerc0122: Numpydoc allows for:
```suggestion
            name : str
                Results name
            inter : bool
                Whether results are intermolecular atom pairs.
            results : np.ndarray
                The results.
```

MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py
```diff
@@ -90,6 +89,11 @@ def initialize(self):
 
         self.numberOfSteps = self.configuration["atom_selection"]["selection_length"]
 
+        self.labels = [
+            (element, (element,))
+            for element in self.configuration["atom_selection"].get_natoms()
+        ]
```
oerc0122, line 95: Seeing this a lot, does this just want to become a method where it returns 
```python
list(
    zip(
        self.configuration["atom_selection"].get_natoms(),
        combinations_with_replacement(
          
  self.configuration["atom_selection"].get_natoms(), 
            n=1
        )
    )
)
```
 for this style, n=2 for pairwise?

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareDeviation.py
```diff
@@ -150,20 +150,27 @@ def finalize(self):
...
+            post_func=lambda x: np.sqrt(x / n_atms),
+            post_label="all",
+            main_result=True,
+        )
 
+        nAtomsPerElement = self.configuration["atom_selection"].get_natoms()
         for element, number in nAtomsPerElement.items():
             self._outputData[f"rmsd_{element}"][:] = np.sqrt(
-                self._outputData[f"rmsd_{element}"]
+                self._outputData[f"rmsd_{element}"] / number
```
oerc0122, line 171: I realise it's in the original, but `number` is a little vague. `natoms`? `n_elem`?

oerc0122 (commented):
All looks reasonable and well tested, just a couple of stylistic things.

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareDeviation.py
ChiCheng45: This is done in a lot of places in MDANSE e.g.

https://github.com/ISISNeutronMuon/MDANSE/blob/93a56c72bd5c9bb17062fbc370a0256d3eab5251/MDANSE/Src/MDANSE/Framework/Jobs/DensityOfStates.py#L252-L254

so I think it's better to leave it, unless we decide to change it across the entire codebase. Also, I think it should be clear what `number` is referring to since it's iterating over `nAtomsPerElement.items()`.

MDANSE/Src/MDANSE/Framework/Jobs/PositionAutoCorrelationFunction.py
ChiCheng45: The n=2 case is more complicated; we have 

```python
    def pair_labels(
        self, *, intra: bool = False, all_pairs: bool = False
    ) -> list[tuple[str, tuple[str, str]]]:
```

in groupinglevelconfigurator for that. For the n=2 case, it needs to be corrected for intra-molecular labels and other things when required.

MDANSE/Src/MDANSE/Framework/Jobs/PositionPowerSpectrum.py
```diff
@@ -246,25 +261,49 @@ def finalize(self):
...
             "pps_%s",
+            self.labels,
+        )
+        self.configuration["grouping_level"].add_grouped_totals(
+            self._outputData,
+            "pacf",
+            "LineOutputVariable",
+            axis="time",
+            units="nm2",
+            main_result=True,
```
MBartkowiakSTFC: I think we should be more conservative with assigning the "main result" status to datasets. For PositionPowerSpectrum, I would not consider the position autocorrelation function to be the main result.

MBartkowiakSTFC:
Regarding the `main_result` and `partial_result` tags with molecule grouping, maybe we should say that:
- {something}_total is the **main** result
- {something}_{molecule}_total is **main** and **partial**

and for finer subdivisions the users will have to plot the results manually. I am just worried that the fast plot will become less convenient if it includes too many datasets.

ChiCheng45:
> Regarding the `main_result` and `partial_result` tags with molecule grouping, maybe we should say that:
> 
> * {something}_total is the **main** result
> * {something}_{molecule}_total is **main** and **partial**
> 
> and for finer subdivisions the users will have to plot the results manually. I am just worried that the fast plot will become less convenient if it includes too many datasets.

I've updated the code to follow the above. I also found that some datasets were set to a main results when they shouldn't have which I have now changed.

MBartkowiakSTFC (approved):
Thanks for the changes. I think this is fine now.

PR #812, opened 2025-06-02
Add optional file widget

oerc0122:
**Description of work**
- Add new `OptionalFileInputWidget` which does not `configure` its `Configurator` until the button is pressed. 

**Fixes**
Adds new Widget available for use with configurators.

**To test**
No tests as unused so far.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Widgets/OptionalInputFileWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Widgets/OptionalInputFileWidget.py
@@ -1,0 +1,57 @@
+#    This file is part of MDANSE_GUI.
+#
+#    MDANSE_GUI is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+from qtpy.QtCore import Slot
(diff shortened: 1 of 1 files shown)


MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/OptionalInputFileWidget.py
```diff
@@ -0,0 +1,58 @@
...
+from MDANSE_GUI.InputWidgets.InputFileWidget import InputFileWidget
+
+
+class OptionalInputFileWidget(InputFileWidget):
+    """Optional file doesn't need to care if underlying file is valid until you try to load."""
+
+    def __init__(self, *args, button_label: str, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.button_label = button_label
```
MBartkowiakSTFC: At the moment, the parent `__init__` method calls the `add_widgets_to_layout()`, so the `self.button_label` would have to be set _before_ calling `super().__init__(*args, **kwargs)`.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/OptionalInputFileWidget.py
```diff
@@ -0,0 +1,58 @@
...
+from qtpy.QtCore import Slot
+from qtpy.QtWidgets import QPushButton
+
+from MDANSE_GUI.InputWidgets.InputFileWidget import InputFileWidget
+
+
+class OptionalInputFileWidget(InputFileWidget):
+    """Optional file doesn't need to care if underlying file is valid until you try to load."""
+
+    def __init__(self, *args, button_label: str, **kwargs):
```
MBartkowiakSTFC: While this widget will work in general, the current way of creating widgets in `Action.py`:
https://github.com/ISISNeutronMuon/MDANSE/blob/8947373572069c1f3baaedfa744afa09eb810c39/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py#L276
will make it difficult to pass the `button_label` if we wanted to use this widget the same way the other InputWidgets are being used at the moment.

MBartkowiakSTFC (approved):

PR #814, opened 2025-06-03
Atom selection saving and loading

### LINKED ISSUE #709: [BUG] Clicking in 3D view does not work with the new atom selection
**Description of the error**
The new atom selection mechanism in #672 does not work with atom selection in 3D view.

Atoms should be added/removed from selection when they have been clicked.

The selection does not get updated.

**Suggested fix**
Clicking on an atom should create a new selection operation in the sequence. Multiple atoms could be included in the same operation, as long as it is the last in the sequence.

**Additional details**
Eventually we will also need a way to select an entire molecule by clicking on a single atom.

### LINKED ISSUE #774: [BUG] Charges get reset unintentionally in Infrared helper dialog
**Description of the error**
The same button in the partial charge helper resets the atom selection AND the charges.

Two buttons should exist, one for resetting the selection, and the other for resetting the charges.
It should be possible to select a subset of atoms, assign a charge to them, and then reset the selection while keeping the charges set.

If I select atoms of one type and assign the charge to them, I need to reset the selection to select another atom type. If I reset it, the charges that had already been set turn back to their original values.

**Suggested fix**
Introduce a separate button for resetting the charges, with a clear label.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Atom selection now can be loaded both from an .mda file and from a dedicated JSON file. Saving to json is also possible in the atom selection dialog.

closes #709
closes #774

**Fixes**
1. AtomSelectionHelper dialog has a "save selection" button, which saves the selection as a JSON file.
2. AtomSelectionWidget uses QListView and shares the selection mode with the helper dialog. There is no need to apply the selection in the helper dialog.
3. The dialogs for atom selection, atom transmutation and partial charge setting respond to atom picking in the 3D view.
4. Manual selection of atoms continues until the user adds another selection operator, closes the dialog, saves the selection, or applies the selection using a button.
5. A placeholder item is added to the selection sequence when the user starts selecting atoms manually. This item is either replaced with a real selection operation, or removed if the user rejects the changes.
6. Transmutation reset and partial charge reset are now separate from selection reset.

**To test**
You should be able to:
- create a selection and run an analysis,
- load the same selection from the analysis results,
- save the selection to a json file and load it from there.


MDANSE/Src/MDANSE/Framework/AtomSelector/general_selection.py
```diff
@@ -76,3 +78,39 @@ def invert_selection(
...
+    """
+    not_selected = select_all(trajectory) - current_selection
+    click_counter = Counter(clicked_atoms)
+    flipped = set()
+    for index, state in click_counter.items():
+        if state % 2:
+            flipped.add(index)
+    removed = current_selection & flipped
+    added = not_selected & flipped
+    return (current_selection - removed) | added
```
oerc0122: Think this could be simplified.
```suggestion
    click_counter = Counter(clicked_atoms)
    # Add current selection count as True
    click_counter.update(current_selection)
    return {index for index, state in click_counter.items() if state % 2}
```

MDANSE/Src/MDANSE/Framework/AtomSelector/group_selection.py
```diff
@@ -42,7 +44,10 @@ def select_labels(
 
     """
     system = trajectory.chemical_system
-    return {system._labels[label] for label in atom_labels if label in system._labels}
+    result = set()
+    for label in atom_labels:
+        result |= set(system._labels.get(label, []))
+    return result
```
oerc0122:
```suggestion
    return {struc 
            for label in atom_labels 
            for struc in system._labels.get(label, []))}
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -233,3 +249,56 @@ def load_from_json(self, json_string: str):
...
+    def load_from_hdf5(self, filename: str):
+        """Load selection from an HDF5 output file (MDA format).
+
+        Parameters
+        ----------
+        filename : str
+            path to an MDA file, given as string
+
+        """
+        source = h5py.File(filename)
```
oerc0122: Should also be able to use `h5py.File` as context manager (i.e. `with`)

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
```diff
@@ -233,3 +249,56 @@ def load_from_json(self, json_string: str):
...
+        try:
+            byte_string = source["metadata/inputs/atom_selection"][0]
+        except KeyError as err:
+            LOG.error(f"atom selection string not found in file {filename}")
```
oerc0122: Given people might(?) have old files, do we want to instead return `{}`?

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -457,6 +575,37 @@ def helper_dialog(self) -> None:
...
+        fname = QFileDialog.getOpenFileName(
+            self._base,
+            "Load selection from a file (JSON or MDA)",
+            str(self._trajectory_path),
+            "MDANSE selection files (*.mda *.json);;HDF5 files (*.h5);;HDF5 files(*.hdf);;All files(*.*)",
+        )[0]
+        if fname is None:
+            return
+        if len(fname) < 1:
+            return
```
oerc0122:
```suggestion
        if not fname:
            return
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -457,6 +575,37 @@ def helper_dialog(self) -> None:
...
+        temp_selection = ReusableSelection()
+        try:
+            temp_selection.load_from_hdf5(fname)
+        except OSError:
+            LOG.warning("File %s could not be read as an HDF5 file", fname)
```
oerc0122: Is this a `warning` or just an `info`? I think the only `warning` should be the failure to load.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomTransmutationWidget.py
```diff
@@ -160,10 +175,26 @@ class AtomTransmutationWidget(AtomSelectionWidget):
...
     _push_button_text = "Atom transmutation helper"
     _default_value = "{}"
-    _tooltip_text = "Specify the atom transmutation that will be used in the analysis. The input is a JSON string, and can be created using the helper dialog."
+    _tooltip_text = (
+        "Specify the atom transmutation that will be used in the analysis."
+        " The input is a JSON string, and can be created using"
+        " the helper dialog."
+    )
+
+    def __init__(self, *args, _use_list_view=True, **kwargs):
```
oerc0122:
```suggestion
    def __init__(self, *args, _use_list_view: bool = True, **kwargs):
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomTransmutationWidget.py
```diff
@@ -160,10 +175,26 @@ class AtomTransmutationWidget(AtomSelectionWidget):
...
+        """Create the main widget for transmuting atom types.
+
+        Parameters
+        ----------
+        _use_list_view : bool, optional
+            If True, a ListView will replace LineEdit, by default True
 
-    def create_helper(self, traj_data: tuple[str, Trajectory]) -> TransmutationHelper:
         """
+        super().__init__(*args, use_list_view=False, **kwargs)
```
oerc0122, line 195: Slightly concerned that this ignores the `_use_list_view` explicitly. Why is it in the args in that case?

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/PartialChargeWidget.py
```diff
@@ -155,10 +169,26 @@ class PartialChargeWidget(AtomSelectionWidget):
...
+        """Create the widget for setting atom charges.
+
+        Parameters
+        ----------
+        _use_list_view : bool, optional
+            ignored here. This widget always sets use_list_view=False
 
-    def create_helper(self, traj_data: tuple[str, Trajectory]) -> ChargeHelper:
         """
+        super().__init__(*args, use_list_view=False, **kwargs)
```
oerc0122, line 189: Again, if you want to solve this, I would put:
```suggestion
        if "use_list_view" in kwargs:
            raise TypeError(f"Cannot use list view with {type(self).__name__}."
        super().__init__(*args, use_list_view=False, **kwargs)
```

MDANSE/Src/MDANSE/Framework/AtomSelector/selector.py
MBartkowiakSTFC: Now it sets the output to an empty dictionary if no selection was found. Additionally, empty selections are ignored by the widget (i.e. it keeps the old selection if the one we are loading is empty) and a warning is logged.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomTransmutationWidget.py
MBartkowiakSTFC: I implemented your suggestion, with the added change that the AtomTransmutationWidget and PartialChargeWidget only raise an exception if they find `use_list_view=True` in the keywords, but they potentially accept `use_list_view=False`.

MDANSE/Src/MDANSE/Framework/AtomSelector/general_selection.py
```diff
@@ -76,3 +78,34 @@ def invert_selection(
...
+        a trajectory containing atoms to be selected
+    current_selection : Set[int]
+        set of indices that had been selected before manual selection
+    clicked_atoms : List[int]
+        list of atom indices that have been clicked so far
+
+    Returns
+    -------
+    Set[int]
+        set of all the indices in the trajectory which were not in the input selection
```
oerc0122:
```suggestion
    trajectory : Trajectory
        A trajectory containing atoms to be selected.
    current_selection : set[int]
        Set of indices that had been selected before manual selection.
    clicked_atoms : list[int]
        List of atom indices that have been clicked so far.

    Returns
    -------
    set[int]
        New selection after manual selection.
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
```diff
@@ -242,13 +310,38 @@ def create_buttons(self) -> list[QPushButton]:
...
+        or JSON text files.
+        """
+        self.selection_model.finalise_manual_selection()
+        fname = QFileDialog.getSaveFileName(
+            self,
+            "Save current selection to a JSON file",
+            str(self._trajectory_path),
+            "MDANSE selection files (*.json);;All files(*.*)",
+        )
+        if len(fname[0]):
```
oerc0122:
```suggestion
        if fname[0]:
```

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -118,7 +118,13 @@ class BasicSelectionWidget(QGroupBox):
 
     new_selection = Signal(str)
 
-    def __init__(self, parent=None, widget_label="Atom selection widget"):
+    def __init__(
+        self,
+        parent=None,
```
```suggestion
        parent: Optional[QWidget] = None,
```
?

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -173,6 +182,31 @@ def create_selection(self):
         self.new_selection.emit(json.dumps(funtion_parameters))
 
 
+class GUISelection(BasicSelectionWidget):
+    """Widget for confirming and cancelling manual selection in 3D view."""
+
+    def __init__(self, parent=None, widget_label="Manual Selection"):
```
```suggestion
    def __init__(
        self, 
        parent: Optional[QWidget] = None,
        *,
        widget_label: str = "Manual Selection"
    ):
```

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -173,6 +182,31 @@ def create_selection(self):
...
+        ----------
+        parent : QWidget, optional
+            parent in the Qt hierarchy, by default None
+        widget_label : str, optional
+            Text over the widget, by default "ALL ATOMS"
+
+        """
+        super().__init__(parent, widget_label=widget_label, add_standard_widgets=False)
+
+    def add_specific_widgets(self):
```
```suggestion
    def add_specific_widgets(self) -> None:
```

oerc0122 (approved):
Looks good to me, interface seems sensible and consistent. Seems to do its job! Just some stylistic/type-hinting things.

Issue #816, opened 2025-06-03
[BUG] Data view preview columns


oerc0122:
**Description of the error**
Preview lines beyond a certain point is ignored.

**Describe the expected result**
With a 10x4 array of data, setting `preview lines` to 40 should display all data. It is still truncated with a `...`. This is probably a result of the `str` method for `np.ndarray`s. 

**Describe the actual result**
```
# Dataset name: /vector_generator/shell_3/hkl_array
# from file /home/jacob/DynamicIncoherentStructureFactor_result1.mda
# Contains axes: vector_generator/coordinates, index1
# data unit is au
# first column is vector_generator/coordinates in units au
# first row is index1 in units N/A
0.0 0.0 1.0 2.0
0.0 -1.0 -1.0 -1.0
1.0 -1.0 0.0 0.0
2.0 0.0 -1.0 0.0
...
```

**Suggested fix**
Always print full arrays using a slice up to preview rows.

**Additional details**
Preview lines also seems to be off by 1. 

With preview lines in comments
```
# Dataset name: /vector_generator/shell_3/hkl_array
# from file /home/jacob/DynamicIncoherentStructureFactor_result1.mda
# Contains axes: vector_generator/coordinates, index1
# data unit is au
# first column is vector_generator/coordinates in units au
# first row is index1 in units N/A
0.0 0.0 1.0 2.0    # Shown when preview lines = 0
0.0 -1.0 -1.0 -1.0 # 1
1.0 -1.0 0.0 0.0   # 2
2.0 0.0 -1.0 0.0   # 3
...
```

oerc0122:
> The plotter expected the q vector arrays to be of the (N, 3) shape, but they are (3,N), which makes the fields "preview lines" and "preview columns" swap their meanings. This is probably the reason for #816 

_Originally posted by @MBartkowiakSTFC in https://github.com/ISISNeutronMuon/MDANSE/issues/817#issuecomment-2935927590_

oerc0122:
The issue might also be related to the fact that even when you are displaying all the data a `...` is printed at the end if `preview only` is ticked.

PR #817, opened 2025-06-03
Add ability to nest data in .mda files

### LINKED ISSUE #813: [ENHANCEMENT] Allow nested data in MDT/MDA
Currently `HDFFormat` replaces `/` in output names with `|` so it is not possible to add sub-elements to the data.

An easy way to nest data sensibly in output files.

N/A

**Additional context**
N/A

oerc0122:
**Description of work**
Adds ability to nest keys in .MDA (and .MDT) files. 

**Fixes**
Closes #813

**To test**
Run a job which writes QVectors and verify qvector data through plot holder.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/OutputFilesConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/OutputFilesConfigurator.py
@@ -105,7 +105,7 @@ class OutputFilesConfigurator(IConfigurator):
                 )
                 return
 
-            if fmt not in IFormat.subclasses():
+            if fmt not in IFormat.indirect_subclasses():
                 self.error_status = f"the output file format {fmt} is not registered as a valid file format."
                 return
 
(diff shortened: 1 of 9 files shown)


MBartkowiakSTFC:
This PR is nearly there. The structure of the q vector information is definitely better now.

I have three observations:
1. Now that the "metadata" group can be split into individual entries, the plotter can try to turn them into SingleDataset instances, which currently fails with a lot of error output, starting with
```
2025-06-03 16:11:42,918 - ERROR - process[52100] - main 32 - EXCEPTION:
<class 'KeyError'>
"Unable to synchronously open attribute (can't locate attribute: 'units')"
```
2. The plotter expected the q vector arrays to be of the (N, 3) shape, but they are (3,N), which makes the fields "preview lines" and "preview columns" swap their meanings. This is probably the reason for #816 
3. It turns out that the SingleDataset handles axis tags wrong if a real dataset is mixed with an 'index' axis, as it is the case now for the q vector arrays.
PlottingContext.py, line 114 is currently:
`                self._axes[axis_key] = np.arange(len(self._data))`
I think that the problem can be fixed if it is replaced with:
`                self._axes[axis_key] = np.arange(self._data.shape[ax_number])`

Point 3 can be addressed immediately in this PR. For the other two, you are welcome to decide if you want to do something about them here, or leave it to be fixed by another PR.

MBartkowiakSTFC (approved):
Looks good. Nested output makes the q-vector output much cleaner.

Issue #763, opened 2025-05-02
[ENHANCEMENT] Add support for NPT/NσT calculations


oerc0122:
**Is your feature request related to a problem? Please describe.**
NPT simulations will require extra support for variable cell calculations. 

NσT will require further consideration given that the algorithm presented in:
https://pubs.acs.org/doi/epdf/10.1021/acs.jctc.3c00308?ref=article_openPDF
seems to suggest it only applies to variable cell-lengths in potentially orthogonal systems.

**Describe the solution you'd like**
Full NPT support in the first instance.

<details><summary>Current jobs</summary>
<p>

- [ ] AngularCorrelation
- [ ] AreaPerMolecule
- [ ] AverageStructure
- [ ] CenterOfMassesTrajectory
- [ ] CoordinationNumber
- [ ] CroppedTrajectory
- [ ] CurrentCorrelationFunction
- [x] Density
- [ ] DensityOfStates
- [ ] DensityProfile
- [ ] DipoleAutoCorrelationFunction
- [ ] DistanceHistogram
- [ ] DynamicCoherentStructureFactor
- [ ] DynamicIncoherentStructureFactor
- [ ] Eccentricity
- [ ] ElasticIncoherentStructureFactor
- [ ] GaussianDynamicIncoherentStructureFactor
- [ ] GeneralAutoCorrelationFunction
- [ ] GlobalMotionFilteredTrajectory
- [ ] Infrared
- [ ] JobStatus
- [ ] McStasVirtualInstrument
- [ ] MeanSquareDisplacement
- [ ] MolecularTrace
- [ ] NeutronDynamicTotalStructureFactor
- [ ] OrderParameter
- [ ] PairDistributionFunction
- [ ] PositionAutoCorrelationFunction
- [ ] PositionPowerSpectrum
- [ ] RadiusOfGyration
- [ ] RigidBodyTrajectory
- [ ] RootMeanSquareDeviation
- [ ] RootMeanSquareFluctuation
- [ ] SolventAccessibleSurface
- [ ] StaticStructureFactor
- [ ] StructureFactorFromScatteringFunction
- [ ] Temperature
- [ ] TrajectoryEditor
- [ ] UnfoldedTrajectory
- [ ] VanHoveFunctionDistinct
- [ ] VanHoveFunctionSelf
- [ ] VelocityAutoCorrelationFunction
- [ ] Voronoi
- [ ] XRayStaticStructureFactor

</p>
</details> 

[...]

PR #819, opened 2025-06-05
Fix issues with density profile

### LINKED ISSUE #818: [BUG] Why does the density profile depend on the number of frames?
I've been looking at the density profile job, and I guess it should be something like the atom density along one of the unit cell axes or something. Is there a reference for this job?

The odd thing is that this job depends on the number of frames. Here, I run the density profile job on a liquid argon trajectory using 1001 and 500 frames. 

Image

The blue plot used 1001 frames, and the orange plot used 500 frames. This issue appears to be due to this part of the code in the density profile job.

https://github.com/ISISNeutronMuon/MDANSE/blob/d9f00436b13464b70d6a73e1103eaab5efa9c801/MDANSE/Src/MDANSE/Framework/Jobs/DensityProfile.py#L175-L177

where

https://github.com/ISISNeutronMuon/MDANSE/blob/d9f00436b13464b70d6a73e1103eaab5efa9c801/MDANSE/Src/MDANSE/Framework/Jobs/DensityProfile.py#L89-L90

This should be wrong, since analysis results should not depend on the number of frames of your trajectory.

oerc0122:
**Description of work**
- Actually average data over frames rather than adding(?)
- Remove scaling of extent with frames(?!)

**Fixes**
Fixes #818 

**To test**
Run 2 density profiles with `Ar_mdmc_h5md.h5` of different lengths (e.g. 5, 39), see how sharp peak spreads out as averaged over more steps, extent is preserved and values of similar scale.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/DensityProfile.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/DensityProfile.py
@@ -14,11 +14,7 @@
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-import collections
-
 import numpy as np
-
-
 from MDANSE.Core.Error import Error
 from MDANSE.Framework.Jobs.IJob import IJob
 from MDANSE.Mathematics.Arithmetic import assign_weights, get_weights, weighted_sum
@@ -31,7 +27,9 @@ class DensityProfileError(Error):
 class DensityProfile(IJob):
     """
     The Density Profile analysis shows the weighted atomic density heterogeneity in the directions of the simulation box axes.
+
     For a lipid membrane, the density variation in the direction perpendicular to the membrane is probed in reflectometry measurements.
+
     The Density Profile Analysis can show segregation or cluster order formation, for example during the formation of micelles.
     """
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
Looks good, the bug from #818 is fixed.

As far as I can tell, there will be other problems with this particular job, but for the moment the known problem has been fixed.

oerc0122:
> Looks good, the bug from #818 is fixed.
> 
> As far as I can tell, there will be other problems with this particular job, but for the moment the known problem has been fixed.

Which problems do you foresee? Is it worth fixing them now?

ChiCheng45 (approved):

oerc0122:
Foreseen problems have been raised in #820

ChiCheng45:
> > Looks good, the bug from #818 is fixed.
> > As far as I can tell, there will be other problems with this particular job, but for the moment the known problem has been fixed.
> 
> Which problems do you foresee? Is it worth fixing them now?

Looks good to me. I think there may be some issues with NPT/NσT, but I guess we have #763 open for this.

oerc0122:
Indeed. I'm not sure how much sense a density profile makes with NPT/NσT, you would have to do density as a proportion of the cell in that axis, which I'm not sure would produce meaningful results.

PR #825, opened 2025-06-09
Fixed plot holder for windows

### LINKED ISSUE #824: [BUG] Plot holder is not updating
**Description of the error**
When I click results in the plot creator, the dataset do not get loaded up, so I'm unable to plot anything. Could be a windows only issue.

**Additional details**

Image

ChiCheng45:
**Description of work**
Fixed the plot holder not loading data when clicked on the windows.

**Fixes**
Closes #824

**To test**
Check that the plot holder continues to work on Unix/Mac

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
@@ -16,7 +16,6 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from pathlib import Path
 from typing import TypeVar
 
 import h5py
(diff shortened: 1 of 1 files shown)


oerc0122 (commented):
While this works on Linux too, I feel we're somewhat stepping away from the clarity that `Path` should bring with this and personally prefer just falling back to naive `str`.

ChiCheng45:
> While this works on Linux too, I feel we're somewhat stepping away from the clarity that `Path` should bring with this and personally prefer just falling back to naive `str`.

Alright, I've updated it to use str. I also found it was better for the child_path to return an empty string when the data_parent is None, so it gets rid of the double forward slash in the plot holder.

image

MBartkowiakSTFC:
Overall, I think that the HDF5 filesystem analogy has been pushed too far here, and we should have never started using anything `Path`-related to find a dataset in a group inside an HDF5 file.

It would be nice if `h5py` provided some object for managing their 'path' inside their data structure. Is there such a thing?

oerc0122:
> Overall, I think that the HDF5 filesystem analogy has been pushed too far here, and we should have never started using anything `Path`-related to find a dataset in a group inside an HDF5 file.
> 
> It would be nice if `h5py` provided some object for managing their 'path' inside their data structure. Is there such a thing?

Toolz provides [`get_in`](https://toolz.readthedocs.io/en/latest/api.html#toolz.dicttoolz.get_in), but I don't think h5py does.

MBartkowiakSTFC:
I'm happy with the `str`-based solution.

oerc0122 (approved):

PR #827, opened 2025-06-09
Use a subset of molecule indices array for distance histograms

### LINKED ISSUE #826: [BUG] PDF: Intermolecular part is not zero for a single molecule
**Description of the error**
For a trajectory composed of multiple molecule types, the DistanceHistogram and VanHoveFunctionDistinct jobs can wrongly attribute atoms to separate molecules, even if they belong to the same one.

More details:
https://github.com/ISISNeutronMuon/MDANSE/pull/810#issuecomment-2955578212

MBartkowiakSTFC:
**Description of work**
The DistanceHistogram and VanHoveFunctionDistinct were using the full array mapping the atom index to the molecule index, even if the selection included only a subset of atoms.

closes #826 

**Fixes**
1. The lookup array is also limited to the elements from the atom selection.
2. Unit test has been added for the C60/cubane trajectory and a selection of a single C60.

**To test**
All tests must pass.


ChiCheng45 (approved):
Looks good. Seems like it is working correctly now.

Issue #450, opened 2024-05-30
[BUG] Numerical derivatives test: inaccuracies for some cases


ChiCheng45:
**Description of the error**
I've been testing the accuracies of the numerical derivatives with the cp2k water trajectory, this trajectory has velocity data so we can compare results. The timestep is 0.5 fs. Here is what I get for the temperatures.

![image](https://github.com/ISISNeutronMuon/MDANSE/assets/14276033/732d06c6-06d5-4a4e-9d38-c9cc8ee6db8e)

![image](https://github.com/ISISNeutronMuon/MDANSE/assets/14276033/a30a9382-753b-4b43-be43-427b59c5e3ff)

temperature_0: order=0
temperature_2: order=2
temperature_3: order=3
temperature_4: order=4
temperature_5: order=5

order 1 and order 2 are the same since #369.

The odd thing is that order > 2 always gives slightly larger temperatures, could this be a bug in the code? The most accurate is order=1 or 2. Note that the first (and also the last) frame has to be calculated with forward numerical derivatives, this gives larger errors in the temperatures with order=1 or 2. 

**Suggested fix**
Replace all numerical derivatives so that it uses numpy.gradient which is essentially order=2 in MDANSE doing this will also resolve #369 since we fix it to order=2. Maybe also drop the first and last frames, although this is probably unnecessary since the error will disappear when we time average something.

ChiCheng45:
Actually, I had a think and I realized that if you used something like the velocity verlet algorithm. The velocity should be equal to a central difference of the positions. This is probably why order=2 is the same as the velocity data from the MD simulation.

ChiCheng45:
Differences are probably expected. However, it is interesting to note that perhaps in some situations it might be better to recalculate the velocities with a higher order method even when MD velocity data is available assuming all frames from the MD simulation are used.

PR #829, opened 2025-06-10
Documentation update, part 2

MBartkowiakSTFC:
**Description of work**
This PR further reduces the amount of documentation, aiming to keep only the part that is still relevant to the current code. Additionally, some classes which are not used anymore have been deleted.

**Fixes**
- Added lists of all the converters, analysis jobs, and inputs they use,
- Reformatted docstrings of configurators and jobs,
- Deleted unused configurators,
- Removed documentation of unused jobs.

**To test**
Standard unit tests should pass.


Doc/docstrings_to_rst.py
```diff
@@ -0,0 +1,109 @@
...
+converter_page = []
+input_page = []
+
+def make_configurator_doc(conf: str, parent: str) -> str:
+    result = ""
+    temp = IConfigurator.create(conf, 'dummy')
+    result += f"\n.. _configurator-{parent}-{conf}:\n\n"
+    result += f"{conf}\n"
+    result += "".join(len(conf)*['-'])+'\n'
+    result += f"\ndefault={temp._default}\n\n"
```
oerc0122, line 23:
```suggestion
    nl = "\n"
    temp = IConfigurator.create(conf, 'dummy')
    return f"""
_configurator-{parent}-{conf}:

{conf}
{:-{len(conf)}s}

default={temp._default}

{nl.join(map(str.strip, temp.__doc__.splitlines())) if temp.__doc__ else f"bad docstring in {conf}"}
"""
```

MDANSE/Src/MDANSE/Framework/Jobs/Infrared.py
```diff
@@ -24,6 +24,13 @@
 
 
 class Infrared(IJob):
+    """Calculates the infrared spectrum of a system of molecules.
+
+    The infrared spectrum is a derivative of the dipole autocorrelation function.
```
ChiCheng45: might be slightly confusing.

"The infrared spectrum is a derivative of the dipole autocorrelation function."

Maybe change to "The infrared spectrum is the autocorrelation function of the time derivative of the molecule's dipole." or something else.

MDANSE/Src/MDANSE/Framework/Jobs/StaticStructureFactor.py
```diff
@@ -24,8 +24,9 @@
 
 
 class StaticStructureFactor(DistanceHistogram):
-    """
-    Computes the static structure factor from the pair distribution function for a set of atoms.
+    """Computes the static structure factor for a set of atoms.
+
+    The static structure factor is calculated from the pair distribution function.
     The result is defined as S(q) = F_coh(q, t = 0).
```
ChiCheng45: This was on the previous version but it's inaccurate. See https://mdanse.readthedocs.io/en/protos/pages/scattering.html#structure-factor-from-scattering-function. I 
think it would be best to put the eq. 97 in the documentation here or delete the line to avoid confusion.

MDANSE/Src/MDANSE/Framework/Jobs/PositionPowerSpectrum.py
```diff
@@ -25,9 +25,16 @@
...
-    """
-    Power spectrum (using Fast Fourier Transform) of atomic trajectories calculated from the Positional Autocorrelation Function (PACF).
-    This method provides a frequency-domain function for a filter (i.e. lowpass) to operate on, so that a desired range of atomic vibrational modes can be isolated.
+    """Calculates the position power spectrum.
+
+    Power spectrum (using Fast Fourier Transform) of atomic trajectories calculated
+    from the Positional Autocorrelation Function (PACF). The calculation result
+    is similar to the density of states, which is calculated from the velocity
+    autocorrelation function. In fact, PPS and DOS may show the same features
+    in systems where position and velocity are strongly correlated (i.e. solids).
```
ChiCheng45, line 34: No need to change the docstring. I'm writing this down for future reference. I think the PPS and DOS are related by a factor of $\omega^2$ because of the properties of the Fourier transform

```math
\mathcal{F} \left\{ \frac{\partial^2 }{ \partial t^2} f(t) \right\} = - \omega^2 \mathcal{F} \left\{  f(t) \right\}
```
 
 and from the fact that

```math
\langle r(0) v(t) \rangle = - \langle v(0) r(t) \rangle
```

since 

```math
```

MDANSE/Src/MDANSE/Framework/Configurators/OutputTrajectoryConfigurator.py
```diff
@@ -25,30 +25,25 @@
...
 class OutputTrajectoryConfigurator(IConfigurator):
-    """
-    This configurator allows to define the output directory, the basename, and the format(s) of the output file(s)
-    resulting from a trajectory conversion.
+    """Specifies how a trajectory should be output to a file.
 
-    Once configured, this configurator will provide a list of files built by joining the given output directory,
-    the basename and the  extensions corresponding to the input file formats.
+    Allows to define the path to the file, precision of the floating point numbers,
+    HDF5 chunk size and compression of the HDF5 datasets.
```
ChiCheng45: Also whether the log will be written out and its output log level.

MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
```diff
@@ -17,16 +17,17 @@
...
+    Normally it is used for calculating atom velocities from their positions.
+    Values from 1 to 5 are allowed. If exact velocities are provided in the
+    trajectory file, you can choose to use them by setting this to 0.
 
-    :attention: it is of paramount importance for the trajectory to be sampled with a very low time \
-    step to get accurate velocities interpolated from atomic coordinates.
+    The velocities calculated from atom positions have a tendency to be
+    underestimated compared to the real values in the simulation, and the
+    error in the calculation increases quickly with the size of the trajectory
+    time step.
```
ChiCheng45: Just one point about velocities from the trajectory. I don't know whether we should say "exact/real" since the velocities in molecular dynamics are also determined numerically.

Also whether the velocities from MDANSE are underestimated or overestimated compared to the velocities from the MD output. Won't that depend on the integration algorithm used and the order of the integration setting in MDANSE?

MDANSE/Src/MDANSE/Framework/Configurators/GroupingLevelConfigurator.py
```diff
@@ -28,22 +28,12 @@
...
-    * 'each molecule': this changes the atom selection so that the atom
-        indices for each molecule will be grouped together. Jobs can
-        then run calculations can be run for each group of indices
-        together.
+    """Define how the partial results will be grouped in the output.
+
+    'atom' - partial results for each chemical elements are output.
+    'molecule' - for each molecule type, a set of outputs will be created,
+      showing how different elements in that molecule contributed to the
+      total signal.
```
ChiCheng45, line 36: There is also "each atom" and "each molecule" it was removed for some reason.

MBartkowiakSTFC: Apparently Root Mean Square Fluctuation still has its own options, "each atom" and "each molecule". My preference would be to replace these custom entries with the standard ones, but we shouldn't do it in a documentation PR. For now, if we need to document "each atom" and "each molecule", I would prefer to have it documented in the section for Root Mean Square Fluctuation.

MDANSE/Src/MDANSE/Framework/Configurators/InstrumentResolutionConfigurator.py
```diff
@@ -23,28 +23,26 @@
...
 
-    where f(t) and r(t) are respectively the time-dependant signal and instrument resolution and
-    F(\omega) and R(\omega) are their corresponding spectrum. Hence, G(\omega) represents the signal
-    convoluted by the instrument resolution and, as such, represents the quantity to be compared directly with
-    experimental results.
-
-    An instrument resolution is represented in MDANSE by a kernel function and a sets of parameters for this function.
-    MDANSE currently supports the aussian, lorentzian, square, triangular and pseudo-voigt kernels.
+    where f(t) and r(t) are, respectively, the time-dependent signal and
+    instrument resolution. :math:`F(\omega)` and :math:`R(\omega)`
```
ChiCheng45: Typo on the math equations I guess it should be $\text{FT}(f(t) * r(t))$. Also, change f(t) and r(t) so they are in math font and I'm not sure $\otimes$ makes sense here, so it can be removed.

However, I'm confused. Does MDANSE actually do this? I think MDANSE does this:

```math
\text{FT}[f(t) r(t)] = F(\omega) * R(\omega) = G(\omega).
```

Doc/pages/dynamics.rst
```diff
@@ -445,3 +450,16 @@ the projected VACF
...
    {C_{\mathbf{vv}\alpha\alpha}{(\hat{\mathbf{q}}, \omega) = \frac{1}{2\pi}} \frac{1}{Nc_{\alpha}}\sum_{j \in \alpha} {\int\mathrm{d} t \,} C_{\mathbf{vv}jj}(\hat{\mathbf{q}}, t) e^{-i \omega t}}.
+
+Provided the VACF decays to zero at long time, the function may be
+integrated mathematically to calculate the diffusion coefficient D, as in:
+
+.. math:: 
+   :label: vacf6
+
+   D = \\frac{1}{3}\int_{0}^{+\infty}{<v(0) \cdot v(t)>dt}
+
```
ChiCheng45:
```suggestion
Provided the VACF decays to zero at long time, the function may be
integrated mathematically to calculate the diffusion coefficient :math:`D`, as in:

.. math:: 
   :label: vacf6

   D = \frac{1}{3}\int\limits_{0}^{\infty}\mathrm{d} t \, \langle \mathbf{v}(0) \cdot \mathbf{v}(t) \rangle.
```

MDANSE/Src/MDANSE/Framework/Jobs/AngularCorrelation.py
```diff
@@ -24,14 +24,14 @@
...
-    of the tail or chain. The correlation is calculated for the angle formed by the same vector at
-    different times
+    Origin and direction can either be an atom or a centre definition
+    (centre of a group of atoms). For example, the origin could be defined by the
+    geometric centre of the head group of a surfactant molecule and the direction
+    simply by the last atom of the tail or chain. The correlation is calculated for
+    the angle formed by the same vector at different simulation frames.
 
     **Calculation:** \n
     angle at time T is calculated as the following: \n
```
ChiCheng45, line 37: That maths that is generated from this docstring doesn't render properly.

image

MDANSE/Src/MDANSE/Framework/Jobs/MeanSquareDisplacement.py
```diff
@@ -22,29 +22,29 @@
...
+    completely random direction from the one before, as analyzed by Albert Einstein
     in a study of Brownian motion. The MSD of a particle in this case
     is proportional to the time elapsed:
 
     .. math:: <r^{2}> = 6Dt + C
 
-    where :math:`<r^{2}>` is the MSD and t is the time. D and C are constants. The constant D is
-    the so-called diffusion coefficient.
+    where :math:`<r^{2}>` is the MSD and t is the time. D and C are constants. The
+    constant D is the so-called diffusion coefficient.
```
ChiCheng45:
```suggestion
    .. math:: \langle d^{2}(t) \rangle = 6Dt + C

    where :math:`\langle d^{2}(t) \rangle` is the MSD and :math:`t` is the time. :math:`D` and :math:`C` are constants. The
    constant :math:`D` is the so-called diffusion coefficient.
```

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
```diff
@@ -37,8 +37,12 @@ class DynamicCoherentStructureFactorError(Error):
...
     """Computes the dynamic coherent structure factor S_coh(Q,w) for a set of atoms.
 
-    It can be compared to experimental data e.g. the energy-integrated, static structure factor S_coh(Q)
-    or the dispersion and intensity of phonons.
+    It can be compared to experimental data e.g. the energy-integrated, static structure
+    factor S_coh(Q) or the dispersion and intensity of phonons.
+
+    The coherent part is derived from correlations between pairs of atoms.
+    This analysis requires the Q vectors to be commensurate with the reciprocal lattice
+    of the simulation box.
```
ChiCheng45, line 45:
```suggestion
    """Computes the dynamic coherent structure factor :math:`S_{\text{coh}}(\mathbf{q}, \omega)` for a set of atoms.
    It can be compared to experimental data e.g. the energy-integrated, static structure
    factor :math:`S_{\text{coh}}(q)` or the dispersion and intensity of phonons.
    The coherent part is derived from correlations between pairs of atoms.
    This analysis requires the :math:`\mathbf{q}` vectors to be commensurate with the reciprocal lattice
    of the simulation box.
```

Just replacing the maths so that it uses maths font.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -25,9 +25,16 @@
...
-    """
-    Computes the dynamic incoherent structure factor S_inc(Q,w) for a set of atoms.
-        It can be compared to experimental data e.g. the quasielastic scattering due to diffusion processes.
+    """Computes the dynamic incoherent structure factor S_inc(Q,w) for a set of atoms.
+
+    It can be compared to experimental data e.g. the quasielastic scattering due to
+    diffusion processes.
+
+    This property is derived from the self-correlation of individual atoms over time.
+    While it does not require the Q vectors to be commensurate with the simulation
```
ChiCheng45:
```suggestion
    """Computes the dynamic incoherent structure factor :math:`S_{\text{inc}}(\mathbf{q},\omega)` for a set of atoms.

    It can be compared to experimental data e.g. the quasielastic scattering due to
    diffusion processes.

    This property is derived from the self-correlation of individual atoms over time.
    While it does not require the :math:`\mathbf{q}`-vectors to be commensurate with the simulation
```

update maths to used math font.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicCoherentStructureFactor.py
ChiCheng45:
```suggestion
    """Computes the dynamic coherent structure factor :math:`S_{\text{coh}}(\mathbf{q}, \omega)` for a set of atoms.

    It can be compared to experimental data e.g. the energy-integrated, static structure
    factor :math:`S_{\text{coh}}(q)` or the dispersion and intensity of phonons.

    The coherent part is derived from correlations between pairs of atoms.
    This analysis requires the :math:`\mathbf{q}`-vectors to be commensurate with the reciprocal lattice
    of the simulation box.
```

Just replacing the maths so that it uses maths font.

Doc/pages/analysis_jobs.rst
```diff
@@ -0,0 +1,1008 @@
...
+
+.. _analysis-reference-Converter:
+
+Converter
+~~~~~~~~~
+
+Outputs a trajectory in the MDT format.
+
+Inputs:
+
```
ChiCheng45: The abstract converter got added, can we remove this?

MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
```diff
@@ -22,7 +22,8 @@
 
 
 class Eccentricity(IJob):
-    """Computes the eccentricity for a set of atoms e.g. in a micelle.
+    """Computes the eccentricity of a selected set of atoms.
+
     The eccentricity is calculated from the principal moments of
     inertia via the equation sqrt(pm3**2 - pm1**2) / pm3 where pm1 and
     pm3 are the smallest and largest principal moments of inertia
```
ChiCheng45:
```suggestion
    The eccentricity is calculated from the principal moments of
    inertia via the equation :math:`\sqrt{\text{pm3}^{2} - \text{pm1}^{2}} / \text{pm3}` where :math:`\text{pm1}` and
    :math:`\text{pm3}` are the smallest and largest principal moments of inertia
```

Updating maths to use math font.

MDANSE/Src/MDANSE/Framework/Jobs/GaussianDynamicIncoherentStructureFactor.py
```diff
@@ -25,8 +25,12 @@
...
 
 class GaussianDynamicIncoherentStructureFactor(IJob):
-    """
-    Computes the dynamic incoherent structure factor S_inc(Q,w) for a set of atoms in the Gaussian approximation.
+    """Computes the dynamic incoherent structure factor in the Gaussian approximation.
+
+    Gaussian approximation is exact for a system of free particles and a system of
+    particles undergoing brownian motion. The results of this analysis will be close
+    to the Dynamic Incoherent Structure Factor analysis in the limits of very
+    short Q and very long Q, and will differ from it for intermediate Q values.
```
ChiCheng45:
```suggestion
    short :math:`\mathbf{q}` and very long :math:`\mathbf{q}`, and will differ from it for intermediate :math:`\mathbf{q}` values.
```

Update maths to used math font.

MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
```diff
@@ -28,8 +28,13 @@ class NeutronDynamicTotalStructureFactorError(Error):
 
 
 class NeutronDynamicTotalStructureFactor(IJob):
-    """
-    Computes the dynamic total structure factor for a set of atoms as the sum of the incoherent and coherent structure factors
+    """Combines the coherent and incoherent dynamic structure factors.
+
+    The partial results need to be calculated before using the Dynamic
+    Coherent/Incoherent Structure Factor jobs with the same Q vector setttings.
```
ChiCheng45:
```suggestion
    Coherent/Incoherent Structure Factor jobs with the same :math:`\mathbf{q}`-vector setttings.
```

Doc/pages/R_parameters.rst
ChiCheng45, line 1: The notation in this page needs to be updated so that it follows https://mdanse.readthedocs.io/en/protos/pages/notation.html

MDANSE/Src/MDANSE/Framework/Jobs/AngularCorrelation.py
MBartkowiakSTFC: This will be fixed if I update that page after we have merged #835.

MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
MBartkowiakSTFC: I could be wrong about the underestimated velocities. I started from these assumptions:
1. In the extreme case of a particle moving in a straight line, the velocity will not change with the size of the time step of trajectory sampling,
2. In the case of a curved trajectory, the length of the curve is approximated by a series of straight segments. The exact distance covered by a particle is reached in the limit of infinitely small sampling step. As the size of the sampling step increases (and the number of segments decreases), the sum of the segment length can only become lower than the actual curve length. Then, it appears that the particle covered a shorter distance in the same time, and so it had a lower velocity.
3. There is no way of drawing points on a curve and connecting the consecutive points with straight segments that would give a sum of segments longer than the actual length of the curve.

It is possible that my assumptions are wrong when higher orders of interpolation are used, but I still think that for any trajectory, as you increase the sampling step, you will reach a range where the temperature only decreases with the step size. I don't have any mathematical proof, though.

Here are some examples where temperature obtained by interpolation is lower than temperature from MD velocities, and increasing the step or lowering the interpolation order only makes the temperature lower:
cubane_temp
water_different_order

water_ord5_different_steps

ChiCheng45: Just a few things. In the first plot why is the order=5 step=1 temperatures so much lower than the order=0? When you dump the trajectory data are you saving every time step? Which integration algorithm are you using?

Some time ago I did some test with temperature calculations and numerical derivatives and found increased temperatures with increasing order setting. I also found that the velocity data from the MD to be equal to order=2 because if you use something like the velocity verlet algorithm, the velocities were effectively equivalent to an order=2 numerical derivative of the positions #450.

Now that I'm thinking about it, whenever a user uses numerical derivatives  and uses the "in steps of" setting >1 maybe we should actually do the derivatives with all time steps and then take the relevant velocities #853.

ChiCheng45 (approved):
I fixed some typos. Should be good to merge now.

PR #830, opened 2025-06-10
Updated to custom atom and property rename in database editor

### LINKED ISSUE #801: [BUG] Atom renaming in Element Editor renames wrong atom
**Description of the error**
When editing the properties of chemical elements, the user can create copies of elements and rename them afterwards. However, if custom atoms exist already in the database, and new atoms are added, renaming is not applied to the correct atom entry.

The first **new** atom should be renamed.

The first **custom** atom (added before) is renamed.

**Suggested fix**
Probably the data model index has to be determined differently.

**Additional details**
To reproduce the problem:
1. Create a new atom in the element editor.
2. Restart MDANSE. The new atom entry will now appear in the table, sorted alphabetically with the other atom entries.
3. Copy an existing atom a few times. New entries will appear at the bottom of the table.
4. Rename the first new copy. The name of the selected atom will not change. Instead, the atom created in step 1 will be renamed.

ChiCheng45:
**Description of work**
The atom database editor has been changed so that the renaming is done based on the row/column of the mouse position on right click.

image

**Fixes**
Closes #801

**To test**
Try copying and renaming custom atoms and properties. The renaming of the atom and properties will be based on the mouse click; check that this follows expectations.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/ElementsDatabaseEditor.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/ElementsDatabaseEditor.py
@@ -174,12 +174,11 @@ class ElementView(QTableView):
 
         data_model = self.parent().data_model
 
-        enable_rename_atoms = len(
-            set(ATOMS_DATABASE.atoms) - set(ATOMS_DATABASE.default_atoms_types)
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
It works, both for atoms and for properties. I did not find any problems.

PR #831, opened 2025-06-10
Atom selection normalisation update.

### LINKED ISSUE #438: [BUG] Unexpected normalisation of DOS results

ChiCheng45:
**Description of work**
PR #810 should be merged before this one. Atom selection has now been updated so that it is always a fraction of the results when all atoms are selected.

**Explanation**
Currently in MDANSE 

```math
F_{\mathrm{inc}, \mathrm{selected}}(q, t) = \frac{1}{N_{\text{selected}}} \sum_{j}^{N} n_{j}w_j \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle 
```

where $N$ is the number of atoms in the trajectory, $n_j$ is 0 or 1 depending on whether the atom has been selected or not. Additionally, the weights $w_j$ are normalised and are also dependent on the selection.

```math
w_j = \frac{u_j}{ \sum_{j}^{N} n_{j} u_j}
```

This update changes MDANSE so that the weights are no longer dependent on the selection and the selected results are always a fraction of the fully selected results

```math
F_{\mathrm{inc}, \mathrm{selected}}(q, t) = \frac{1}{N}\sum_{j}^{N} n_{j}w_j \langle \exp[-iq \cdot r_{j}(0)]\exp[iq \cdot r_{j}(t)] \rangle 
```
and the weight are now.

```math
w_j = \frac{u_j}{ \sum_{j}^{N} u_j}
```

**Examples**
- Running calculations by selecting all atoms of one type is now equivalent to the results of that element but with all atoms selected. Here I used this H2O and H2S trajectory.
image

image

- Similarly, the total H2O results with group and the total results with only H2O selected with atom grouping are the same.

image

- The sum of the results of a selection and its inverse is now equal to the total.

image
image

image

- Selected results are a fraction of the results with all atoms selected. The `Apply weights?` can be used on total results to obtain results scaled so that it is as if the system contained only those atoms. This is most useful for calculations like PDF. The normalization of the weights is maintained. Here I have the PDF but with only water selected.

image

The large r-values are less than one because they are a fraction of the total. I can use `Apply weights?` so that they are normalized.

image

Similarly with DISF, with `Apply weights?`, the sum of the partials is equal to the total. The total is selected so its a fraction of the total with all selected. 

image

We can uncheck `Apply weights?` so that they are normalized with f(q,t=0)=1.

image

**Fixes**
Fixes #438

**To test**
Run calculations with atom selection and without atom selection, and compare the results. Calculations with the selection should always be a fraction of the total with all atoms selected.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
@@ -108,7 +108,9 @@ class WeightsConfigurator(SingleChoiceConfigurator):
         self["property"] = value
         self.error_status = "OK"
 
-    def get_weights(self, prop: Optional[str] = None):
+    def get_weights(
+        self, *, prop: Optional[str] = None
+    ) -> tuple[dict[str, float], dict[str, float]]:
         """Generate a dictionary of weights.
 
         Parameters
@@ -119,30 +121,34 @@ class WeightsConfigurator(SingleChoiceConfigurator):
 
         Returns
         -------
-        dict[str, float]
+        tuple[dict[str, float], dict[str, float]]
             The dictionary of the weights.
         """
         if not prop:
             prop = self["property"]
 
-        atom_selection_configurator = self._configurable[
-            self._dependencies["atom_selection"]
-        ]
+        atm_select = self._configurable[self._dependencies["atom_selection"]]
 
-        weights = defaultdict(float)
-        for name, elements in itertools.islice(
-            zip(
-                atom_selection_configurator["names"],
-                atom_selection_configurator["elements"],
+        weights = []
+        for n_elements, atm_names, atm_elements in [
+            (atm_select.get_natoms(), atm_select["names"], atm_select["elements"]),
+            (
+                atm_select.get_all_natoms(),
+                atm_select["all_names"],
+                atm_select["all_elements"],
             ),
-            atom_selection_configurator["selection_length"],
-        ):
-            weights[name] += sum(
-                self._trajectory.get_atom_property(element, prop)
-                for element in elements
-            )
+        ]:
+            w = defaultdict(float)
+            for name, elements in itertools.islice(
+                zip(atm_names, atm_elements),
+                len(atm_names),
+            ):
+                w[name] += sum(
+                    self._trajectory.get_atom_property(element, prop)
+                    for element in elements
+                )
+            for element, num_atoms in n_elements.items():
+                w[element] /= num_atoms
+            weights.append(w)
 
-        for element, num_atoms in atom_selection_configurator.get_natoms().items():
-            weights[element] /= num_atoms
-
-        return weights
+        return tuple(weights)
--- a/MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
@@ -139,10 +139,7 @@ class WeightsConfigurator(SingleChoiceConfigurator):
             ),
         ]:
             w = defaultdict(float)
-            for name, elements in itertools.islice(
-                zip(atm_names, atm_elements),
-                len(atm_names),
-            ):
+            for name, elements in zip(atm_names, atm_elements):
                 w[name] += sum(
                     self._trajectory.get_atom_property(element, prop)
                     for element in elements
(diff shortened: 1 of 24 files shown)


MDANSE/Src/MDANSE/Framework/Configurators/WeightsConfigurator.py
```diff
@@ -119,30 +121,34 @@ def get_weights(self, prop: Optional[str] = None):
...
-        for element, num_atoms in atom_selection_configurator.get_natoms().items():
-            weights[element] /= num_atoms
-
-        return weights
+        ]:
+            w = defaultdict(float)
+            for name, elements in itertools.islice(
+                zip(atm_names, atm_elements),
+                len(atm_names),
+            ):
```
oerc0122: `zip` automatically zips shortest. So we don't need the `islice` here.
```suggestion
            for name, elements in zip(atm_names, atm_elements):
```

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -44,11 +54,59 @@ def get_weights(
...
+    )
+    _, normFactor = adjust_weights(all_props, all_contents, n_atms, dim, conc_exp)
+
+    normalise = True
+    try:
+        len(normFactor)
+    except TypeError:
+        normalise = abs(normFactor) > 0.0  # if normFactor is 0, all weights are 0 too.
+    if normalise:
+        for k in list(weights.keys()):
```
oerc0122: Because you aren't changing the keys (only the values), don't need to fix keys.
```suggestion
        for k in weights:
```

oerc0122 (commented):
Just a couple of things

MBartkowiakSTFC:
I just tried running the Static Structure Factor (SSF) for 20 CO2 molecules (60 atoms in total). I ran it several times:
- for all atoms
- for the first 30 atoms
- for the other 30 atoms.

The results from each half do not add up to the total result:
grouping_halves

Was this supposed to happen?

gonzalezma:
IMO yes. For any collective property, you cannot expect that the sum over two halves will be the same as the calculation over the full system. A simple example, if you compute the g(r) of a cluster containing 4 atoms you will get 6 peaks at different distances (12, 13, 14, 23, 24, 34), but if you do separate calculations for the pairs 1-2 and 3-4, you will only get 2 peaks. This will apply to the SSF, DCSF and any analysis involving a sum over pairs of atoms.

ChiCheng45:
> I just tried running the Static Structure Factor (SSF) for 20 CO2 molecules (60 atoms in total). I ran it several times:
> 
> * for all atoms
> * for the first 30 atoms
> * for the other 30 atoms.
> 
> The results from each half do not add up to the total result: grouping_halves
> 
> Was this supposed to happen?

@gonzalezma is right, they are not equal to the total because it's missing the cross terms between the first 30 and the other 30 atoms. Here is the h2o/h2s trajectory where I select only h2o or only h2s and also with all selected, but with molecule grouping so I can get the h2o/h2s cross terms. The sum of the three is equal to the total.

image

@MBartkowiakSTFC I've had a think and I also want to make a few changes to the PR so that users can use the `Apply weight?` box to the total result, so that they can still get the same total results as on protos with selection.

MBartkowiakSTFC:
Sounds good. Thanks to both of you for the explanation.

MBartkowiakSTFC (approved):
As far as I can tell, the results are consistent. Also, the option to undo the scaling of the total result is a good idea.

PR #835, opened 2025-06-13
AngularCorrelation input change

### LINKED ISSUE #399: [ENHANCEMENT] Update angular correlation and molecule related jobs
For the angular correlation job it would be nice to be able to select different molecular axis e.g. the molecules rotational axis. 
For all molecule related jobs it would be able to select more then one molecule.

Additionally I think that bulk structures should not be selectable in molecule related jobs.

### LINKED ISSUE #556: [BUG] Angular correlation results depend on the atom ordering of the molecule.
**Description of the error**
The angular correlation is dependent on the molecule atom ordering. Here I have created a CH4 molecule trajectory that simply rotates around the CH bond. 

Animation46

In this example, I calculate the AC where the atom is ordered so that the first atom is the carbon (similar results when the first atom is the CH hydrogen). 

image

Here is an example where I used one of the hydrogen which is not the CH hydrogen we are rotating around.

image

The AC dependency on the atom order will be an issue if the same molecules are different in the same trajectory.

**Suggested fix**
We need a scheme that is not dependent on the atom ordering of the molecule. We could use the Kabsch algorithm to determine the optimal rotation matrix to map the molecule at time t1 to the molecule at time t0 we can then obtain that angle and use this to determine the angular correlation function. This will be a more computationally expensive approach but will be much more robust.

MBartkowiakSTFC:
**Description of work**
The molecule axis which is analysed in AngularCorrelation is now defined explicitly as one of the inputs.

Also, AngularCorrelation has been replaced by Reorientational Time Correlation Function, which uses the information about the orientation of the specified vector in a molecule to calculate calculates the average of different-order Legendre polynomials of the cosine of the angle between the same vector at different time intervals.

The user should have more information on what has been calculated, and be aware of what was used as the molecule axis, so:
closes #556
closes #399

**Fixes**
1. `AxisSelectionConfigurator` takes molecule name and up to two atom indices as input,
2. If no indices are specified, `AngularCorrelation` calculates the principal axes of inertia, and correlates the smallest one.
3. If one index is specified, `AngularCorrelation` follows a vector between the atom at that index and the molecule centre of mass,
4. If two indices are specified, `AngularCorrelation` follows a vector between the two atoms defined by these indices.
5. `AxisSelectionConfigurator` stores an entry labeled `details` where the current axis selection is explained.
6. The 1/3 scaling factor has been removed from the correlation results, so that at t=0 the correlation starts at 1.

**To test**
For a trajectory containing molecules, run the Reorientational Time Correlation Function analysis.
The results should contain an explanation of the selected molecule axis in the metadata summary in the plotter. Different axis selection will be possible in the GUI.


ChiCheng45:
Should we change the name of this job to reorientational time correlation function? This seems to be the more standard name for this type of calculation [1,2,3]. Currently, this is the l=1 reorientational time correlation function but in the future we could allow for l>1 reorientational time correlation function to be calculated.

[1] https://pubs.acs.org/doi/epdf/10.1021/acs.jpcb.4c08804?ref=article_openPDF
[2] Theory of simple liquids pg 384
[3] https://doi.org/10.1016/0009-2614(82)83383-6

MBartkowiakSTFC:
> Should we change the name of this job to reorientational time correlation function? This seems to be the more standard name for this type of calculation [1,2,3]. Currently, this is the l=1 reorientational time correlation function but in the future we could allow for l>1 reorientational time correlation function to be calculated.
> 
> [1] https://pubs.acs.org/doi/epdf/10.1021/acs.jpcb.4c08804?ref=article_openPDF [2] Theory of simple liquids pg 384 [3] https://doi.org/10.1016/0009-2614(82)83383-6

I am happy to change it. I already tried to incorporate part of the explanation in the docstring, but it makes sense to change the name.

gonzalezma:
I agree it is a good idea, as well as allowing the user to choose the degree l of the Legendre polynomial.

oerc0122:
Related: #399 ?

MDANSE/Src/MDANSE/Framework/Jobs/ReorientationalTimeCorrelationFunction.py
```diff
@@ -0,0 +1,296 @@
...
+    np.ndarray
+        An (N - corr_window + 1,) array of correlation results
+
+    """
+    array_length = len(signal)
+    result_length = array_length - corr_window + 1
+    poly = legendre(poly_order)
+    results = np.zeros(result_length)
+    for t0 in range(corr_window):
+        results += poly(np.dot(signal[t0], signal[t0 : t0 + result_length].T))
```
oerc0122, line 52: Is this a [`sliding_window`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.sliding_window)

MDANSE/Src/MDANSE/Framework/Jobs/ReorientationalTimeCorrelationFunction.py
```diff
@@ -0,0 +1,296 @@
...
+    In principle, reorientational time-correlation functions can be Legendre
+    polynomials of different order. At the moment, this analysis will calculate
+    all the orders up the maximum Legendre polynomial order specified as one
+    of the input parameters.
+
+    Angle at time T is calculated as the following:
+
+    .. math::
+
+       \\overrightarrow{vector} =  \\overrightarrow{direction} - \\overrightarrow{origin}
```
oerc0122: You have a raw string with double escaped slashes, was this intended?

MDANSE/Src/MDANSE/Framework/Jobs/ReorientationalTimeCorrelationFunction.py
```diff
@@ -0,0 +1,296 @@
...
+                    masses,
+                )
+                _, eigenvectors = np.linalg.eigh(moi)
+                if i > 0 and np.dot(diff[i - 1], eigenvectors[0]) < 0:
+                    diff[i] = -eigenvectors[0]
+                else:
+                    diff[i] = eigenvectors[0]
+                continue
+            diff[i] = coordinates[self.inner_index1] - ref_pos
+        modulus = np.sqrt(np.sum(diff**2, 1))
```
oerc0122: Are you after [`norm`](https://numpy.org/doc/2.2/reference/generated/numpy.linalg.norm.html#numpy.linalg.norm) along an axis or am I missing something?

MDANSE/Src/MDANSE/Framework/Jobs/ReorientationalTimeCorrelationFunction.py
```diff
@@ -0,0 +1,296 @@
...
+            self._outputData[f"rtcf/l={l_order}"] += x[l_order - 1]
+
+            if self.configuration["per_axis"]["value"]:
+                self._outputData[f"rtcf/per_axis/l={l_order}"][index, :] = x[
+                    l_order - 1
+                ]
+
+    def finalize(self):
+        """Normalise and write out the results."""
+        for l_order in range(1, self.legendre_order + 1):
```
oerc0122: Might want a 
```python
@property
def legendre_orders(self):
    return range(1, self.legendre_order + 1)
```
somewhere.

MDANSE/Src/MDANSE/Framework/Jobs/RotationAutocorrelation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        molecules = self.configuration["trajectory"][
+            "instance"
+        ].chemical_system._clusters[self.configuration["molecule"]["value"]]
+
+        self._indices = set(
+            idx
+            for idxs in self._configuration["atom_selection"]["indices"]
+            for idx in idxs
+        )
```
oerc0122:
```suggestion
        self._indices = {
            idx
            for idxs in self._configuration["atom_selection"]["indices"]
            for idx in idxs
        }
```

MDANSE/Src/MDANSE/Framework/Jobs/RotationAutocorrelation.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self.valid_molecules = [mol for mol in molecules if set(mol) <= self._indices]
```
oerc0122:
```suggestion
        self.valid_molecules = [mol for mol in molecules if self._indices.issuperset(mol)]
```
May be cleaner than casting to `set`?

MDANSE/Src/MDANSE/Framework/Jobs/RotationAutocorrelation.py
```diff
@@ -0,0 +1,214 @@
...
+                self.configuration["trajectory"][
+                    "instance"
+                ].chemical_system.number_of_molecules(
+                    self.configuration["molecule"]["value"],
+                ),
+            ),
+            units="au",
+        )
+
+        for axis in ["x", "y", "z"]:
```
oerc0122:
```suggestion
        for axis in "xyz":
```

MDANSE/Src/MDANSE/Mathematics/Geometry.py
```diff
@@ -68,6 +68,48 @@ def center_of_mass(coords, masses=None):
 center = center_of_mass
 
 
+def moment_of_inertia(
```
oerc0122, line 71: Do we have many other operators on diag/off-diag voigt-like symmetric matrices?

MDANSE/Tests/UnitTests/Analysis/test_rtcf.py
```diff
@@ -0,0 +1,92 @@
...
+    parameters["instrument_resolution"] = ("Gaussian", {"sigma": 1.0, "mu": 0.0})
+    parameters["interpolation_order"] = "3rd order"
+    parameters["projection"] = None
+    parameters["grouping_level"] = "atom"
+    parameters["weights"] = "equal"
+    return parameters
+
+
+@pytest.mark.parametrize("polynomial_order", [1,2,3])
+def test_rtcf(tmp_path, parameters, polynomial_order):
```
oerc0122: Do you need `generate_benchmarks`?

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculeAndAxisWidget.py
```diff
@@ -0,0 +1,112 @@
...
+            cbox.clear()
+            cbox.addItems(index_elements)
+            cbox.setCurrentText("None") if box_index else cbox.setCurrentText("0")
+            try:
+                numval = int(current_value)
+            except (TypeError, ValueError):
+                continue
+            else:
+                if numval in range(n_atoms):
+                    cbox.setCurrentText(current_value)
```
oerc0122:
```suggestion
            if numval in range(n_atoms):
                cbox.setCurrentText(current_value)
```
Don't need an else after `continue`.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculeAndAxisWidget.py
```diff
@@ -0,0 +1,112 @@
...
+        if self.selected_mol is None:
+            for cbox in self.index_combo_boxes:
+                cbox.setEnabled(False)
+            return
+        for cbox in self.index_combo_boxes:
+            cbox.setEnabled(True)
+        if self.index_combo_boxes[0].currentText() == "None":
+            self.index_combo_boxes[1].setEnabled(False)
+
+    def get_widget_value(self) -> list[str, int | None, int | None]:
```
oerc0122: What is this `list` type signature supposed to mean? `list` only accepts one arg.

c.f.
https://docs.python.org/3/library/typing.html#annotating-tuples

To simplify the logic, I think
```python
    @staticmethod
    def parse_combo_box(cbox: QComboBox) -> int | None:
        try:
            return int(cbox.currentText())
        except (TypeError, ValueError):
            return None

    def get_widget_value(self) -> tuple[str, int | None, int | None]:
        """Get the molecule name and atom indices.
        Returns
        -------
        tuple[str, int | None, int | None]
            Molecule name, optional atom index 1, optional atom index 2
        """
        return (super().get_widget_value(),
                *map(parse_combo_box, self.index_combo_boxes))
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculeAndAxisWidget.py
```diff
@@ -0,0 +1,112 @@
...
+        """Get the molecule name and atom indices.
+
+        Returns
+        -------
+        list[str, int | None, int | None]
+            Molecule name, optional atom index 1, optional atom index 2
+
+        """
+        result = [super().get_widget_value(), None, None]
+        for nbox, cbox in enumerate(self.index_combo_boxes):
```
oerc0122: Even if you don't choose the solution above, `enumerate` takes an optional startindex.
```suggestion
        for nbox, cbox in enumerate(self.index_combo_boxes, 1):
```
Which means you can just use
```python
result[nbox] = index
```
below

MDANSE/Src/MDANSE/Mathematics/Geometry.py
MBartkowiakSTFC: I don't think we do, off the top of my head.

MDANSE/Src/MDANSE/Framework/Jobs/ReorientationalTimeCorrelationFunction.py
MBartkowiakSTFC: I am happy to make this code more compact in the future, once we have all agreed that it does what it is supposed to. So far I am at the stage where I know it is correct for l=1, and it _looks_ OK for l>1, so I would prefer to leave it more verbose for potential further discussion of this approach.

oerc0122 (approved):

PR #838, opened 2025-06-16
Deselect dummy atoms

MBartkowiakSTFC:
**Description of work**
Brings back the old MDANSE behaviour where the dummy atoms are not included in the analysis unless requested. This time it is done in atom selection.

**Fixes**
1. Added a `select_dummy` atom selector, which uses the "dummy" property from atom database,
2. Included `{"function_name": "select_dummy", "operation_type": "difference"}` in the default selection in the GUI.
3. Added a unit test for `select_dummy`.

**To test**
For trajectories without dummy atoms there will be no difference in any analysis results. For trajectories containing dummy atoms, the analysis will not include dummy atoms.


MDANSE/Src/MDANSE/Framework/AtomSelector/atom_selection.py
```diff
@@ -77,3 +77,30 @@ def select_atoms(
...
+        A set of indices which have been selected
+
+    """
+    system = trajectory.chemical_system
+    element_list = system.atom_list
+    dummy_types = [
+        type
+        for type in set(element_list)
+        if trajectory.get_atom_property(type, "dummy")
+    ]
```
oerc0122:
```suggestion
    dummy_types = {
        type
        for type in set(element_list)
        if trajectory.get_atom_property(type, "dummy")
    }
```
No need to store duplicates if we're just checking contains.

MDANSE/Tests/UnitTests/test_selection.py
```diff
@@ -32,7 +32,11 @@
     "Converted",
     "2vb1.mdt"
 )
-
+traj_with_dummy = os.path.join(
+    os.path.dirname(os.path.realpath(__file__)),
+    "Converted",
+    "dummy_atoms.mdt"
+)
```
oerc0122:
```suggestion
traj_with_dummy = Path(__file__).parent / "Converted" / "dummy_atoms.mdt"
```

oerc0122 (approved):
Looks good.

Issue #638, opened 2025-01-17
[BUG] Discover converter does not appear to be reading the correct timestep


ChiCheng45:
**Description of the error**
Running a temperature calculation on the Discover trajectories found in the MDANSE-Examples repo shows large differences between the temperatures calculated using the velocities from the MD output and with numerical derivatives. Here are the results with order = 0 (blue) and order = 5 (orange) with cubane (top) and sushi (bottom).

![Image](https://github.com/user-attachments/assets/ba83ac91-eba7-48cd-af20-177c2818832f)

![Image](https://github.com/user-attachments/assets/f7f67587-e323-437b-91a2-3dac48a2aa58)

The velocities from the MD output are correct since they give a temperature of around 300 and 200 K for cubane and sushi respectively. The numerical derivatives are correct since they provide accurate results with other trajectories. Both Discover trajectories have a time step of 1 fs seems suspiciously like some default value.

MBartkowiakSTFC:
Let's add a deprecation warning to the converter.

MBartkowiakSTFC:
closed by #839

PR #839, opened 2025-06-16
Remove Discover trajectories

RobBuchananCompPhys:
**Description of work**
This change will remove all Discover converters, trajectory files and tests.

**Fixes**
Discover is no longer supported in MDANSE.

**To test**
None - tests removed.


MBartkowiakSTFC:
Please convert the sample trajectories to .mdt format before deleting them, so we keep the binary version for reference. We still don't know if they are any good (when it comes to their time step), but the system is useful for testing.

RobBuchananCompPhys:
Closes #638

MBartkowiakSTFC (approved):

PR #840, opened 2025-06-16
Add missing get_widget_value in transmutation and charge widgets

### LINKED ISSUE #833: [BUG] Atom transmutation in the GUI has stopped working
**Description of the error**
Atom transmutation has stopped working. Anything can be entered into the atom transmutation widget, and a job can be run, but the setting is not used valid or not. I think this was introduced in #814.

MBartkowiakSTFC:
**Description of work**
Brings back the missing `get_widget_value` for the widgets that inherit from `AtomSelectionWidget`, to make transmutation and charge setting work in the GUI.

closes #833 

**Fixes**
Implemented `get_widget_value` for the atom transmutation and partial charge setting widgets.

**To test**
Run a TrajectoryEditor with transmutation and charge setting. The changes should be written into the output trajectory.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
@@ -520,6 +520,9 @@ class AtomSelectionWidget(WidgetBase):
             load_button.clicked.connect(self.load_selection_from_file_dialog)
         else:
             self._field = QLineEdit(self._base)
+            default_text = str(self._configurator.default)
+            self._field.setPlaceholderText(default_text)
+            self._field.setText(default_text)
         traj_config = self._configurator._configurable[
             self._configurator._dependencies["trajectory"]
         ]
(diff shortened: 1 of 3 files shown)


ChiCheng45 (commented):
Transmutation and partial charge setting works but invalid setting can be inputted leading to a job failure.

image


Also in the infrared and dipole jobs the default setting is invalid even though the trajectory has charges.
image

MBartkowiakSTFC:
Thanks for pointing this out. It is strange that TrajectoryEditor uses the same widget for charges as Infrared, and does not produce the error on empty input. In any case, this should be fixed now.

ChiCheng45:
Looks like the above issues are fixed, but just one thing with the default setting with the atom transmutation and atom charges.

image

image

The default for both those settings should be "{}" which should be valid settings.

MBartkowiakSTFC:
> The default for both those settings should be "{}" which should be valid settings.

Thanks, I changed it so now PartialChargeWidget and AtomTransmutationWidget use the default value "{}" if the text field is empty.

ChiCheng45 (approved):
Looks good. I don't see any other issues.

PR #841, opened 2025-06-16
Add GUI warnings for the trajectory widget

### LINKED ISSUE #764: [ENHANCEMENT] A visible GUI warning for job inputs

MBartkowiakSTFC:
**Description of work**
Highlights the trajectory widget and the run button if the input trajectory does not have a usable unit cell definition.

This PR is a subset of changes from PR #780.

closes #764

**Fixes**
- `IConfigurator` stores a warning status that can be checked by GUI widgets,
- trajectory widget and the run button both show a warning if the trajectory does not have a usable unit cell,
- Both error and warning checks are now performed when the job is first being set up (which was not the case before), as well as after parameter changes.

**To test**
Load a trajectory without a unit cell, or with unit cell set to all zeros. Pick an analysis type. The warning should be highlighted in the GUI.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
@@ -155,6 +155,7 @@ class IConfigurator(dict, metaclass=SubclassFactory):
         self._valid = True
 
         self._error_status = "OK"
+        self.warning_status = ""
 
         self._original_input = ""
 
--- a/MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
@@ -13,16 +13,26 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
+from __future__ import annotations
 
 import abc
 import json
 from pathlib import Path
+from typing import TYPE_CHECKING
+from warnings import warn
 
 import numpy as np
 from more_itertools import value_chain
 
 from MDANSE.Core.Error import Error
 from MDANSE.Core.SubclassFactory import SubclassFactory
+from MDANSE.MLogging import LOG
+
+if TYPE_CHECKING:
+    from MDANSE.Framework.Configurable import Configurable
+
+
+ERROR_LENGTH_MIN = len("OK")
 
 
 class CustomEncoder(json.JSONEncoder):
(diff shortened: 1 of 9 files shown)


MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -155,15 +153,17 @@ def __init__(self, name, **kwargs):
...
+        """Output all the configurator attributes and dict entries as text."""
         return "\n".join(
             value_chain(
                 "",
                 (
-                    f"{label}={str(getattr(self, label, 'Not set'))}"
+                    f"{label}={getattr(self, label, 'Not set')!s}"
                     for label in self._printable_attributes
                 ),
                 (f"{key}={str(self.get(key, 'Not set'))}" for key in self),
```
oerc0122: `str` is the default call for an f-string, so we can probably do without the `!s`, up to you, but we should be consistent if we're changing it in one place.

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -235,27 +235,51 @@ def valid(self):
...
         error_text : str
             Text explaining why the current input is invalid
+
         """
         self._error_status = error_text
-        if len(self._error_status) > 2:
+        if len(self._error_status) > ERROR_LENGTH_MIN:
             self._valid = False
         else:
             self._valid = True
```
oerc0122: `self._valid = self._error_status == "OK"` ?

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -283,9 +291,18 @@ def _load_unit_cells(self):
                         )
                     uc = UnitCell(temp_array)
                     self._unit_cells.append(uc)
+                    if not self.unit_cell_warning:
+                        if abs(uc.volume) < CELL_SIZE_LIMIT:
```
```suggestion
                    if not self.unit_cell_warning and  abs(uc.volume) < CELL_SIZE_LIMIT:
```

MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
```diff
@@ -253,6 +260,16 @@ def _load_unit_cells(self):
             self._unit_cells = [UnitCell(uc) for uc in self._h5_file["unit_cell"][:]]
         else:
             self._unit_cells = None
+            self.unit_cell_warning = NO_CELL
+        if not self.unit_cell_warning:
+            if abs(self._unit_cells[0].volume) < CELL_SIZE_LIMIT:
```
oerc0122: Would a negative cell volume not be a more concerning result in a cartesian system? I think that would mess up a lot of analyses

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/WidgetBase.py
```diff
@@ -157,19 +158,44 @@ def configure_using_default(self):
...
+
+        Parameters
+        ----------
+        warning_text : str
+            Message displayed on hover-over.
+        """
+        if warning_text:
+            self.has_warning = True
+            self._base.setStyleSheet(
+                "QWidget#InputWidget { background-color:rgb(220,210,30); font-weight: bold }"
```
oerc0122: Does this want to be a constant on the class, same for the error format.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
```diff
@@ -410,14 +411,30 @@ def show_output_prediction(self):
     @Slot()
     def allow_execution(self):
         allow = True
+        has_warning = False
         for widget in self._widgets:
             if not widget._configurator.valid:
                 allow = False
+                widget.mark_error(widget._configurator.error_status, silent=True)
+            if widget.has_warning:
+                has_warning = True
```
oerc0122:
```suggestion
            has_warning = has_warning or widget.has_warning
```
Same meaning, style choice, no preference really.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
```diff
@@ -410,14 +411,30 @@ def show_output_prediction(self):
...
         if self.execute_button is not None:
             if allow:
                 self.execute_button.setEnabled(True)
             else:
                 self.execute_button.setEnabled(False)
```
oerc0122:
```suggestion
            self.execute_button.setEnabled(allow)
```

oerc0122 (approved):
Looks good. Like the solution that we can still get a cli warning. We can always clean up later if we decide this structure isn't what we need.

MBartkowiakSTFC:
Thanks for the comments. I think I implemented all the changes you suggested.

PR #850, opened 2025-06-24
Scattering Length Density Profile

### LINKED ISSUE #820: [BUG] DensityProfile results: scale and unit
**Description of the error**
A few questions about the DensityProfile job:

1. What does "density" imply for the scale of the curves.
Here I have the same result, once with weights "equal" and again with weights "atomic_weight". All the atoms in the system were carbon atoms. Normally, it would make sense if density scaled by mass was 12 times higher if we scale by atom mass. Does it make sense to apply our normal weight scheme here?
Image

2. Should the results depend on bin size?
Here are two results of the same calculation. The only difference is that in one of the runs the bin size r was 0.01, and in the other 0.04. Larger bin results in fewer points with larger values. Again, if we were just counting atoms or their mass, that would probably be fine, but since we are talking about **density**, I would expect the curves to be on the same scale, independent of the bin size.
Image

I guess that both questions depend also on what we think the unit of the output is, but since we allow weights to be applied, the unit likely cannot be fixed, and would have to be assigned depending on the weights chosen.

In principle it makes sense to allow different weights, as we may want to calculate the profile of scattering length density. This may have even been the main use case, based on the documentation string of this job.

**Suggested fix**
Please let me know if you think this is a problem that needs fixing.

MBartkowiakSTFC:
**Description of work**
DensityProfile job is replaced by a more specific ScatteringLengthDensityProfile.

closes #820

**Fixes**
1. DensityProfile is renamed to ScatteringLengthDensityProfile.
2. The unit of the output is 1e-6 / A^2.
3. Values are normalised to the slice volume.

**To test**
Unit tests have been added.


MDANSE/Src/MDANSE/Framework/Jobs/ScatteringLengthDensityProfile.py
```diff
@@ -177,10 +203,15 @@ def combine(self, index: int, data: tuple[float, dict[str, np.ndarray]]) -> None
...
+        slice_volume, density_profile = data
 
         for element, hist in density_profile.items():
-            self._outputData[f"dp_{element}"] += hist
+            self._outputData[f"dp_{element}"] += hist / slice_volume
+            slen_dict = self.scattering_lengths[element]
+            for nb, component in enumerate(["", "_incoherent", "_total"]):
+                self._outputData[f"sldp{component}"] += (
+                    slen_dict[nb] * hist / slice_volume
+                )
```
oerc0122:
```suggestion
            for component, scat_len in zip(["", "_incoherent", "_total"], slen_dict):
                self._outputData[f"sldp{component}"] += (
                    scat_len * hist / slice_volume
                )
```

MDANSE/Src/MDANSE/Framework/Jobs/ScatteringLengthDensityProfile.py
```diff
@@ -177,10 +203,15 @@ def combine(self, index: int, data: tuple[float, dict[str, np.ndarray]]) -> None
             Axis length and density profile.
         """
 
-        self._extent, density_profile = data
+        slice_volume, density_profile = data
 
         for element, hist in density_profile.items():
-            self._outputData[f"dp_{element}"] += hist
+            self._outputData[f"dp_{element}"] += hist / slice_volume
+            slen_dict = self.scattering_lengths[element]
```
oerc0122: Is this a dict? Looks below like a sequence?

oerc0122 (approved):
Looks good, just a couple of very minor things

MDANSE/Src/MDANSE/Framework/Jobs/ScatteringLengthDensityProfile.py
MBartkowiakSTFC: You're right, it is not a dict. `self.scattering_lengths` was a dict, but `slen_dict` is not. I renamed it to `slen_list`.

PR #851, opened 2025-06-24
Atoms selection and molecule grouping results in the documentation

ChiCheng45:
**Description of work**
Added a few sections explaining how the results work when atom selections are made or when the molecule grouping is used. Fixed equations in the weights section. Reformatted the atom selection section to reduce the number of subtitles.

### DIFF
--- a/Doc/pages/grouping.rst
+++ b/Doc/pages/grouping.rst
@@ -8,10 +8,11 @@ Results Grouping
 Atom Grouping
 ^^^^^^^^^^^^^
 
-With the default settings, MDANSE will group results by atom so that results are
-summed for all atoms of each type in the system and those results are
-summed using the :ref:`weighting-scheme` to obtain the total. Lets
-consider a system of water and ethanol. With atom grouping,
+With the default settings, MDANSE will group results by atom type, so that
+partial results are summed into datasets, with one dataset per atom type
+present in the system. The total result is calculated from these partial
+results using the :ref:`weighting-scheme`. Let's
+consider a system consisting of a mixture of water and ethanol. With atom grouping,
 the partial coherent intermediate scattering
 functions are obtained for each unique pair of atoms in the system:
 CC, CH, CO, HH, HO, and OO. For example, the C and H partial coherent
@@ -53,7 +54,7 @@ see :ref:`weighting-scheme` for further details.
 Molecule Grouping
 ^^^^^^^^^^^^^^^^^
 
-MDANSE can also group results by molecule so that they are summed for
+MDANSE can also group results by molecule type, so that they are summed for
 all atoms of each type on each type of molecule, again we will consider
 a system of water and ethanol. With molecule grouping, the partial
 coherent intermediate scattering functions are obtained for each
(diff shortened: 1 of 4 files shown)


Doc/pages/grouping.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. math::
+   :label: moleculepartial1
+
+   F_{\text{coh},\text{HO}}^{[\text{EtOH}][\text{H2O}]}{(\mathbf{q},t) =  \frac{1}{N \sqrt{c_{\text{H}}^{\text{EtOH}}c_{\text{O}}^{\text{H2O}}}}}{\sum\limits_{j \in (\text{EtOH}\, \cap\, \text{H})}{\sum\limits_{k \in (\text{H2O}\, \cap\, \text{O})}\left\langle {\exp\left\lbrack {{- i}\mathbf{q}\cdot\mathbf{r}_{j}\left( 0 \right)} \right\rbrack\exp\left\lbrack {i\mathbf{q}\cdot\mathbf{r}_{k}\left( t \right)} \right\rbrack} \right\rangle}},
+
+where :math:`c_{\text{H}}^{\text{EtOH}} = N_{\text{H}}^{\text{EtOH}} / N` and
+:math:`c_{\text{O}}^{\text{H2O}} = N_{\text{O}}^{\text{H2O}} / N`. Here,
+:math:`N_{\text{H}}^{\text{EtOH}}` and :math:`N_{\text{O}}^{\text{H2O}}` are
+the total number of atoms of hydrogen in ethanol and oxygen in water respectively,
```
MBartkowiakSTFC, line 75: Do you think that "total number of atoms of hydrogen in ethanol" is ambiguous? As in, it could read as "number of hydrogen atoms in all ethanol molecules", but also possibly "number of hydrogen atoms in an ethanol molecule"?

Doc/pages/grouping.rst
```diff
@@ -0,0 +1,188 @@
...
+setting which does follow the above mechanism.
+
+Root Mean Square Deviation
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Grouping in the root mean squared deviation (RMSD) analysis calculation
+works similarly to the above mechanism except for the fact that the RMSD
+analysis does not use a weight scheme and the grouping is done before
+the square root function is applied to the mean square displacement.
+Grouping is done in this way so that the grouping with the RMSD analysis
```
MBartkowiakSTFC: This sentence is potentially hard to follow. At the moment, I tend to read it as:
"In RMSD, grouping is implemented in such a way that the result for a group is the same as the total result with a selection which includes only the atoms from this group."
Did I understand it right?

Doc/pages/grouping.rst
ChiCheng45: I think its fine otherwise it becomes a bit too verbose.

"the total number of atoms of hydrogen in all ethanol molecules and oxygen in all water molecules respectively"

Doc/pages/grouping.rst
ChiCheng45: Yea that's right. I've updated the text a bit so hopefully should be a bit clearer now.

MBartkowiakSTFC (approved):
OK, thanks for the answers and changes. I think this should be good to go.

PR #852, opened 2025-06-25
Test benchmark auto-generation

oerc0122:
**Description of work**
Add feature to enable `-G` flag to regenerate tests with static files.

Run `ruff format` in some of the test files.

Currently, it generates in-place overwriting existing test file. Could generate instead with the original temporary name and copy only the needed file to the result file.

**Fixes**
Relieves frustration at manually copying test data back?

**To test**
1. Change a test
2. Run `pytest -G <test you changed>` and see there are newly generated `.mda` files.

### DIFF
--- a/MDANSE/Tests/UnitTests/Analysis/test_average_structure.py
+++ b/MDANSE/Tests/UnitTests/Analysis/test_average_structure.py
@@ -1,8 +1,8 @@
 import pytest
-from MDANSE.MolecularDynamics.Trajectory import \
-    Trajectory
-from MDANSE.Framework.Jobs.IJob import IJob
 from test_helpers.paths import CONV_DIR
+
+from MDANSE.Framework.Jobs.IJob import IJob
+from MDANSE.MolecularDynamics.Trajectory import Trajectory
 
 short_traj = CONV_DIR / "short_trajectory_after_changes.mdt"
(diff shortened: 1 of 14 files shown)


MBartkowiakSTFC:
Overwriting the files in place is fine. Overall, it works, and I checked that the failing (deliberately?) tests will work if the results are regenerated, and if `compare_hdf5` gets matching flags
```
        scale_result=True,
        scale_benchmark=True,
```
It seems to me that we still have some old reference files, generated a while ago, which is why the `scale_...` flags sometimes don't match, but they should be set to matching values after regenerating the reference files.

oerc0122:
> Overwriting the files in place is fine. Overall, it works, and I checked that the failing (deliberately?) tests will work if the results are regenerated, and if `compare_hdf5` gets matching flags
> 
> ```
>         scale_result=True,
>         scale_benchmark=True,
> ```
> 
> It seems to me that we still have some old reference files, generated a while ago, which is why the `scale_...` flags sometimes don't match, but they should be set to matching values after regenerating the reference files.

Yes, two accidentally leftover broken tests which I broke to test breaking and left broken.

MBartkowiakSTFC (approved):
I think this will be useful. Looks good.

PR #854, opened 2025-06-27
Rework hdf trees

### LINKED ISSUE #661: [ENHANCEMENT] Better labels for clusters containing isotopes
The current implementation of ChemicalSystem assigns labels to atom groups/clusters based on the chemical elements of the atoms included in the group. Water becomes `H2_O1`, a C60 fullerene becomes a `C60`, etc.

Unfortunately, the MDANSE element database labels isotopes with the digit _after_ the element symbol. At the moment, deuterium is `H2`, and heavy water gets the label `H22_O1`.

Should we include additional separators between the chemical elements symbols and numbers in the formula? Or do it only conditionally, if the element symbol ends with a digit?

We could also shift the isotope mass before the chemical symbol, which is closer to the correct chemical notation, but more problematic for all the components that rely on the old labels.

**Additional context**
N/A

### LINKED ISSUE #828: [BUG] Order of the results are jumbled compared to previous commits
**Description of the error**
The order of the results has been changed compared to previous commits of MDANSE. This has made it more difficult to find what you are looking for. Previously, it looked like it was ordered alphabetically, but now it looks random.

**Additional details**
Latest commit
Image

8947373572069c1f3baaedfa744afa09eb810c39
Image

oerc0122:
**Description of work**
- Restructure MDA output data into (hopefully) logical trees
- Relabel groups to use `<>` rather than `[]`
- Relabel isotopes to use `[]`  

**Fixes**
Fixes #828
Fixes #661

**To test**
Updated tests should pass. 
Will need manual verification that results appear sane as all test data have been updated.


MBartkowiakSTFC:
So far I am mainly concerned about quick plotting. I have been testing the 'Dynamics' group of analysis jobs, and so far it seems that:
1. Double-clicking the file name produces an empty plot.
2. If I double-click a specific group in the output file, the contents get plotted, but the axes data sets get pulled into the plot as well.
Screenshot 2025-06-30 at 12 32 29

To fix 1. we could either scan the entire data tree for all the datasets with the "main" tag, or change the mechanism so we store the information about which datasets to plot somewhere central in the HDF5 file.

To fix 2. we could either change the file structure (which is a lot of work), or limit the recursion of the double-click plotting, so it only plots one level of nesting (which seems reasonable, since we don't normally want too many datasets in one plot).

oerc0122:
I have implemented solution 1. We can limit recursion semi-trivially if we wish.

MBartkowiakSTFC:
Sorry, the plotter changes should probably be undone and we will go with #860 instead. In this branch it seems to me that automatic loading of job results in the GUI stopped working too.

MBartkowiakSTFC:
I think I managed to test and compare results of all the analysis jobs. The cases I covered were:
1. Grouping by atom,
2. Grouping by molecule with 1 molecule type present.
For these cases, the results agreed for all the jobs.

Note: ScatteringLengthDensityProfile will need to have the "main" tag set, but we wanted to change that job anyway, so this does not have to be done here.

MBartkowiakSTFC (approved):
I did not notice any problems while testing this branch. Looks good.

PR #856, opened 2025-06-30
Reorder jobs

### LINKED ISSUE #846: [ENHANCEMENT] Review the analysis job group in the tree.
Some of the analysis jobs may not be placed in the correct group. For example, both RMSD and RMSF are under the Structure analysis job group, but maybe they should be under dynamics. Similarly, staticstructurefactor and xraystaticstructure factor could be moved to the scattering group.

MBartkowiakSTFC:
**Description of work**
Two main goals of this PR are to order the jobs alphabetically in the GUI, and to assign the jobs to the right categories.

closes #846 

**Fixes**
1. Add alphabetical sorting of the jobs,
2. Remove the "Materials Studio" category from the converters,
3. Move RMSD and RMSF to dynamics, and SSF and XSSF to scattering.
4. Rearrange documentation to reflect the changes.

**To test**
Start the GUI. The converters and analysis jobs should both appear in alphabetical order.


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobTree.py
```diff
@@ -57,7 +57,24 @@ def populateTree(self, parent_class=None, filter=None):
...
+            if not full_dict[class_name].enabled:
+                continue
+            cat_tuple = getattr(full_dict[class_name], "category", None)
+            if cat_tuple and len(cat_tuple) > 1:
+                temp = cat_dicts.get(cat_tuple[0], list())
+                temp.append(cat_tuple[1])
+                cat_dicts[cat_tuple[0]] = temp
+        sorted_cats = sorted(cat_dicts.keys())
+        for cat in sorted_cats:
+            cat_dicts[cat] = sorted(cat_dicts[cat])
```
oerc0122: This won't resort the `cat_dicts` dictionary, because it's already filled it'll maintain the original order.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobTree.py
```diff
@@ -57,7 +57,24 @@ def populateTree(self, parent_class=None, filter=None):
...
+        for cat in sorted_cats:
+            for subcat in cat_dicts[cat]:
+                self.parentsFromCategories((cat, subcat))
+        for class_name in sorted_keys:
+            class_object = full_dict[class_name]
             if class_object.enabled:
                 self.createNode(class_name, class_object, filter)
```
oerc0122:
```suggestion
        sorted_keys = sorted(full_dict)
        cat_dicts = defaultdict(list)
        for class_name in sorted_keys:
            if not full_dict[class_name].enabled:
                continue
            cat_tuple = getattr(full_dict[class_name], "category", None)
            if cat_tuple and len(cat_tuple) > 1:
                cat_dicts[cat_tuple[0]].append(cat_tuple[1])
                
        cat_dicts = {cat: sorted(cat_dicts[cat]) for cat in sorted(cat_dicts)}

        for cat, vals in cat_dicts.items():
            for subcat in vals:
                self.parentsFromCategories((cat, subcat))

        for class_name in sorted_keys:
            class_object = full_dict[class_name]
            if class_object.enabled:
                self.createNode(class_name, class_object, filter)
```

oerc0122 (commented):
I think we ought to split the docs changes out of this PR.

MBartkowiakSTFC:
> I think we ought to split the docs changes out of this PR.

I would argue here that this PR changed the code to make four jobs appear in different categories than before. The documentation changes are moving the same four jobs to their new categories, so that the categories in the documentation match those in the code. From my point of view, it makes sense to make both changes at the same time.

Unless you mean those two equation labels which I changed because they were colliding?

oerc0122 (approved):

PR #860, opened 2025-07-01
Plot all files

MBartkowiakSTFC:
**Description of work**
In connection with PR #854, changes to the MDANSE_GUI plotter are proposed here that would make it possible to apply quick plotting to both the old (flat) and new (nested) MDA files.

**Fixes**
1. Add optional limit to the recursion depth when searching the data file.
2. Modify the `quick_plot_data` function to optionally filter out the datasets without the main tag.

**To test**
The expected behaviour for both old (protos branch) and new (PR #854) MDA files is:
1. When the file name is double-clicked, only the "main" datasets are plotted.
2. When a dataset is double-clicked, this dataset is plotted.
Additionally, for the new files:
3. If a group is double-clicked, only its direct children are plotted.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
@@ -59,9 +59,15 @@ class PlotDataView(QTreeView):
         model = self.model()
         inner_node = model.inner_object(index)
         if isinstance(inner_node, MDADataStructure):
-            self.quick_plot_file(inner_node)
+            data_nodes = model.itemFromIndex(index).recursive_children(
+                recursion_limit=10
+            )
+            file_node = model.parent_object(index)
+            self.quick_plot_data(data_nodes, file_node, main_only=True)
         else:
-            data_nodes = model.itemFromIndex(index).recursive_children()
+            data_nodes = model.itemFromIndex(index).recursive_children(
+                recursion_limit=1
+            )
             file_node = model.parent_object(index)
             self.quick_plot_data(data_nodes, file_node)
 
@@ -136,35 +142,12 @@ class PlotDataView(QTreeView):
             except Exception:
                 self.item_details.emit("No additional information included.")
 
-    def quick_plot_file(self, mda_data_structure: MDADataStructure):
-        """Automatically select, format and plot datasets from one file.
-
-        Parameters
-        ----------
-        mda_data_structure : MDADataStructure
-            Object containing and MDA data file.
-
-        """
-        model = PlottingContext()
-        for key in mda_data_structure._file:
-            try:
-                if "main" in mda_data_structure._file[key].attrs["tags"]:
-                    if "partial" in mda_data_structure._file[key].attrs["tags"]:
-                        dataset = SingleDataset(
-                            key, mda_data_structure._file, linestyle="--"
-                        )
-                    else:
-                        dataset = SingleDataset(key, mda_data_structure._file)
-
-                    model.add_dataset(dataset)
-            except KeyError:
-                LOG.error(f"No attribute called Tag found in {key}, skipping")
-        self.fast_plotting_data.emit(model)
-
     def quick_plot_data(
         self,
         data_nodes: list[BasicPlotDataItem],
         mda_data_structure: MDADataStructure,
+        *,
+        main_only: bool = False,
     ):
         """Plot several datasets in a new plot instance.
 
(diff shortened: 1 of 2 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
```diff
@@ -59,9 +59,15 @@ def mouseDoubleClickEvent(self, e: QMouseEvent) -> None:
         model = self.model()
         inner_node = model.inner_object(index)
         if isinstance(inner_node, MDADataStructure):
-            self.quick_plot_file(inner_node)
+            data_nodes = model.itemFromIndex(index).recursive_children(
+                recursion_limit=10
+            )
```
oerc0122: If you want this unlimited, I would say set it to `-1`

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
```diff
@@ -174,12 +157,28 @@ def quick_plot_data(
...
+            if True, only plot datasets with the 'main' tag
 
         """
         model = PlottingContext()
         file = mda_data_structure._file
         for data_node in data_nodes:
-            dataset = SingleDataset(data_node.child_path, file)
+            if not data_node.child_path:
+                continue
+            if main_only:
```
oerc0122, line 169: Is the logic backwards here? Why does "main_only" include "partial"?

MBartkowiakSTFC: We probably could have picked better words for these tags. Effectively, the meaning of the tags is:
'main' - include it in the quick plot,
'partial' - plot it with a dashed line in the quick plot.
So, a dataset must have a 'main' tag to be included in the quick plot. 'partial' modifies the appearance. Setting a 'partial' tag on a dataset that is not 'main' is meaningless, since 'partial' is only check in quick plotting, which is only done on 'main' datasets.

All this applies if you want to create a quick plot of a data file. If you are creating a quick plot of a group or a single dataset, the tags are ignored.

oerc0122:
```python
        def main_filter(node):
            if not hasattr(node, "child_path"):
                return False
            return not main_only or "main" in file[node.child_path].attrs["tags"]

        for data_node in filter(main_filter, data_nodes):
            style = "--" if "partial" in file[data_node.child_path].attrs["tags"] else "-"
            dataset = SingleDataset(data_node.child_path, file, linestyle=style)
 ```

?

File: MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py [hunk 3]
```diff
@@ -174,12 +157,28 @@ def quick_plot_data(
...
+                except KeyError:
+                    continue
+                if "main" not in tags:
+                    continue
+                if "partial" in tags:
+                    dataset = SingleDataset(data_node.child_path, file, linestyle="--")
+                else:
+                    dataset = SingleDataset(data_node.child_path, file)
+            else:
+                dataset = SingleDataset(data_node.child_path, file)
```
RobBuchananCompPhys:
```suggestion
                if "partial" in tags:
                    dataset_kwargs.update({"linestyle": "--"})
            model.add_dataset(
                SingleDataset(data_node.child_path, file, **dataset_kwargs)
            )
```

If we store a dictionary dataset_kwargs = {} at the top, we could save a couple of lines of if/else statements here.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
MBartkowiakSTFC: This would be more compact, but even with some adjustments I only got it to the stage where it plotted the old files, but not the new ones. (Without adjustments, I could not plot either.)

RobBuchananCompPhys (approved):
Small suggestion you may want to consider to save a few lines in `PlotDataView.py`, otherwise Looks good!

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
oerc0122: Fair enough.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
oerc0122: Doesn't even need to be kwargs, because we know the default (`-`), can just set style:
```python
style = "--" if "partial" in tags else "-"
dataset = SingleDataset(data_node.child_path, file, linestyle=style)
```

oerc0122 (approved):

PR #863, opened 2025-07-01
Make COM trajectory use real cluster composition

### LINKED ISSUE #861: [BUG] CentreOfMassTrajectory fails when using TrajectoryEditor output
**Description of the error**
CentreOfMassTrajectory works for ordinary trajectories. However, it fails for trajectories in which the chemical elements have been changed using TrajectoryEditor.

For example, replacing H with H2 (deuterium) in water makes the COM trajectory fail when it tries to write out the averaged atom properties.

It should work the same for all trajectories, no matter how they were written out.

```
2025-07-01 10:04:51,960 - CRITICAL - process[15658] - IJob 449 - Job failed with traceback: Traceback (most recent call last):
File "MDANSE/Framework/Jobs/IJob.py", line 443, in run
self.finalize()
File "MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py", line 201, in finalize
self._output_trajectory.write_atom_database(
File "MDANSE/MolecularDynamics/Trajectory.py", line 686, in write_atom_database
property_dict = create_average_atom(
^^^^^^^^^^^^^^^^^^^^
File "MDANSE/MolecularDynamics/Trajectory.py", line 457, in create_average_atom
total += converted * entry[1]
TypeError: 'float' object is not iterable
```

**Suggested fix**
It is possible that TrajectoryEditor writes the internal atom database of the output trajectory in an incorrect way.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
To determine the atomic composition of a cluster/molecule, we look up the atom types of all the component atoms in the original trajectory's ChemicalSystem. The old mechanism was trying to read it by parsing the cluster label.

closes #861

**Fixes**
1. Create a `self.composition_lookup` dictionary in CentreOfMassesTrajectory.
2. Use the composition dictionary in `TrajectoryWriter.write_atom_database`.

**To test**
Take a trajectory containing molecules.
Use TrajectoryEditor to replace one of the elements in the molecules with an isotope.
Run CentreOfMassesTrajectory on the trajectory with the isotope. It should work correctly (and fail on the main branch).

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
@@ -84,11 +84,17 @@ class CenterOfMassesTrajectory(IJob):
         self.numberOfSteps = self.configuration["frames"]["number"]
         chemical_system = self.configuration["trajectory"]["instance"].chemical_system
 
+        self.cluster_composition = {}
+        original_atom_list = chemical_system.atom_list
         new_element_list = []
         used_up_atoms = set()
         new_chemical_system = ChemicalSystem()
         for cluster_name in chemical_system._clusters.keys():
             for cluster in chemical_system._clusters[cluster_name]:
+                if cluster_name not in self.cluster_composition:
+                    self.cluster_composition[cluster_name] = [
+                        original_atom_list[ind] for ind in cluster
+                    ]
                 new_element_list.append(cluster_name)
                 used_up_atoms.update(set(cluster))
         for index in chemical_system._atom_indices:
@@ -201,6 +207,7 @@ class CenterOfMassesTrajectory(IJob):
         self._output_trajectory.write_atom_database(
             self._unique_atoms,
             self.configuration["trajectory"]["instance"],
+            self.cluster_composition,
             time_averaged_radii,
         )
         # The input trajectory is closed.
--- a/MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
+++ b/MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from MDANSE.Chemistry.Databases import AtomsDatabase
 
+from collections import Counter
 import copy
 import math
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
```diff
@@ -84,11 +84,17 @@ def initialize(self):
...
+        original_atom_list = chemical_system.atom_list
         new_element_list = []
         used_up_atoms = set()
         new_chemical_system = ChemicalSystem()
         for cluster_name in chemical_system._clusters.keys():
             for cluster in chemical_system._clusters[cluster_name]:
+                if cluster_name not in self.cluster_composition:
+                    self.cluster_composition[cluster_name] = [
+                        original_atom_list[ind] for ind in cluster
+                    ]
```
oerc0122, line 97: I have seen a lot of this structure throughout the code. I wonder if at some point we might want to replace a lot of this with [`operator.itemgetter`](https://docs.python.org/3/library/operator.html#operator.itemgetter)

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -643,7 +644,8 @@ def write_atom_database(
         self,
         symbols: list[str],
         database: "AtomsDatabase",
-        optional_molecule_radii: dict[str, float] = None,
+        composition_lookup: dict[str, list[str]],
```
oerc0122, line 647: These arguments are added, but are they applied everywhere? Is `CenterOfMassesTrajectory` the only function which uses this function?

oerc0122 (approved):
Looks good if the second question is true.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
MBartkowiakSTFC: It is the only one. The others just work with the existing atom types, but `CenterOfMassesTrajectory` creates new types.

MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
MBartkowiakSTFC: I will have a look at `operator.itemgetter`. You are right that at the moment this kind of list comprehension is used a lot in MDANSE.

PR #873, opened 2025-07-09
Update neutron weights squaring in analysis and documentation

### LINKED ISSUE #872: [BUG] Sort out DOS so it can weighted with neutron scattering lengths and check weight squaring in a few jobs.
**Description of the error**
In some cases you may want to calculate the DOS weighted with neutron scattering lengths. They should use what was b_incoherent2, see molecular hydrodynamics eq 2.2.30. Since we got rid of b_incoherent2 we need to update the DOS job and probably others so the b_incoherent is squared. 

Also I just noticed that all weights seem to be squared in some jobs for some reason. For DISF this makes sense with b_incoherent but I don't think it should for anything else. 

https://github.com/ISISNeutronMuon/MDANSE/blob/e2006b0a4355f6c22f52511774a7d12edd4a3213/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py#L291

MBartkowiakSTFC:
**Description of work**
Following the removal of `b_incoherent2`, changes have been made to the analysis type that used `b_incoherent2` to ensure that the same results are obtained as before the changes. This PR aims to make analysis types related to each other apply the weights consistently, and to update the documentation so the way the weights are applied is correctly described in the text.

closes #872

**Fixes**
1. Documentation should mention every analysis type where the weights are squared.
2. In DOS,  the weights are only squared if they are b_coherent or b_incoherent. This way the atomic_weight weights will produce a valid DOS, and b_incoherent will produce a neutron-measured DOS.
3. In PACF, VACF and PPS the weights are only squared if they are b_coherent or b_incoherent. For VACF, it was done to stay consistent with DOS, and for PACF and PPS it was done to stay consistent with VACF.
4. For DISF, EISF and GDISF, the weights are always squared. This is meant to put the results on the same scale as DCSF, where the f_AA pairs end up with the square of the weight for atom A.
5. Other documentation fixes: removed AngularCorrelation again, added MDANSE to documentation dependencies (which should fix a build error on ReadTheDocs.)

**To test**
Check is the documentation is consistent with the code.


oerc0122 (approved):
Looks good to me.

ChiCheng45:
I've made some updates to the documentation. However there are still a few things.

The FT of the self part of the van hove function is related to the intermediate scattering function. I think the weights for the self part of the van hove function be squared as well.

Since there is always some inconsistency e.g. technically to be consistent with ourselves all weights for the DOS calculations and etc should be squared since they are related to the DISF through some relations.

```math
\mathrm{DOS}(\omega) = \lim_{k  \rightarrow 0} \left( \frac{\omega}{k}\right)^2 S_{\text{inc}}(k, \omega)
```

I wonder we should have left it so that weight are never squared and included `b_incoherent2` and `b_coherent2` or maybe added the option to allow the user to apply some power to the weights before they are rescaled and used.

MBartkowiakSTFC:
Thanks for the changes!

You are right that there are still problems with the current implementation.

In general, I did not like the idea of storing `b_incoherent2` as a value in our atom database. But we could create `b_incoherent2` (and `b_coherent2` if we need it) as input for WeightsConfigurator, and have the values squared there as needed. This brings us closer to the other idea, to have the power added as an extra parameter. This would make sense, but I can also how this increases the complexity of the input and the chance that some of our users end up calculating something else than they wanted. We would then have to highlight the valid combinations of(weight, power) pairs than have some known physical meaning.

Should we merge the changes for now, and come back to this problem once we have decided on the best solution?

ChiCheng45:
> Thanks for the changes!
> 
> You are right that there are still problems with the current implementation.
> 
> In general, I did not like the idea of storing `b_incoherent2` as a value in our atom database. But we could create `b_incoherent2` (and `b_coherent2` if we need it) as input for WeightsConfigurator, and have the values squared there as needed. This brings us closer to the other idea, to have the power added as an extra parameter. This would make sense, but I can also how this increases the complexity of the input and the chance that some of our users end up calculating something else than they wanted. We would then have to highlight the valid combinations of(weight, power) pairs than have some known physical meaning.
> 
> Should we merge the changes for now, and come back to this problem once we have decided on the best solution?

I happy to merge now and figure a better solution latter. What did you think of the self part of the van Hove function? I think the weight should be squared for this calculation as well since its related to the DISF.

MBartkowiakSTFC:
You are right, in the self part of the van Hove function the weights should be squared too.

ChiCheng45 (approved):
Looks good to merge now.

PR #875, opened 2025-07-15
Remove part of the nested dependencies of configurators

MBartkowiakSTFC:
**Description of work**
Currently, `Configurator` objects have a mixed function of both parsing input and processing the trajectory. Also, atom transmutation and grouping both work implicitly on the atom selection, and `AtomSelectionConfigurator` is the access point to the information about atom selection, transmutation and grouping.

This PR is meant to restructure the code, so the `Configurator`s can be limited to checking the validity of the input, and the changes triggered by the input are applied by the `Trajectory` class.

Reducing the number of operations performed by `Configurator` classes is meant to improve the responsiveness of the GUI, as the `.configure()` methods are called whenever the input string has changed.

**Fixes**
1. `IJob.initialize()` executes a new method, `IJob.set_up_trajectory()`, which applies atom selection, atom transmutation and grouping to the `Trajectory` instance when the job is starting.
2. Result grouping functions have been moved from `GroupingLevelConfigurator` to `MDANSE.Framework.AtomGrouping.grouping`.
3. `AtomTransmutationConfigurator` and `GroupingLevelConfigurator` depend only on `trajectory`.
4. Information about the chemical elements, selected atom indices and atom labels is now provided by properties of `Trajectory` and not by `Configurator` objects.

**To test**
There should be no change to the results from the point of view of the end user.


MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+
+    if dim == 1:
+        for grp in trajectory.group_lookup:
+            grp_ele = sorted(
+                set(
+                    trajectory.atom_types[x]
+                    for cluster in trajectory.chemical_system._clusters[grp]
+                    for x in cluster
+                    if x in trajectory.atom_indices
+                )
```
oerc0122:
```suggestion
                {
                    trajectory.atom_types[x]
                    for cluster in trajectory.chemical_system._clusters[grp]
                    for x in cluster
                    if x in trajectory.atom_indices
                }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+        A function which is applied to the results.
+    post_label: str
+        The label to be added for grouped summed results.
+    """
+    tot_n_atms = len(trajectory.atom_indices)
+
+    if trajectory._grouping_level == "atom":
+        return
+
+    if dim == 1:
```
oerc0122, line 73: Wonder if this should be broken into two `_` functions, leaving `add_grouped_interface` as a selector/wrapper.

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+    elif dim == 2:
+        if intra:
+            for grp in trajectory.group_lookup:
+                eles = sorted(
+                    set(
+                        trajectory.atom_types[x]
+                        for cluster in trajectory.chemical_system._clusters[grp]
+                        for x in cluster
+                        if x in trajectory.atom_indices
+                    )
```
oerc0122:
```suggestion
                    {
                        trajectory.atom_types[x]
                        for cluster in trajectory.chemical_system._clusters[grp]
                        for x in cluster
                        if x in trajectory.atom_indices
                    }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+        for grp_i, grp_j in it.combinations_with_replacement(
+            trajectory.group_lookup, 2
+        ):
+            eles_i = sorted(
+                set(
+                    trajectory.atom_types[x]
+                    for cluster in trajectory.chemical_system._clusters[grp_i]
+                    for x in cluster
+                    if x in trajectory.atom_indices
+                )
```
oerc0122:
```suggestion
                {
                    trajectory.atom_types[x]
                    for cluster in trajectory.chemical_system._clusters[grp_i]
                    for x in cluster
                    if x in trajectory.atom_indices
                }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+            )
+            eles_j = sorted(
+                set(
+                    trajectory.atom_types[x]
+                    for cluster in trajectory.chemical_system._clusters[grp_j]
+                    for x in cluster
+                    if x in trajectory.atom_indices
+                )
```
oerc0122:
```suggestion
                {
                    trajectory.atom_types[x]
                    for cluster in trajectory.chemical_system._clusters[grp_j]
                    for x in cluster
                    if x in trajectory.atom_indices
                }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+            conc_j = trajectory.group_lookup[grp_j] / tot_n_atms
+            if grp_i != grp_j:
+                # for the cross terms we divide by 2 since f(q,t)_OH
+                # includes only OH or HO, it gets summed to the total
+                # properly by the weight scheme. Similarly, we will
+                # have a factor of two in the scaling factor for the
+                # cross terms of the molecular case.
+                conc = 2 * (conc_i * conc_j) ** conc_exp
+            else:
+                conc = (conc_i * conc_j) ** conc_exp
```
oerc0122:
```suggestion
            conc = (conc_i * conc_j) ** conc_exp
            if grp_i != grp_j:
                # for the cross terms we divide by 2 since f(q,t)_OH
                # includes only OH or HO, it gets summed to the total
                # properly by the weight scheme. Similarly, we will
                # have a factor of two in the scaling factor for the
                # cross terms of the molecular case.
                conc *= 2
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+    Returns
+    -------
+    list[tuple[str, str]]
+        A list of label pairs.
+    """
+    if all_pairs:
+        iterable = it.product(labels, repeat=2)
+    else:
+        iterable = it.combinations_with_replacement(labels, 2)
+    return sorted(iterable)
```
oerc0122: Try to avoid materialising large generators unless absolutely necessary. Sorted must materialise and store.

If you want sorted generators here, sort the input `labels` and the generators will use that order.

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+        return labels
+
+    if intra:
+        for grp in trajectory.group_lookup:
+            eles = sorted(
+                set(
+                    trajectory.atom_types[index]
+                    for cluster in trajectory.chemical_system._clusters[grp]
+                    for index in cluster
+                )
```
oerc0122:
```suggestion
                {
                    trajectory.atom_types[index]
                    for cluster in trajectory.chemical_system._clusters[grp]
                    for index in cluster
                }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+            )
+            for ele_i, ele_j in label_pairs(eles, all_pairs=all_pairs):
+                pair_label = f"<{grp}>/{ele_i}{ele_j}"
+                label_i = f"<{grp}>/{ele_i}"
+                label_j = f"<{grp}>/{ele_j}"
+                labels.append((pair_label, (label_i, label_j)))
+        return labels
+
+    for grp_i, grp_j in label_pairs(
+        trajectory.group_lookup.keys(), all_pairs=all_pairs
```
oerc0122:
```suggestion
        trajectory.group_lookup, all_pairs=all_pairs
```
Strictly if this is a dictionary don't need the `keys`

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+    ):
+        eles_i = sorted(
+            set(
+                trajectory.atom_types[index]
+                for cluster in trajectory.chemical_system._clusters[grp_i]
+                for index in cluster
+            )
```
oerc0122:
```suggestion
            {
                trajectory.atom_types[index]
                for cluster in trajectory.chemical_system._clusters[grp_i]
                for index in cluster
            }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+        eles_j = sorted(
+            set(
+                trajectory.atom_types[index]
+                for cluster in trajectory.chemical_system._clusters[grp_j]
+                for index in cluster
+            )
```
oerc0122:
```suggestion
            {
                trajectory.atom_types[index]
                for cluster in trajectory.chemical_system._clusters[grp_j]
                for index in cluster
            }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+                output_data[f"{name}/{ele_i}{ele_j}"][...] = result
+        return
+
+    for grp_i, grp_j in it.combinations_with_replacement(trajectory.group_lookup, 2):
+        eles_i = sorted(
+            set(
+                trajectory.atom_types[index]
+                for cluster in trajectory.chemical_system._clusters[grp_i]
+                for index in cluster
+            )
```
oerc0122:
```suggestion
            {
                trajectory.atom_types[index]
                for cluster in trajectory.chemical_system._clusters[grp_i]
                for index in cluster
            }
```

MDANSE/Src/MDANSE/Framework/AtomGrouping/grouping.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+        eles_j = sorted(
+            set(
+                trajectory.atom_types[index]
+                for cluster in trajectory.chemical_system._clusters[grp_j]
+                for index in cluster
+            )
```
oerc0122:
```suggestion
            {
                trajectory.atom_types[index]
                for cluster in trajectory.chemical_system._clusters[grp_j]
                for index in cluster
            }
```

MDANSE/Src/MDANSE/Framework/Configurators/AtomSelectionConfigurator.py
oerc0122, line 1: Nice to see some big clear out of junk code.

MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py
```diff
@@ -147,11 +147,15 @@ def configure(self, value: str):
             self.error_status = "Key of transmutation map should be castable to int"
             return
 
-        for idx, element in value.items():
-            if idx not in idxs:
-                self.error_status = "Inputted setting not valid - atom index not found in the current system."
-                return
+        keys = {index for index in value.keys()}
```
oerc0122:
```suggestion
        keys = set(value.keys())
```
Although it should be noted that [`KeysView`s](https://docs.python.org/3/library/stdtypes.html#dict-views) behave like sets in most circumstances

MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py
```diff
@@ -147,11 +147,15 @@ def configure(self, value: str):
...
+        if not keys.issubset(idxs):
+            self.error_status = (
+                "Input setting not valid - atom index not found in the current system."
+            )
+            return
 
+        elements = {element for element in value.values()}
```
oerc0122:
```suggestion
        elements = set(value.values())
```

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -186,6 +186,31 @@ def initialize(self):
...
+        grouping = self.configuration.get("grouping_level")
+        if trajectory is None:
+            return
+        self.trajectory = trajectory["instance"]
+        if selection is not None:
+            self.trajectory.set_selection(selection["flatten_indices"])
+        if transmutation is not None:
+            self.trajectory.set_transmutation(transmutation.transmutation)
+        if grouping is not None:
+            self.trajectory.set_grouping(grouping["level"])
```
```suggestion
        if (trajectory := self.configuration.get("trajectory")) is None:
            return
        self.trajectory = trajectory["instance"]
        if (selection := self.configuration.get("atom_selection")) is not None:
            self.trajectory.set_selection(selection["flatten_indices"])
        if (transmutation := self.configuration.get("atom_transmutation")) is not None:
            self.trajectory.set_transmutation(transmutation.transmutation)
        if (grouping := self.configuration.get("grouping_level")) is not None:
            self.trajectory.set_grouping(grouping["level"])
```
Could use walrus here.

MDANSE/Src/MDANSE/Framework/Jobs/McStasVirtualInstrument.py
```diff
@@ -135,56 +135,34 @@ def initialize(self):
...
         self._mcStasPhysicalParameters = {"density": 0.0}
         self._mcStasPhysicalParameters["V_rho"] = 0.0
         self._mcStasPhysicalParameters["weight"] = sum(
-            [
-                self.configuration["trajectory"]["instance"].get_atom_property(
-                    s, "atomic_weight"
-                )
-                for s in symbols
-            ]
+            [self.trajectory.get_atom_property(s, "atomic_weight") for s in symbols]
```
oerc0122:
```suggestion
            self.trajectory.get_atom_property(s, "atomic_weight") for s in symbols
```

ChiCheng45 (commented):
Looks good, a few issues so far.

- With the DFTB water trajectory running molecule grouping with only one type of atom selected causes a job failure. This worked previously so that grouping with only H atoms selected gives only the /dos/<H2_O1>/H results.
- Molecule grouping can be run on a system which does not have molecules in it. Previously the GUI would block molecule grouping from being used.
- "Show trajectory attenuation" in the trajectory filter filter designer error - actually this also doesn't work on protos (#889) but there is now an error from the atoms selection part.

ChiCheng45:
I've found a difference between protos and this branch that you may want to fix.

On protos with atom selection, only selected atoms types are saved into the results. On this branch all atom types from the original trajectory are saved even when they are not selected. Unselected atom type results will be zero.

MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
```diff
@@ -78,13 +78,18 @@ class TrajectoryFilter(IJob):
         "AtomSelectionConfigurator",
         {"dependencies": {"trajectory": "trajectory"}},
     )
+    settings["atom_transmutation"] = (
```
oerc0122, line 81: You seem to be adding `atom_transmutaion`, but is it used anywhere? Does trajectory implicitly transform?

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+        self._element_from_label = {}
+        self._atom_names = {}
         self._atom_cache = {}
+        self._selection = []
+        self.selection_getter = None
+        self._transmutation = {}
+        self._transmuted_types = []
+
+    @property
+    def atom_indices(self) -> Sequence[int]:
```
```suggestion
    def atom_indices(self) -> list[int]:
```
Since this is explicitly a list.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+    @property
+    def atom_types(self) -> Sequence[str]:
+        """Chemical elements of ALL atoms, with transmutation applied."""
+        if self._transmuted_types:
+            return self._transmuted_types
+        if not self._transmutation:
+            return self._trajectory.chemical_system.atom_list
+        temp = copy.deepcopy(self._trajectory.chemical_system.atom_list)
+        for index, type in self._transmutation.items():
+            temp[index] = type
```
oerc0122, line 89:
```suggestion
        temp = copy.deepcopy(self._trajectory.chemical_system.atom_list)
        temp.update(self._transmutation)
```
If `temp` is a dictionary.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+                self._atom_names = copy.deepcopy(self.atom_types)
+                for k, v in temp_names.items():
+                    self._atom_names[k] = v
+        return self._atom_names
+
+    @property
+    def element_from_label(self) -> dict[str, str]:
+        """Mapping from atom label to its chemical element name."""
+        if not self._element_from_label:
+            self.atom_names
```
```suggestion
            return self.atom_names
```
Is this meant to have a `return`?

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+    @property
+    def group_lookup(self) -> dict[str, int] | dict[str, list[int]]:
+        """Dictionary of currently existing groups.
+        
+        The keys are names of the group. The values can be the count
+        of all atoms belonging to the group, or (for 'each molecule' only)
+        a list of atom indices belonging to the individual molecule.
+        """
+        if not self._group_lookup:
+            self.atom_names
```
```suggestion
            return self.atom_names
```

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+                    ):
+                        if not set(cluster).issubset(self.atom_indices):
+                            continue
+                        self._group_lookup[mol_name] += len(cluster)
+                        for x in cluster:
+                            temp_names[x] = (f"<{mol_name}>/{self.atom_types[x]}")
+                self._element_from_label = {v: self.atom_types[k] for k, v in temp_names.items()}
+                self._atom_names = copy.deepcopy(self.atom_types)
+                for k, v in temp_names.items():
+                    self._atom_names[k] = v
```
```suggestion
                self._atom_names.update(temp_names)
```

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+            return self._transmuted_types
+        if not self._transmutation:
+            return self._trajectory.chemical_system.atom_list
+        temp = copy.deepcopy(self._trajectory.chemical_system.atom_list)
+        for index, type in self._transmutation.items():
+            temp[index] = type
+        self._transmuted_types = temp
+        return self._transmuted_types
+
+    @property
```
oerc0122: By setting "_atom_names" in this function are you just wanting `functools.cached_property`?

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+                self.get_all_natoms(),
+                self.atom_names,
+                self.atom_types,
+            ),
+        ]:
+            w = defaultdict(float)
+            for name, element in zip(atm_names, atm_elements):
+                w[name] += self._trajectory.get_atom_property(element, prop)
+            for element, num_atoms in n_elements.items():
+                w[element] /= num_atoms
```
oerc0122: Are we sure this is correct? The switch of `name` to `element` worries me that they may not align. If they *are* the same keys, suggest using `name` in place of `element` in the second loop.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,11 +60,223 @@ def __init__(self, filename, trajectory_format=None):
...
+        Returns
+        -------
+        dict
+            A dictionary of the number of atom per element.
+
+        """
+        all_atom_names = self.atom_names
+        if self._selection:
+            return Counter(self.selection_getter(all_atom_names))
+        return Counter(all_atom_names)
```
oerc0122: Any reason for the temporary name assign?
```suggestion
        if self._selection:
            return Counter(self.selection_getter(self.atom_names))
        return Counter(self.atom_names)
```

MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
MBartkowiakSTFC: In this PR, transmutation is applied to the trajectory by `IJob.set_up_trajectory()`, as long as both the transmutation and trajectory are present. (Before the changes, transmutation would modify the atom types inside the `AtomSelectionConfigurator`.)

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
MBartkowiakSTFC: `temp` is a copy of `ChemicalSystem.atom_list`, which is really a list this time.

MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py
```diff
@@ -147,14 +147,14 @@ def configure(self, value: str):
             self.error_status = "Key of transmutation map should be castable to int"
             return
 
-        keys = {index for index in value.keys()}
+        keys = set(value.keys())
```
oerc0122, line 150: `KeysView`s already behave like `set`s so there may be no need to cast if not used elsewhere.

MDANSE/Src/MDANSE/Framework/Jobs/McStasVirtualInstrument.py
```diff
@@ -141,23 +141,23 @@ def initialize(self):
         self._mcStasPhysicalParameters = {"density": 0.0}
         self._mcStasPhysicalParameters["V_rho"] = 0.0
         self._mcStasPhysicalParameters["weight"] = sum(
-            [self.trajectory.get_atom_property(s, "atomic_weight") for s in symbols]
+            self.trajectory.get_atom_property(s, "atomic_weight") for s in symbols
         )
         self._mcStasPhysicalParameters["sigma_abs"] = (
             np.mean(
-                [self.trajectory.get_atom_property(s, "xs_absorption") for s in symbols]
+                self.trajectory.get_atom_property(s, "xs_absorption") for s in symbols
```
oerc0122: This probably won't work. `np.mean` requires an ArrayLike rather than generators.
```
>>> import numpy as np
>>> np.mean(i for i in range(10))
  File "<stdin>", line 1, in <module>
  File "/home/jacob/.pyenv/versions/3.10.14/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 3504, in mean
    return _methods._mean(a, axis=axis, dtype=dtype,
  File "/home/jacob/.pyenv/versions/3.10.14/lib/python3.10/site-packages/numpy/core/_methods.py", line 131, in _mean
    ret = ret / rcount
TypeError: unsupported operand type(s) for /: 'generator' and 'int'
```

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -66,88 +67,91 @@ def __init__(self, filename, trajectory_format=None):
...
+        If grouping is used, atom labels may contain the molecule name
+        as well as the chemical element of the atom, and will not match
+        the entries in the atom database. This dictionary allows to get
+        a valid atom database key for each atom in the system.
+        """
+        mapping = {element : element for element in self.unique_elements}
+        if self._grouping_level == "molecule":
+            temp_names = {}
+            for mol_name in self.chemical_system._clusters:
+                for cluster in self.chemical_system._clusters[mol_name]:
```
```suggestion
            for mol_name, clusters in self.chemical_system._clusters.items():
                for cluster in clusters:
```

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -66,88 +67,91 @@ def __init__(self, filename, trajectory_format=None):
...
+        elif self._grouping_level == "molecule":
+            for mol_name in self.chemical_system._clusters:
+                temp_dict.setdefault(mol_name, 0)
+                for mol_number, cluster in enumerate(
+                    self.chemical_system._clusters[mol_name],
+                ):
+                    overlap = set(cluster).intersection(self.atom_indices)
+                    if not overlap:
+                        continue
+                    temp_dict[mol_name] += len(overlap)
```
oerc0122: `if not overlap` the `len` will be `0`, so may just want to remove the `if` for simpler code?

MDANSE/Src/MDANSE/Framework/Configurators/AtomTransmutationConfigurator.py
MBartkowiakSTFC: The main reason to make it a set is to use the `.issubset()` method in the next line. Without making it a set, I would get `AttributeError: 'dict_keys' object has no attribute 'issubset'`

MBartkowiakSTFC:
As far as I can tell, I implemented all the changes, except those where I wrote down my reasons not to.

oerc0122 (approved):

PR #876, opened 2025-07-15
Updated to GUI so that the 3D viewer side panel can be hidden.

ChiCheng45:
**Description of work**
Updated to GUI so that the 3D viewer side panel can be hidden. This should be helpful in the atom selection and related helper dialogs where a larger 3D view might be needed. 

Peek 2025-07-15 15-24
Peek 2025-07-15 15-25

**To test**
Check that 3D view and viewer side panel continue to work in the trajectory tab and the selection and related helpers.


MBartkowiakSTFC (approved):
It works for me.

One thing we may want in the future is a way to remind the user that the side panel was there, but this is a minor detail.

PR #878, opened 2025-07-15
Invert icon colours if in dark mode

### LINKED ISSUE #874: [BUG] New periodic table icon is hard to see when dark mode is used.
**Description of the error**
The periodic table icon is hard to see when dark mode is used. It can just about be made out when the mouse is hovered over it.

**Additional details**

Image

Image

MBartkowiakSTFC:
**Description of work**
Since some of the icons do not work well in dark mode, we create two versions of each icon (original and with inverted colours), and use the lighter version when in dark mode, and darker version otherwise.

closes #874

**Fixes**
1. Check if the system is using dark mode on GUI startup.
2. Check if icons need to be changed when changing the style sheets.
3. Compare each icon to its colour-inverted version, and assign them to dark/light mode based on their RGBA values.

**To test**
Start the GUI using different system colour profiles. Change the style sheets manually to a dark setting and back. Check that at each step the icons are easily recognised.


MBartkowiakSTFC:
I added the background to the periodic table icon, but for now I've kept the icon colour inversion mechanism in.

Also, I removed most of the icons from MDANSE_GUI, and replaced the one for GUI customisation.

ChiCheng45:
Great seems to look fine in dark mode.
image

and in light mode.

image

Just one thing is that the units icon arrows are difficult to see in both dark and light mode.

MBartkowiakSTFC:
> Just one thing is that the units icon arrows are difficult to see in both dark and light mode.

I think that this has always been the case, but you are right that this is a good moment to fix it. I'll make the arrows thicker.

ChiCheng45:
> > Just one thing is that the units icon arrows are difficult to see in both dark and light mode.
> 
> I think that this has always been the case, but you are right that this is a good moment to fix it. I'll make the arrows thicker.

I think the issue is the color of the arrows.

ChiCheng45 (approved):
Looks good.

Issue #877, opened 2025-07-15
[BUG] Partial charges input field cannot hold the entire JSON string


MBartkowiakSTFC:
**Description of the error**
For trajectories with enough atoms, it will become impossible to set the charges in the GUI, as the JSON string becomes truncated.

Here I had 4000 atoms in the trajectory, but I could only set the charges for 2583 of them.

**Describe the expected result**
It should be possible to set the partial charges for any number of atoms.

**Describe the actual result**

<img width="526" height="349" alt="Image" src="https://github.com/user-attachments/assets/f9c2a68c-00de-4688-a891-2d82db5bf5ae" />
The helper widget (above) shows all the charges to be set.
The actual input in the main widget ends suddenly:
<img width="1002" height="77" alt="Image" src="https://github.com/user-attachments/assets/62a264e9-5b84-46ec-8abd-406ad92b82da" />

**Suggested fix**
Temporarily, we could relax the limits on the maximum string length in the main widget's QLineEdit.

In the long run, it would also be good to switch to a more compact way of writing the charges.

**Additional details**
N/A

PR #879, opened 2025-07-16
Improve input of partial charges

MBartkowiakSTFC:
**Description of work**
In order to deal with #877, changes have been made to the partial charge input.

This PR does not completely remove the limitations on the number of atoms in a trajectory. However, it makes it possible to use larger trajectories than before.

**Fixes**
1. Changed the input length limit of the partial charge QLineEdit to 2147483647.
2. Enabled new format of partial charges, where dictionary keys are charges, and dictionary values are lists of atom indices:
`{"charge1": [ind1, ind2, ...], "charge2": [indx, indx+1, ...]}`

**To test**
Unit tests must pass.
Try setting charges for a trajectory containing molecules. Check that running from GUI and saving to script both work.


ChiCheng45 (approved):
Looks good. Works fine on my end.

PR #881, opened 2025-07-16
Load trajectories in a background thread

MBartkowiakSTFC:
**Description of work**
Loading large trajectories can block the GUI for a noticeably long time. This PR moves the trajectory load operation to a background thread.

This PR should improve the responsiveness of the GUI, but does NOT solve the problem with input widgets in `Action` loading slowly when a new trajectory/analysis is picked.

**Fixes**
1. `TrajectoryModel` replaces `GeneralModel` for storing trajectories.
2. `Trajectory` instances are created and stored in `TrajectoryModel`, and then re-used by other parts of the GUI (i.e. `Action`).
3. Input widgets in `Action` only re-run `.configure(input)` if input has changed. Output file widgets are an exception, because they need to re-run configure to check if the output file names are not protected by the GUI.
4. Each entry in TrajectoryModel starts as a placeholder saying "Loading TRAJECTORY NAME...". This gets renamed once the trajectory has been loaded.

**To test**
Manual testing of the GUI is necessary to confirm that no unexpected behaviour occurs during the time when a trajectory placeholder item is already on the list, but the trajectory is not ready.
Some of the operations that may cause problems:
- loading several trajectories,
- checking that the analysis job updates when the trajectory has been loaded,
- deleting trajectories.


MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -231,6 +231,23 @@ def configure(self, value: str):
...
+            Input parameters as string
+
+        Returns
+        -------
+        bool
+            If True, self.configure(new_input) needs to be run
+        """
+        if self.configured and self._original_input == new_input:
+            return False
+        return True
```
```suggestion
        return self.configured and self._original_input == new_input
```

MDANSE/Src/MDANSE/Framework/Configurators/InputFileConfigurator.py
```diff
@@ -37,6 +37,8 @@ def __init__(self, name, wildcard="All files (*)", **kwargs):
         # The base class constructor.
         IConfigurator.__init__(self, name, **kwargs)
 
+        self._instance = kwargs.get("instance")
```
oerc0122: Instead of hiding this in kwargs, would prefer to see it explicit.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/HDFTrajectoryWidget.py
```diff
@@ -13,31 +13,40 @@
...
-            self._configurator.configure(str(filename))
+        if trajectory_instance:
+            self._configurator._instance = trajectory_instance
+            self._configurator.configure_from_instance()
+            filename = trajectory_instance.filename
+        else:
+            try:
+                filename = source_object
+            except AttributeError:
+                filename = None
```
oerc0122: How would this raise an `AttributeError`?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -0,0 +1,156 @@
...
+    free_name = Signal(str)
+
+    def __init__(self, parent: QObject = None):
+        super().__init__(parent=parent)
+        self.mutex = QMutex()
+        self._node_numbers = []
+        self._trajectory_paths = {}
+        self._trajectory_instances = {}
+        self._loading_threads = {}
+        self._next_number = 0
```
oerc0122: Probably want to use `itertools.count` with `next`
```suggestion
        self._next_number = itertools.count()
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @Slot(tuple)
+    def append_object(self, input: tuple):
+        full_name, label = input
+        self._node_numbers.append(self._next_number)
+        self._trajectory_paths[self._next_number] = full_name
+        self._trajectory_instances[self._next_number] = None
+        retval = int(self._next_number)
+        self._next_number += 1
```
oerc0122: With above change
```suggestion
        retval = next(self._next_number)
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -0,0 +1,156 @@
...
+        thread.results.connect(self.accept_results)
+        self._loading_threads[index] = thread
+        thread.start()
+
+    def get_trajectory(self, index: int):
+        result = None
+        if index not in self._loading_threads:
+            LOG.info("Requesting a missing trajectory with index %s", index)
+            return result
+        elif index not in self._trajectory_instances:
```
```suggestion
        if index not in self._trajectory_instances:
```
Because the other is a `return`

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -0,0 +1,156 @@
...
+            return f"Loading trajectory {self._trajectory_paths[index]}"
+        return self._trajectory_instances[index]
+
+    @Slot(object)
+    def accept_results(self, result_tuple):
```
```suggestion
    def accept_results(self, result_tuple) -> None:
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -0,0 +1,156 @@
...
+        if isinstance(traj, Trajectory):
+            return True
+        return False
+
+    def removeRow(self, row: int, parent: QModelIndex = None):
+        try:
+            node_number = self.item(row).data()
+        except AttributeError:
+            return
+        if parent is None:
```
oerc0122: Is the signature: `super().removeRow(row, parent=None)`? In which case you can just pass parent regardless.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/HDFTrajectoryWidget.py
MBartkowiakSTFC: You're right, this is just a leftover from the time when the line above was `filename = source_object.filename`.

oerc0122 (approved):

PR #883, opened 2025-07-17
Add the missing character in IJob.save

### LINKED ISSUE #880: [BUG] Scripts don't work
**Description of the error**
Scripts saved by the current version of MDANSE don't work, as there is a `)` missing in the line where the job instance gets created.
https://github.com/ISISNeutronMuon/MDANSE/blob/9a64024480b863ee0e827a2e534956abd20b1d6e/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py#L253

MBartkowiakSTFC:
**Description of work**
Corrects the typo causing scripts to fail.

closes #880 

**Fixes**
1. Close the bracket in the `IJob.create(...` part of the script template.

**To test**
Save a script from the GUI and run it. It should run the same as when started from the GUI.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -250,7 +250,7 @@ parameters = {{
 ########################################################
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


ChiCheng45 (approved):
Looks good, seems to be fixed.

PR #884, opened 2025-07-17
Adjust 3D view side panel size policy

ChiCheng45:
**Description of work**
Adjusted 3D view side panel size policy. This should ensure that the 3D view expands to fill the GUI.

image

**To test**
Check 3D view and controls.


MBartkowiakSTFC (approved):
I checked that the VTK 3D view is using more space now than before. Looks good.

PR #890, opened 2025-07-22
Fixed 3d view controls while animation is playing

### LINKED ISSUE #882: [BUG] Changing the "time per frame" or "show every N frames" setting causes the 3D view to play
**Description of the error**
Changing the "time per frame" or "show every N frames" setting causes the 3D view to play. It shouldn't play if it is stopped but should continue to play if it is playing.

Image

ChiCheng45:
**Description of work**
Fixed the 3D view so that when the "time per frame" or "show every N frames" are changed the 3D view does not start playing if it was not already playing.

Also fixed the 3D view so that when it is fast forwarding or backwarding then changes the "time per frame" or "show every N frames" setting leaves it fast forwarding or backwarding but with the new "time per frame" or "show every N frames" options.

**Fixes**
Fixes #882

**To test**
- 3D view should not play when "time per frame" or "show every N frames" are changed.
- 3D view should continue to play when "time per frame" or "show every N frames" are changed if it was playing.
- 3D view should continue fast forwarding or backwarding when "time per frame" or "show every N frames" are changed.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
@@ -106,6 +106,7 @@ class ViewerControls(QWidget):
         self._animation_timer.timeout.connect(self.advance_frame)
         self._mutex = QMutex()
         self._frame_step = 1
+        self._current_step_size = 1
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
Looks good. I tried changing different display options, and it did not switch the animation on or off.

PR #895, opened 2025-07-24
Fix PPS preview in TrajectoryFilter

### LINKED ISSUE #889: [BUG] Trajectory filter "Show trajectory attenuation" option no longer works
**Description of the error**
When the "Show trajectory attenuation" option is selected in the filter designer the window closes with an error seen in the console.

```
    weight_dict = get_weights(weights, nAtomsPerElement, 1)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: get_weights() missing 2 required positional arguments: 'all_contents' and 'dim
```

### LINKED ISSUE #894: [BUG] Potential issue in `Signal`
**Description of the error**
In [`MDANSE.Mathematics.Signal.power_spectrum`](https://github.com/ISISNeutronMuon/MDANSE/blob/0e9402bee3bfdd1e34e4c29a65391e98318acd7a/MDANSE/Src/MDANSE/Mathematics/Signal.py#L1397) it calls `get_weights` with what appears to be invalid arguments (passing 1 where it should pass a dict)

Is this a failure? Is this unused, untested or simply undocumented?

MBartkowiakSTFC:
**Description of work**
Recent changes to the way the analysis jobs handle their input caused the Position Power Spectrum preview in the Trajectory Filter to stop working. This PR aims to solve this problem, and make some adjustments to the way the preview works.

closes #889
closes #894

**Fixes**
1. The PPS in the preview is now calculated using a PositionPowerSpectrum job instance, and not in a custom function in `Signal`.
2. The PPS calculation runs in a background thread. This should start immediately when TrajectoryFilter is created, to reduce the waiting time when a user checks the preview checkbox.
3. The PPS should NOT be recalculated if the preview box has been unchecked and checked again, unless the input parameters of the filter have been changed in the meantime.
4. Instead of resampling the calculated PPS to match the fine grid of the filter attenuation curve, the filter attenuation is interpolated to match the romega axis of the PPS. This way the PPS in the preview matches the one in the PPS job.

**To test**
1. Load a trajectory and calculate the Position Power Spectrum.
2. Go to TrajectoryFilter and open the filter creator dialog. Check the "Show trajectory attenuation" box. The GUI should remain responsive during the calculation, and the result in the plot should match the result of the Position Power Spectrum job calculated in step 1.


MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
```diff
@@ -358,18 +409,34 @@ def as_grid(self) -> QGridLayout:
...
+
+        The checkbox will not be possible to uncheck while the calculation
+        is running, and the label text will inform the user that the
+        calculation is in progres..
+        """
+        self.pps_checkbox.setEnabled(enable)
+        if enable:
+            self.pps_label.setText("Show trajectory attenuation")
+        else:
+            self.pps_label.setText("Calculating PPS...")
```
RobBuchananCompPhys: It's neat that the text displays in place of the label, but I previously thought about handling this with a `QMessageBox`. Would it be preferable to use one instead, with a timeout of 2000ms or something in that ball park?
The downside of that compared to temporarily overwriting the label as you have done, is that a message box may obscure the UI even once the data is loaded.

MBartkowiakSTFC: My preference is normally to give a user _some_ indication that the some operation has been triggered, but without getting in the user's way. So far, if MDANSE uses a message box, it normally comes with a check box which effectively says "don't show this again". A `QMessageBox` would do the job, of course.

oerc0122:
```suggestion
        message = "Show trajectory attenuation" if enable else "Calculating PPS..."
        self.pps_label.setText(message)
```
Is this clearer?

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
```diff
@@ -1027,15 +1092,65 @@ def resample_and_normalise(self, values, to_len):
         """
         return signal.resample(values, to_len) / values.max()
 
+    @Slot(object)
+    def accept_results(self, res_dict: dict[str, np.ndarray]):
```
oerc0122:
```suggestion
    def accept_results(self, res_dict: dict[str, npt.NDArray[float]]):
```
Gives better typing information to type checkers.

oerc0122 (approved):
Looks like a nice simplification. Just a couple of stylistic suggestions.

ChiCheng45 (approved):
Looks good.

PR #897, opened 2025-07-24
Added option to display atom labels in the 3D viewer

### LINKED ISSUE #499: [ENHANCEMENT] Add atom labels to 3D viewer.
Add labels to the atoms in the 3D viewer. This should help the user with atom selection for example.

Atom labels should be drawn next to the atom in the 3D viewer.

**Additional context**
Here is an example from the Mercury program.

image

ChiCheng45:
**Description of work**
Adds atom labels to the 3D viewer. Currently supports the following options.

- none
- index
- label
- atom
- molecule

For the cases where there are no labels/molecules no labels are shown in the 3D view. I find that there is a slight slow down to the 3D viewer when playing but I think this should be acceptable since in most cases you want be playing the 3D view while looking at the labels.

Let me know if you have any suggestion for more labels or have any options on font size etc.

image

image

**Fixes**
Closes #499

**To test**
Check labels in the 3D view including atom selection and transmutation. Check that labels update correctly when switching or deleting the trajectory.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -116,6 +116,7 @@ class MolecularViewer(QtWidgets.QWidget):
         self._label_renderer = vtk.vtkRenderer()
         self._label_renderer.SetLayer(1)
         self._label_renderer.SetBackgroundAlpha(0)
+        self._label_renderer.SetInteractive(False)
 
         self._iren.GetRenderWindow().SetNumberOfLayers(2)
         self._iren.GetRenderWindow().AddRenderer(self._renderer)
@@ -1147,7 +1148,7 @@ class MolecularViewerExtended(MolecularViewer):
         if self._last_coords is None:
             return
 
-        picker = vtk.vtkPropPicker()
+        picker = vtk.vtkCellPicker()
 
         picker.AddPickList(self.atom_actor)
         picker.PickFromListOn()
@@ -1200,7 +1201,7 @@ class MolecularViewerWithPicking(MolecularViewer):
         if self._picking_domain is None:
             return
 
-        picker = vtk.vtkPropPicker()
+        picker = vtk.vtkCellPicker()
 
         picker.AddPickList(self._picking_domain)
         picker.PickFromListOn()
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -623,7 +623,9 @@ class MolecularViewer(QtWidgets.QWidget):
         if not self.atom_label_actors or self.atom_label_type == "none":
             return
 
-        for follower, coord in zip(self.atom_label_actors, self._reader.read_frame(self._current_frame)):
+        for follower, coord in zip(
+            self.atom_label_actors, self._reader.read_frame(self._current_frame)
+        ):
             follower.SetPosition(*coord)
 
     def clear_atom_labels(self):
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -16,6 +16,7 @@
 import copy
 from typing import Any, Optional
 
+import more_itertools
 import numpy as np
 import vtk
 from qtpy import QtWidgets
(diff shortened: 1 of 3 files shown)


MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -273,6 +285,20 @@ def _change_axes(self, axes_option: str):
         self._iren.GetRenderWindow().Render()
         self._iren.Render()
 
+    def _change_atom_labels(self, label_option: str):
```
```suggestion
    def _change_atom_labels(self, label_option: str) -> None:
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -531,9 +557,91 @@ def clear_trajectory(self, clear_isosurfaces=True):
 
         del self._actors
 
+    def create_atom_label_actors(self):
+        """Creates atom label actors, setting the text to the chosen
+        atom_label_type.
+        """
+        self.atom_label_actors = []
+        if self._reader is None:
+            return
```
oerc0122, line 568: Silently returning may not be the best thing here, do we want to raise or warn on invalid/conflicting input?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -531,9 +557,91 @@ def clear_trajectory(self, clear_isosurfaces=True):
...
+        elif self.atom_label_type == "label":
+            label_dict = self._reader._trajectory.chemical_system._labels
+            if not label_dict:
+                return
+            keys = []
+            vals = []
+            for k, v in label_dict.items():
+                keys += [k] * len(v)
+                vals += v
+            labels, _ = zip(*sorted(zip(keys, vals), key=lambda x: x[1]))
```
```suggestion
            keys = more_itertools.run_length.decode(((k, len(v)) for k, v in label_dict.items()))
            labels = sorted(keys, key=label_dict.__getitem__)
```
Should do the same

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -531,9 +557,91 @@ def clear_trajectory(self, clear_isosurfaces=True):
...
+            label_dict = self._reader._trajectory.chemical_system._clusters
+            if not label_dict:
+                return
+            keys = []
+            vals = []
+            for k, v in label_dict.items():
+                indices = [i for j in v for i in j]
+                keys += [k] * len(indices)
+                vals += indices
+            labels, _ = zip(*sorted(zip(keys, vals), key=lambda x: x[1]))
```
```suggestion
            label_dict = {k: list(more_itertools.collapse(v)) for k, v in label_dict.items()}
            keys = more_itertools.run_length.decode(((k, len(v)) for k, v in label_dict.items()))
            labels = sorted(keys, key=label_dict.__getitem__)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -531,9 +557,91 @@ def clear_trajectory(self, clear_isosurfaces=True):
...
+
+        self.atom_label_actors = atom_label_actors
+
+    def update_atom_label_actors(self):
+        """Updates the atom label follwer positions."""
+        if self._reader is None:
+            return
+
+        if not self.atom_label_actors or self.atom_label_type == "none":
+            return
```
```suggestion
        if (
            self._reader is None
            or not self.atom_label_actors
            or self.atom_label_type == "none"
        ):
            return
```

oerc0122 (approved):
Just a couple of stylistic things, looks good overall if it does what it says it does.

MBartkowiakSTFC (commented):
Please test one thing for me: set labels to index, and change the side panel to atom trace. Normally, you should be able to set the atom index by clicking an atom in the 3D view. On my machine, it seems that the presence of the labels affects the accuracy of selecting atoms by clicking.

This does not necessarily need fixing, since the labels allow you to read the index and type it into the selection field.

ChiCheng45:
> Please test one thing for me: set labels to index, and change the side panel to atom trace. Normally, you should be able to set the atom index by clicking an atom in the 3D view. On my machine, it seems that the presence of the labels affects the accuracy of selecting atoms by clicking.
> 
> This does not necessarily need fixing, since the labels allow you to read the index and type it into the selection field.

Thanks for that, should be fixed now. There seems to be some bug or strange interaction with atom picking and the label renderer that is layered over the main renderer. I think the latest commit should have sorted it out. Can you test on your machine.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: Maybe, I think we should leave this for until we decide to refactor the MolecularViewer. Its possible for the reader to be None if there is no trajectory loaded in the 3D viewer. I return none here because an attempt to create atom labels occurs when the atom label option is changed. This kinda of thing is done in MolecularViewer quite often.

MBartkowiakSTFC (approved):
Everything works fine now.

PR #903, opened 2025-07-28
Rework plotter to add visible range to 'Use it?'

### LINKED ISSUE #902: [ENHANCEMENT] Set a default value for the "Use it?" text column
Currently the `Use it?` text box is initially blank and thus invisible. It would be easy to make it display 1-n to show users that it can be set.

Default value for textbox is 1-n

N/A

**Additional context**
N/A

oerc0122:
**Description of work**
Add default range to `Use it` column on data load
Clean up plotting context

**Fixes**
Fixes #902 

**To test**
Check plotting context makes sense. For 3D datsets will display product of axis lengths.


MBartkowiakSTFC (approved):
Looks good.

Issue #891, opened 2025-07-22
[BUG] Adjusting the job cell widths while a job is running


ChiCheng45:
**Description of the error**
The job cell widths are unable to be changed while a job is running. Whenever the cells information updates it jumps back to some default width.

**Additional details**
![Image](https://github.com/user-attachments/assets/6d0aec14-c3d3-4536-97e1-b10b6a454da6)

oerc0122:
This is a side-effect of the `setText` rescaling the bar in `JobHolder.update_fields`

PR #908, opened 2025-07-30
Add  timings to GUI

### LINKED ISSUE #559: [ENHANCEMENT] Add job running and completion time.
Add approximate completion and running time to the jobs tab table. We should be able to work this out based on the number of steps it has already completed.

oerc0122:
**Description of work**
Add timings to GUI progress bars in step/s and %/s.
Add estimate on completion.
Add timing report in info box.

**Fixes**
Fixes #559 

**To test**

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/JobStatus.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/JobStatus.py
@@ -13,10 +13,12 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
+from __future__ import annotations
+
 import time
 from dataclasses import dataclass
 from enum import Enum, auto
-from typing import Optional, TypedDict
+from typing import Literal, TypedDict
 
 from MDANSE import PLATFORM
 from MDANSE.Framework.Status import Status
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
This is going to be really useful. I have two observations at this stage:
1. It would be good to have one more field for job timing, with something like elapsed time.
2. On my machine, the progress bar keeps resizing the table column it is sitting in. For a fast job, the progress/progress rate text keeps moving and it becomes tiring trying to read it. Maybe there is a way to keep the width of that column fixed, or to stop the progress bar from changing dimensions?

oerc0122:
> This is going to be really useful. I have two observations at this stage:
> 
> 1. It would be good to have one more field for job timing, with something like elapsed time.
> 2. On my machine, the progress bar keeps resizing the table column it is sitting in. For a fast job, the progress/progress rate text keeps moving and it becomes tiring trying to read it. Maybe there is a way to keep the width of that column fixed, or to stop the progress bar from changing dimensions?

By (1) do you mean an extra column in the `QStandardItem` list or an extra entry in the text log?
(2) is in #891

MBartkowiakSTFC:
> By (1) do you mean an extra column in the `QStandardItem` list or an extra entry in the text log? (2) is in #891

I would like an extra entry in the text log showing how long the job has been running.

#891 is related to the problem I am describing, but there it was about the progress bar resetting to the same width all the time, and for me it is resetting do a _different_ width all the time. But you are probably right that fixing that one will fix this problem too.

MBartkowiakSTFC:
> 2. On my machine, the progress bar keeps resizing the table column it is sitting in. For a fast job, the progress/progress rate text keeps moving and it becomes tiring trying to read it. Maybe there is a way to keep the width of that column fixed, or to stop the progress bar from changing dimensions?

I can confirm that, from my point of view, changes introduced in #915 solved this problem.

MBartkowiakSTFC (approved):
I think this is essentially ready.

One optional change: if you think it makes sense, we could change the text in the job status from "Estimated completion" to "Estimated remaining time".

PR #906, opened 2025-07-29
Add optional CLI progress bars

oerc0122:
**Description of work**
- Add progress bars to CLI

**Fixes**
Fixes #605 

**To test**
1. Install mdanse with optional `cli`
   ```sh
   pip install "mdanse[cli]"
   ```
   (or `pip install tqdm`)
2. Run a manual job with `prog_bar=True`, e.g. provided test job.
 
   [tmp.txt](https://github.com/user-attachments/files/21491315/tmp.txt)

PR #910, opened 2025-07-30
Add comment about prog_bar availability

oerc0122:
**Description of work**
Add check to avoid confusing users who don't have prog_bars available.

**Fixes**
Update to #906 

**To test**
No functional change

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -293,6 +293,8 @@ parameters = {{
 
 if __name__ == "__main__":
     {cls.__name__.lower()} = IJob.create({cls.__name__!r})
+    # Progress bars only available if tqdm available.
+    # Install with `cli` optional dependency.
     {cls.__name__.lower()}.run(parameters, status=True, prog_bar=True)
 """)
 


MBartkowiakSTFC:
Sorry, but I will start a discussion here.

I am not sure if we really want to save a different script depending on what was installed on the machine running the GUI.

Some work went into making the following scenario possible:
1. A user launches the GUI on a (weak) laptop, configures a job and saves a script.
2. The script is copied onto a cluster/workstation and launched there in the shell.
I would prefer not to assume that the script will be run on the same machine as the one that it was saved on.

oerc0122:
> Sorry, but I will start a discussion here.
> 
> I am not sure if we really want to save a different script depending on what was installed on the machine running the GUI.
> 
> Some work went into making the following scenario possible:
> 
> 1. A user launches the GUI on a (weak) laptop, configures a job and saves a script.
> 2. The script is copied onto a cluster/workstation and launched there in the shell.
>    I would prefer not to assume that the script will be run on the same machine as the one that it was saved on.

That's reasonable, given it's ignored in the case the user doesn't have `tqdm`. Do you think it will confuse users to have `prog_bars` if they don't actually appear?

MBartkowiakSTFC:
> That's reasonable, given it's ignored in the case the user doesn't have `tqdm`. Do you think it will confuse users to have `prog_bars` if they don't actually appear?

It may confuse them, but may also help them notice that the CLI progress bar is now possible. It may be a valid assumption that the users who attempt to save scripts are more independent and will not be derailed by this option.

Another thought: should the output script have a _comment_ explaining that `prog_bar=True` will be ignored unless `tqdm` is installed?

oerc0122:
> Another thought: should the output script have a comment explaining that prog_bar=True will be ignored unless tqdm is installed?

That seems like a sensible solution. Done

MBartkowiakSTFC (approved):
I am happy with this solution.

PR #912, opened 2025-07-30
Remove iteration over directories; switch to explicit import/re-export

### LINKED ISSUE #747: [ENHANCEMENT] Use `pkgutils` for walking files
Currently in a lot of `__init__.py` files (c.f. https://github.com/ISISNeutronMuon/MDANSE/blob/protos/MDANSE/Src/MDANSE/Framework/Jobs/__init__.py) there are manual walks though the contents of the package.

This incurs both maintenance costs and captures invalid files such as temporary files created by linters. 

Manually importing all the required registered subclasses within each init. This would be the usual recommended way as this explicitly names the files to be imported and allows static checkers to validate their use. 

If this is not desirable, [`pkgutil`](https://docs.python.org/3/library/pkgutil.html) provides both the `walk_modules` and `iter_modules` functions in stdlib for doing this kind of job. 

N/A

**Additional context**
The `SubclassFactory` metaclass requires the imports in order to create the classes which register themselves as children.

oerc0122:
**Description of work**
As in title

**Fixes**
Prevents trying to load temporary files and allows exclusion of files if unused.
Fixes #747 

**To test**
Code should run normally.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/__init__.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/__init__.py
@@ -13,21 +13,131 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
-import importlib
-from pathlib import Path
 
-current_path = Path(__file__).parent
-
-modnames = (
-    fname.stem for fname in current_path.glob("*.py") if fname.stem != "__init__"
+from .ASEFileConfigurator import ASEFileConfigurator as ASEFileConfigurator
+from .AseInputFileConfigurator import (
+    AseInputFileConfigurator as AseInputFileConfigurator,
 )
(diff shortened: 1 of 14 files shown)


MBartkowiakSTFC:
It works fine, as far as I can tell.

One question: why do we need to use the pattern
`from .SOMETHING import NAME as NAME`,
instead of just
`from .SOMETHING import NAME`?

oerc0122:
> It works fine, as far as I can tell.
> 
> One question: why do we need to use the pattern `from .SOMETHING import NAME as NAME`, instead of just `from .SOMETHING import NAME`?

This is a semi-standard indication that this is being imported for re-export, not for use within the module. The alternative is listing all of them in `__all__ = []` which says what should be imported on a `from X import *`

MBartkowiakSTFC (approved):
Thanks for the explanation, looks good.

PR #915, opened 2025-07-31
Stop column width in RunTable from resetting

### LINKED ISSUE #891: [BUG] Adjusting the job cell widths while a job is running

MBartkowiakSTFC:
**Description of work**
The column width in the table of running jobs is still adjusted automatically when the contents change, but the progress bar column is excluded from the changes.

closes #891

**Fixes**
1. Added a custom method `selective_resize` to `RunTable`.

**To test**
Run a job in the GUI. It should be possible to change the width of the progress bar column, and it will not reset.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/RunTable.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/RunTable.py
@@ -15,8 +15,8 @@
 #
 from typing import Union
 
-from qtpy.QtCore import QAbstractItemModel, QModelIndex, Qt, Signal, Slot
-from qtpy.QtGui import QContextMenuEvent, QStandardItem
+from qtpy.QtCore import QModelIndex, Qt, Signal, Slot
+from qtpy.QtGui import QContextMenuEvent, QStandardItem, QStandardItemModel
 from qtpy.QtWidgets import QAbstractItemView, QMenu, QMessageBox, QTableView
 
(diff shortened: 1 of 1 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/RunTable.py
```diff
@@ -36,15 +38,21 @@ def __init__(self, *args, **kwargs):
...
-    def setModel(self, model: QAbstractItemModel) -> None:
+    def setModel(self, model: QStandardItemModel) -> None:
         result = super().setModel(model)
-        self.model().dataChanged.connect(self.resizeColumnsToContents)
+        model.itemChanged.connect(self.selective_resize)
         return result
 
+    @Slot("QStandardItem*")
+    def selective_resize(self, item: QStandardItem):
+        if colind := item.column() == PROGBAR_COLUMN:
```
oerc0122: Are you sure this takes what you want? `colind` here is the boolean value.

MBartkowiakSTFC: Thank you, you are right. I corrected it now.

oerc0122 (approved):
If this solves the problem I'm happy with it.

ChiCheng45 (approved):
Looks good.

PR #916, opened 2025-07-31
Fixed grouped results axis and updated compare hdf

### LINKED ISSUE #914: [BUG] Total results for molecule does not plot with the correct axis
**Description of the error**
 Total results for molecule does not plot with the correct axis. Here I plot all possible results using the four_molecules.mdt trajectory for the DOS calculation.

Image

The total results for the C60 and cubane molecules are not plotting against the correct axis.

ChiCheng45:
**Description of work**
Fixed grouped results so that they are plotted against the correct axis.

**Fixes**
Closes #914

**To test**
Run a calculation with molecule grouping and plot all results. Check that they are plotted against the correct axes. Check for 1D and 2D analysis results.


oerc0122 (approved):
Looks reasonable.

MBartkowiakSTFC (approved):
I think this works fine now.

PR #919, opened 2025-08-04
Fix unit parsing in ASE

### LINKED ISSUE #913: [BUG] Wrong velocities with ASE converter and janus extxyz
**Description of the error**
The ASE converter does not convert the velocity units correctly when reading .extxyz trajectories output by janus.

This problem can be noticed when calculating the temperature of the system.

In the example shown below, the temperature calculated from velocity values and from atom positions should agree within 1% tolerance.

The temperature from the atom positions is correct, but the values calculated from MD engine velocities are on a completely different scale.
Image

**Suggested fix**
The header line of each frame in the .extxyz file contains a field listing the units used in the file. ASE converter should parse that entry (if it is present) and base the unit conversion on the units specified there.

**Additional details**
The janus output files contain momenta of atoms. These seem to be given in units of (eV*u)^1/2.

oerc0122:
**Description of work**
Fix issues with parsing units from extxyz.

**Fixes**
Fixes #913 

**To test**
Standard tests. (not added yet)


MBartkowiakSTFC:
This already works well for the files I created within the last two weeks.

A year ago I was running janus simulations of some solid systems. I see that those files do not have unit information in the header, but apparently they must have already been using (eV*u)^1/2 for momenta, since the temperature from velocities seems to be off by the same factor.
ASE_temperature_problem

Since we can't always guess what MD engine the user was running and what units it was writing, what do you think we should do? If we wanted to do it properly, the user should have:
1. a chance to see what unit we have guessed the input to be in,
2. a chance to correct the unit if it is not right.
To give the user an obvious warning about the units being wrong, we could even pre-calculate the temperature from the velocities of the first frame.

In any case, many things can be done in the future. For now, should we guess that momenta are given in (eV*u)^1/2?

MBartkowiakSTFC (approved):
Looks good. For the new janus output files with the units specification, the results are correct.

PR #920, opened 2025-08-05
Show clearer error output from the plotter

### LINKED ISSUE #664: [BUG] Setting "Use it?" on another plot tab causes the other plots to reset
**Description of the error**
Setting "Use it?" on another plot tab causes the other plots to reset. This means that on one plot if I zoom in and then go to another plot and change the "Use it?" setting I lose my zoom-in on the first plot.

The zoom-in should be maintained on the plots when I change the setting on the other plots.

**Additional details**

Image

MBartkowiakSTFC:
**Description of work**
As part of the work on improving the GUI, this PR changes the behaviour of the plotter. Updating the plots should take less time, and the plot will be replaced with an error message if plotting fails.

closes #664 

**Fixes**
1. Only refresh the current plot when changing the settings of a specific plot.
2. If re-plotting fails, show an error message in the plotter instead of keeping the old figure.
3. Update the unit tests of the PlottingContext.

**To test**
Create several plots with 1D and 2D data sets. Input an index outside of the valid range into the "Use it?" text box to see if the error message is displayed in the plot tab.


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -470,15 +470,25 @@ def curves_vs_axis(
...
+                self._curves[index_tuple] = self.data[index_slicer].squeeze()
+            except IndexError:
+                LOG.warning(
+                    "Skipping the plot of dataset %s at index %s", self._name, index
+                )
+            else:
+                self._curve_labels[index_tuple] = self.generate_curve_label(
+                    index_tuple,
+                    label_lookup,
+                )
```
oerc0122: Bear in mind that `slice`s are allowed to go out of range
```
>>> x = []
>>> x[4:]
[]
```
So this will not always catch invalid bounds. Not something to be fixed as this is expected behaviour in Python, but we may want a comment.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -564,15 +574,16 @@ def planes_vs_axis(
 class PlottingContext(QStandardItemModel):
     """Data model storing data and user input used for plotting."""
 
-    needs_an_update = Signal()
+    needs_an_update = Signal(int)
 
-    def __init__(self, *args, unit_lookup=None, **kwargs):
+    def __init__(self, *args, unit_lookup=None, unique_number: int = -1, **kwargs):
```
oerc0122: What is `unique_number` doing?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Plotter.py
```diff
@@ -278,3 +278,16 @@ def plot(
         axes = target.add_subplot(111)
         self._axes = [axes]
         self.apply_settings(plotting_context)
+
+    def plot_blank(self, draw_cross: bool = True):
```
oerc0122:
```suggestion
    def plot_blank(self, *, draw_cross: bool = True):
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Plotter.py
```diff
@@ -278,3 +278,16 @@ def plot(
...
+        """Inform the user that no data could be plotted."""
+        figure = self.get_figure()
+        axes = figure.add_subplot(111)
+        if draw_cross:
+            axes.plot([0, 1], [0, 1], "k-")
+            axes.plot([0, 1], [1, 0], "k-")
```
oerc0122:
```suggestion
            axes.axline([0, 1], [0, 1], color="k", linestyle="-")
            axes.axline([0, 1], [1, 0], color="k", linestyle="-")
```
For infinitely long lines. More obvious they're not data?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -323,7 +327,7 @@ def add_ord(val: Iterable[int]) -> None:
                 )
                 continue
 
-        self._data_limits = list(complete_subset.keys()) if complete_subset else None
+        self._data_limits = list(complete_subset.keys()) if complete_subset else []
```
```suggestion
        self._data_limits = list(complete_subset)
```
is all you need in this case.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
```diff
@@ -144,6 +142,13 @@ def plot(
             except KeyError:
                 _, best_axis = dataset.longest_axis()
             for key, curve in islice(dataset._curves.items(), self._plot_limit):
+                counter += 1
+                if counter > self._plot_limit:
+                    LOG.warning(
+                        "Curves above the current limit of %s will be ignored",
+                        self._plot_limit,
+                    )
+                    break
```
oerc0122, line 151: These blocks will never be hit because of the `islice`. Also `counter` should be `enumerate` if it's needed.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
```diff
@@ -268,6 +268,10 @@ def plot(
                 limits = limits[2:] + limits[:2]
             for xnum in range(len(all_datasets)):
                 if startnum > self._plot_limit:
+                    LOG.warning(
+                        "Datasets above the current limit of %s will be ignored",
+                        self._plot_limit,
+                    )
                     break
                 axes = target.add_subplot(gridsize, gridsize, startnum)
                 startnum += 1
```
oerc0122, line 277: In this case is `startnum` just `xnum`?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Plotter.py
```diff
@@ -278,3 +278,16 @@ def plot(
         axes = target.add_subplot(111)
         self._axes = [axes]
         self.apply_settings(plotting_context)
+
+    def plot_blank(self, *, draw_cross: bool = True):
+        """Inform the user that no data could be plotted."""
```
oerc0122: Add explanation of what `draw_cross` does

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Single.py
```diff
@@ -227,6 +227,9 @@ def plot(
                         return
         if len(self._backup_curves) > 1:
             self.enable_slider(allow_slider=True)
+        elif len(self._backup_curves) == 0:
+            self.plot_blank()
+            return
```
oerc0122:
```suggestion
        elif not self._backup_curves:
            self.plot_blank()
            return
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/PlotHolder.py
```diff
@@ -124,17 +128,35 @@ def plotter(self):
...
+            Object id of the PlottingContext in which the settings got changed.
+        """
+        for plotter in self._plotter:
+            if plotter.unique_id == plot_number:
+                try:
+                    plotter.plot_data(update_only=True)
+                except Exception:
+                    LOG.error("Plotting failed: %s", traceback.format_exc())
+                    plotter.plot_blank()
+                    return
```
oerc0122: I still think `self._plotter` may want to be a `dict[id, plotter]` so you can always iterate over `values()` where you need all of them.
```suggestion
        plotter = more_itertools.first_true(self._plotter, pred=lambda plot: plot.unique_id == plot_number)
        if plotter is None:
            return
        try:
            plotter.plot_data(update_only=True)
        except Exception:
            LOG.error("Plotting failed: %s", traceback.format_exc())
            plotter.plot_blank()
```
Is this clearer? It's certainly flatter.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
MBartkowiakSTFC: `self._plot_limit` is meant to be the upper limit on the **total** number of subplots, also when using multiple datasets. `islice` is only dealing with one dataset at a time, and so would `enumerate`. If you are plotting more than one dataset in the same plot, this block gets called and prevents more subplots from being created.

oerc0122: Do we want to `chain` them, then `islice` then?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
MBartkowiakSTFC: This is less meaningful now since I have set `self._plot_limit = 1` for `Heatmap`, but `startnum` normally keeps track of all the subplots, while `xnum` is connected just to a single dataset, similar to my previous comment dealing with the `Grid` plotter. It might still be meaningful to plot several heatmaps next to each other, so I preferred to keep this option open, if we decided to increase the limit again. (The main problem with multiple plots was the interaction with the sliders changing the colormap.)

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/PlotHolder.py
MBartkowiakSTFC: Changing `self._plotter` to a dictionary would require us to change the way we associate individual plot tabs with the `Plotter` instances. `self._plotter` being a `list` has its advantages when we start deleting plot tabs. `QTabWidget` does not assign permanent keys or id numbers to individual plot tabs, so deleting a tab in the middle re-numbers all the following ones. This is consistent with the indexing of `list` elements.

I am sure that the changes you suggested are possible, but I would prefer this PR to focus just on making sure that plotting works.

oerc0122 (approved):

PR #923, opened 2025-08-05
Replace custom encoders with a single centralised one

oerc0122:
**Description of work**
Centralise JSON encoding into MDANSE encoder in IOUtils

**Fixes**
N/A

**To test**
Standard code should run and JSON dumps should work correctly.


MBartkowiakSTFC (approved):
Looks good.

PR #924, opened 2025-08-07
Fix trajectory editor with atom selection and trajectories with molecules.

### LINKED ISSUE #667: [BUG] Atom selection with trajectory editor and DFTB trajectory
**Description of the error**
Trajectory editing job fails when I use the DFTB trajectory and atom selection. The job fails when I use the settings ` {"all": false, "hs_on_heteroatom": true}`, `{"all": false, "hs_on_element": ["O"]}`, `{"all": false, "index": [276, 277, 278]}` and others. This even fails when whole molecules are selected.

**Additional details**
I get the following error in the logs.

```
self.rdkit_mol.AddBond(
RuntimeError: Range Error
atomIdx2
Violation occurred on line 497 in file Code\GraphMol\RWMol.cpp
Failed Expression: 200 RDKIT: 2024.09.3
BOOST:
```

ChiCheng45:
**Description of work**
Fix trajectory editor with atom selection and trajectories with molecules. Molecule groups may change depending on the selection. 

**Fixes**
Fixes #667

**To test**
Run the trajectory editor with atom selection and a trajectory with molecules.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -21,6 +21,7 @@ from functools import reduce
 from typing import Any, SupportsInt
 
 import h5py
+from more_itertools import padded
 import networkx as nx
 import numpy as np
 from rdkit import Chem
@@ -316,7 +317,7 @@ class ChemicalSystem:
             # we will pad them with -1, we will ignore these values
             # when the trajectory get loaded up see self.load
             size = max(len(val) for val in vals)
-            new_vals = [val + [-1] * (size - len(val)) for val in vals]
+            new_vals = [list(padded(val, fillvalue=-1, n=size)) for val in vals]
             clusters_group.create_dataset(key, data=new_vals)
 
     def load(self, trajectory: h5py.File | str):
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -311,8 +311,13 @@ class ChemicalSystem:
         for key, value in self._labels.items():
             label_group.create_dataset(key, data=value)
         clusters_group = grp.create_group("clusters")
-        for key, value in self._clusters.items():
-            clusters_group.create_dataset(key, data=value)
+        for key, vals in self._clusters.items():
+            # unable to store array with inhomogeneous row lengths
+            # we will pad them with -1, we will ignore these values
+            # when the trajectory get loaded up see self.load
+            size = max(len(val) for val in vals)
+            new_vals = [val + [-1] * (size - len(val)) for val in vals]
+            clusters_group.create_dataset(key, data=new_vals)
 
     def load(self, trajectory: h5py.File | str):
         """Read the ChemicalSystem information from the trajectory.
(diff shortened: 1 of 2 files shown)


MBartkowiakSTFC (commented):
This PR fixed the problem.

I am still wondering if the renaming of molecules based on their composition is consistent with our decision NOT to rename molecules on atom transmutation. What do you think?

ChiCheng45:
> This PR fixed the problem.
> 
> I am still wondering if the renaming of molecules based on their composition is consistent with our decision NOT to rename molecules on atom transmutation. What do you think?

Yea I would agree that it shouldn't rename. I went with this way because there were some issues with how the cluster indexes are stored. Basically we can't store something like

[[1, 2, 3]
[4, 5]]

which will happen if the whole atom was not selected. Anyway I had a think and perhaps we should use some dummy indexes to allow this so for the above we can used -1 or something.

[[1, 2, 3],
[4, 5, -1]]

MBartkowiakSTFC:
Is the problem basically with saving the cluster definitions to the HDF5 file? You are right that using a placeholder value for removed indices would solve it. -1 is OK, I think it is possible to save a NaN as well.

yes basically you can't store these types of arrays

```
    data = np.asarray(data, order="C", dtype=as_dtype)
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (4,) + inhomogeneous part.
```

I'll try to update it with -1 or nan.

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -311,8 +311,13 @@ def serialize(self, h5_file: h5py.File) -> None:
...
             label_group.create_dataset(key, data=value)
         clusters_group = grp.create_group("clusters")
-        for key, value in self._clusters.items():
-            clusters_group.create_dataset(key, data=value)
+        for key, vals in self._clusters.items():
+            # unable to store array with inhomogeneous row lengths
+            # we will pad them with -1, we will ignore these values
+            # when the trajectory get loaded up see self.load
+            size = max(len(val) for val in vals)
+            new_vals = [val + [-1] * (size - len(val)) for val in vals]
```
oerc0122: Are you padding to fixed size? 

https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.padded

MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryEditor.py
```diff
@@ -145,11 +146,25 @@ def initialize(self):
...
-                )
+            selected_idxs = set(self._indices)
+            indx_map = {j: i for i, j in enumerate(self._indices)}
+
+            selected_bonds = []
+            for bond_idxs in self._input_chemical_system._bonds:
+                if set(bond_idxs).issubset(selected_idxs):
+                    i, j = bond_idxs
+                    selected_bonds.append([indx_map[i], indx_map[j]])
+            new_chemical_system.add_bonds(selected_bonds)
```
oerc0122:
```suggestion
            selected_bonds = [
                (indx_map[i], indx_map[j])
                for i, j in self._input_chemical_system._bonds
                if i in selected_idxs and j in selected_idxs
            ]
            new_chemical_system.add_bonds(selected_bonds)
```

MBartkowiakSTFC (approved):
I tried creating some pretty pathological trajectories (different number of atoms per molecule, molecules with a single atom) using the TrajectoryEditor, and it seemed to work fine. They could still be read and analysed.

PR #927, opened 2025-08-08
Updated MDANSE CLI

### LINKED ISSUE #905: Check and clean up mdanse.py
`MDANSE/Scripts/mdanse.py` used to be the main part of the CLI in the 1.5.x version of MDANSE. While the file still exists, it has not been used or tested in a long time. It is expected that at least part of the functionality `mdanse.py` used to provide will not work. The code should be cleaned up, leaving only the part that is still relevant.

MBartkowiakSTFC:
**Description of work**
This is the initial idea for updating the contents of `mdanse.py`. Some of the original features of the script have become obsolete and need to be removed.

Ideally, this should provide a comparable level of pre-configuring the job scripts as is provided by the GUI (e.g. it should fill out the number of frames correctly if the trajectory file is provided, etc.)

closes #905

**Fixes**
1. Switched input parsing to `optparse`.
2. Removed options related to running jobs.
3. Added options for extracting basic information from .mdt and .mda files.

**To test**
Execute `mdanse` in the shell and see if you can find out what it can do based on the built-in documentation.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/Databases.py
+++ b/MDANSE/Src/MDANSE/Chemistry/Databases.py
@@ -587,18 +587,22 @@ class AtomsDatabase(_Database):
         """
         # A delimiter line.
         delimiter = "-" * 70
-        tab_fmt = " {:<20}{!s:>50}"
+        tab_fmt = "{:<20}{!s:>40}{!s:>10}"
 
         info = [
             delimiter,
             f"{atom:^70}",
-            tab_fmt.format("property", "value"),
+            tab_fmt.format("property", "value", "unit"),
             delimiter,
         ]
 
         # The values for all element's properties
         for pname in sorted(self._properties):
-            info.append(tab_fmt.format(pname, self._data[atom].get(pname, None)))
+            info.append(
+                tab_fmt.format(
+                    pname, self._data[atom].get(pname), self._units.get(pname)
+                )
+            )
 
         info.append(delimiter)
         info = "\n".join(info)
--- a/MDANSE/Src/MDANSE/Chemistry/Databases.py
+++ b/MDANSE/Src/MDANSE/Chemistry/Databases.py
@@ -84,6 +84,58 @@ def color(color_string: str | None = None):
     return color_string
 
 
+def atom_info(atom: str, database: AtomsDatabase | None = None) -> str:
+    """Return as string all the information about the input atom.
+
+    Parameters
+    ----------
+    atom : str
+        Atom type.
+
+    Returns
+    -------
+    str
+        Multi-line list of all the atom properties.
+
+    """
+    if database is None:
+        database = ATOMS_DATABASE
+    if isinstance(database, AtomsDatabase):
+        atoms = database.atoms
+        properties = database.properties
+        units = database.units
+    else:
+        atoms = database.atoms_in_database
+        properties = database.properties_in_database
+        units = defaultdict(lambda: "none")
+
+    if atom not in atoms:
+        raise KeyError(f"Atom {atom} is not in the atom database {database}.")
+    property_dict = {
+        prop_name: database.get_atom_property(atom, prop_name)
+        for prop_name in properties
+    }
+    # A delimiter line.
+    delimiter = "-" * 70
+    tab_fmt = "{:<20}{!s:>40}{!s:>10}"
+
+    info = [
+        delimiter,
+        f"{atom:^70}",
+        tab_fmt.format("property", "value", "unit"),
+        delimiter,
+    ]
+
+    # The values for all element's properties
+    for pname in sorted(properties):
+        info.append(tab_fmt.format(pname, property_dict.get(pname), units.get(pname)))
+
+    info.append(delimiter)
+    info = "\n".join(info)
+
+    return info
+
+
 class _Database(metaclass=Singleton):
     """Base class for all the databases."""
 
@@ -425,6 +477,11 @@ class AtomsDatabase(_Database):
     def properties(self) -> list[str]:
         """Return the names of the properties stored in the atoms database."""
         return sorted(self._properties.keys())
+
+    @property
+    def units(self) -> dict[str, str]:
+        """Return the dictionary mapping properties to their physical units."""
+        return self._units
 
     def get_property(self, pname: str) -> dict[str, str | int | float | list]:
         """Return the values of a property for all atoms.
--- a/MDANSE/Src/MDANSE/Chemistry/Databases.py
+++ b/MDANSE/Src/MDANSE/Chemistry/Databases.py
@@ -128,7 +128,10 @@ def atom_info(atom: str, database: AtomsDatabase | None = None) -> str:
 
     # The values for all element's properties
     for pname in sorted(properties):
-        info.append(tab_fmt.format(pname, property_dict.get(pname), units.get(pname)))
+        if pname.strip():
+            info.append(
+                tab_fmt.format(pname, property_dict.get(pname), units.get(pname))
+            )
 
     info.append(delimiter)
     info = "\n".join(info)
--- a/MDANSE/Src/MDANSE/Chemistry/Databases.py
+++ b/MDANSE/Src/MDANSE/Chemistry/Databases.py
@@ -20,7 +20,7 @@ import json
 from collections import defaultdict
 from collections.abc import ItemsView
 from pathlib import Path
-from typing import Any, Optional, SupportsComplex, Union
+from typing import Any, SupportsComplex
 
 from MDANSE.Core.Error import Error
 from MDANSE.Core.Platform import PLATFORM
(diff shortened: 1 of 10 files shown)


MDANSE/Src/MDANSE/Chemistry/Databases.py
```diff
@@ -84,6 +84,61 @@ def color(color_string: str | None = None):
...
+
+    Returns
+    -------
+    str
+        Multi-line list of all the atom properties.
+
+    """
+    if database is None:
+        database = ATOMS_DATABASE
+    if isinstance(database, AtomsDatabase):
```
oerc0122: What other types could `database` be? The annotation suggests it can only be this.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -45,6 +46,68 @@
...
+
+    val = "\n".join(val)
+
+    return val
+
+def chemical_system_summary(cs: ChemicalSystem) -> str:
+    text = "\n ==== Chemical System summary ==== \n"
+    atoms, counts = np.unique(cs.atom_list, return_counts=True)
+    for ind in range(len(atoms)):
+        text += f"Element: {atoms[ind]}; Count: {counts[ind]}\n"
```
```suggestion
    for atom, count in zip(atoms, counts):
        text += f"Element: {atom}; Count: {count}\n"
```

MDANSE/Tests/UnitTests/test_databases.py
```diff
@@ -202,18 +202,22 @@ def test_has_property(self):
...
-            " electronegativity                                                  2.2\n"
-            " family                                                       non metal\n"
-            " nucleon                                                              0\n"
-            " symbol                                                               H\n"
-            "----------------------------------------------------------------------",
-            ATOMS_DATABASE.info("H"),
-        )
+        info_text = ATOMS_DATABASE.info("H")
+        properties = []
+        for nline, line in enumerate(info_text.split('\n')):
```
oerc0122:
```suggestion
        for nline, line in enumerate(info_text.splitlines()):
```

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,435 @@
...
+
+def show_single_job(job_name: str):
+    if job_name in IJob.indirect_subclasses():
+        instance = IJob.create(job_name)
+    elif job_name in Converter.indirect_subclasses():
+        instance = Converter.create(job_name)
+    else:
+        raise KeyError(f"{job_name} is not a converter or analysis included in MDANSE.")
+    result = f"{job_name}\n"
+    result += "".join(len(job_name) * ["~"]) + "\n\n"
```
oerc0122:
```suggestion
    result += "~"*len(job_name) + "\n\n"
```

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,435 @@
...
+        help="List all the converter types.",
+    )
+    converter.add_argument(
+        "-n", "--name", help="Name of the specific converter to be used."
+    )
+    converter.add_argument(
+        "-o",
+        "--output",
+        help="Use this file name for the output Python script.",
+    )
```
oerc0122: Do these want to be mutex, likewise for the others?

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,435 @@
...
+        help="Create a script to analyse an MD trajectory.",
+        formatter_class=RawDescriptionHelpFormatter,
+        description=textwrap.dedent(
+            """
+        MDANSE can perform different analysis types on .mdt trajectories.
+        mdanse analysis commands let you view the available analysis types,
+        and create analysis scripts which you can run after adjusting the parameters.
+
+        Examples:
+        ---------
```
oerc0122, line 282: If this wants to look like numpydoc, we should remove the colon

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,435 @@
...
+        return
+    if args.name:
+        show_single_job(args.name)
+
+
+def execute_results(args: Namespace):
+    show_results_contents(args.file_name, verbose=args.verbose)
+
+
+def build_parsers() -> ArgumentParser:
```
oerc0122: Might want to break this down into several private functions each constructing one of the parsers to make it easier to maintain.

MDANSE/Tests/UnitTests/test_databases.py
```diff
@@ -202,18 +202,22 @@ def test_has_property(self):
...
+            if nline == 1:
+                self.assertEqual(line.strip(), "H")
+            elif nline ==2:
+                tokens = line.split()
+                for keyword in {"property", "value", "unit"}:
+                    self.assertIn(keyword, tokens)
```
oerc0122:
```suggestion
                self.assertTrue({"property", "value", "unit"}.issubset(tokens))
```

MDANSE/Tests/UnitTests/test_databases.py
```diff
@@ -202,18 +202,22 @@ def test_has_property(self):
...
+            elif nline > 3:
+                tokens = line.split()
+                if len(tokens) > 2:
+                    properties.append(tokens[0])
+        for keyword in self.properties:
+            self.assertIn(keyword, properties)
```
oerc0122: Possibly with lines broken up a bit better.
```suggestion
        info_text = ATOMS_DATABASE.info("H")
        lines = iter(info_text.splitlines())
        self.assertEqual(next(lines).strip(), "H")
        self.assertTrue({"property", "value", "unit"}.issubset(next(lines).split()))
        
        properties = {
            tokens[0] for line in lines if len(tokens := line.split()) > 2
        }
        self.assertTrue(set(self.properties) < properties)
```

MDANSE/Src/MDANSE/Chemistry/Databases.py
MBartkowiakSTFC: Ideally it should be possible to use either `AtomsDatabase` or `Trajectory`. Currently the `Trajectory` class provides part of the functionality of `AtomsDatabase`, but the naming of the methods is not consistent between the two.

MBartkowiakSTFC:
Thanks for the comments so far.

@oerc0122 @ChiCheng45 you are welcome to have a look at the commands and options so far and think if the current option tree makes sense.

I really wanted the `element` parser to take the atom name as an argument, not an option. But then it throws an error when I request a list (`mdanse element --list`) because the argument is missing. Is there an easy way to make it work? Would it make sense to have it work this way?

oerc0122:
> Thanks for the comments so far.
> 
> @oerc0122 @ChiCheng45 you are welcome to have a look at the commands and options so far and think if the current option tree makes sense.
> 
> I really wanted the `element` parser to take the atom name as an argument, not an option. But then it throws an error when I request a list (`mdanse element --list`) because the argument is missing. Is there an easy way to make it work? Would it make sense to have it work this way?

`nargs="?"` or `nargs=argparse.REMAINDER` if you want to be able to specify multiple

MBartkowiakSTFC:
I think that I implemented all the recommended changes to the code.

My long-term plan is to make `mdanse analysis SomeAnalysisName -t traj_file.mdt -o script.py` generate a script with the same starting values of the input parameters that we would get from the GUI. But this should be done in another PR, and for now this command will raise a `NotImplementedError`.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -50,6 +51,68 @@
...
+            try:
+                val.append(f"\t- {k}: {v['value'].shape}")
+            except KeyError:
+                continue
+
+    val.append("\nConversion history:")
+    metadata = check_metadata(traj.file)
+    if metadata:
+        for k, v in metadata.items():
+            val.append(f"{k}: {v}")
```
oerc0122, line 95: I feel like the number of times we do this we want an IOUtils function for this. 

Not for this PR though.

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,462 @@
...
+    instance = Trajectory(trajectory_name)
+    result = trajectory_summary(instance)
+    result += chemical_system_summary(instance.chemical_system)
+    traj_arrays = get_hdf5_contents(instance.file)
+    result += "====DATA ARRAYS====\n"
+    result += "\n".join(
+        f"{name}: type={instance.file[name].dtype}, shape={instance.file[name].shape}"
+        for name in traj_arrays
+    )
+    print(result)  # noqa: T201
```
oerc0122, line 79: Do we want to pass in an `out_file` which defaults to `sys.stdout` here and `write` the result in case people want to save it somewhere?

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,462 @@
...
+                "==Converters==",
+                *sorted(converters),
+            ]
+        )
+    else:
+        analyses = []
+        for job_name in IJob.indirect_subclasses():
+            instance = IJob.create(job_name)
+            if instance.category[0] != "Converters" and instance.enabled:
+                analyses.append(list(getattr(instance, "category", [])) + [job_name])
```
oerc0122:
```suggestion
                analyses.append([*getattr(instance, "category", []), job_name])
```

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,462 @@
...
-
-        @param parser: the MDANSE option parser.
-        @type parser: instance of MDANSEOptionParser
+        ),
+    )
+    converter.add_argument(
+        "-l",
+        "--list",
+        action="store_true",
+        default=False,
```
oerc0122:
```suggestion
```
This is implied by `store_true`

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,462 @@
...
-            LOG.info("Saved template for job %r as %r", name, filename)
-
-    def save_job_template(self, option, opt_str, value, parser):
+        ),
+    )
+    analysis.add_argument(
+        "-l",
+        "--list",
+        action="store_true",
+        default=False,
```
oerc0122:
```suggestion
```

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,436 +15,462 @@
...
-        "-r",
-        "--registry",
-        action="callback",
-        callback=parser.query_classes_registry,
-        help="Display the contents of MDANSE classes registry.",
+    results.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        default=False,
```
oerc0122:
```suggestion
```

oerc0122 (approved):
Looks good, just a couple of minor things.

MDANSE/Src/MDANSE/Scripts/mdanse.py
MBartkowiakSTFC: Sure, we can do it. Just to check with you: do you mean adding `out_file = sys.stdout` as a function keyword argument, and then adding an optional output file name to the parser, in case the user wanted to send the output to a file?

How much better is it going to be than just typing `mdanse some_command > saved_output.txt` ?

oerc0122: That's why I've left it as a question. I'm not sure it is what we want to do. 

I was referring to using a file-handle as the input coming from a cli arg with the default being `sys.stdout`.

MBartkowiakSTFC: OK, I think that I will not implement this change just yet.

oerc0122 (approved):

PR #929, opened 2025-08-12
Update compare_hdf5 and fix rmsf with molecule grouping

### LINKED ISSUE #926: [BUG] RootMeanSquareFluctuation with molecule grouping has wrong output length
**Description of the error**
For the DFTB water trajectory, (300 atoms, 100 molecules), the RMSF run produces an axis array with 100 indices, but the result array still has 300 elements.

The `rmsf` array and the index array should have the same size.

The `rmsf` array size matches the number of atoms, not the molecules.

**Suggested fix**
Further checks are needed to see if the results are correct (i.e. the first 100 values are the result, and the rest is not set).

**Additional details**
This is not picked up by unit tests.
Plotting fails because of the mismatched arrays.

ChiCheng45:
**Description of work**
Update compare_hdf5 and fix rmsf with molecule grouping. In some cases compare_hdf5 was not comparing results because the keys were not generated properly which resulted in a passing test.

**Fixes**
Fixes #926 
Fixed bug with ideal instrument, results included ideal results when ideal instrument is used so it was saved twice.

**To test**
Tests should pass. Check the RMSF with each molecule grouping works as expected.


MDANSE/Tests/CommonFunctions/test_helpers/compare_hdf5.py
```diff
@@ -41,10 +41,12 @@ def compare_hdf5(
 
     with h5py.File(result_path) as result, h5py.File(benchmark_path) as benchmark:
         if startswith:
-            keys = (key for key in result.keys() if key.startswith(comparison_keys))
+            keys = [key for key in result.keys() if any(key.startswith(comparison_key) for comparison_key in comparison_keys)]
```
oerc0122: `startswith` accepts a tuple and does that `any` check more efficiently.

Perhaps changing the typing from `Sequnce[str]` to `tuple[str, ...]` or simply casting to `tuple` might resolve the issue more clearly.

Perhaps 
```python
comparison_keys = tuple(more_itertools.always_iterable(comparison_keys))
```

oerc0122 (approved):

PR #930, opened 2025-08-13
Reduce some code duplication in selection and add load from dict

oerc0122:
**Description of work**
Minor change to avoid repeated code and add a "from dict" option.

**Fixes**
N/A

**To test**
Should be no changes.


ChiCheng45 (approved):
LGTM

PR #931, opened 2025-08-14
Fix converter scripts saved from GUI

### LINKED ISSUE #922: [BUG] Converter scripts saved from GUI don't work
**Description of the error**
All scripts created by MDANSE create job instances using `IJob.create(JOB_NAME)`. It seems that now IJob does not have converters registered in it.

It used to be possible to create converters as child classes of `IJob`.

After importing `IJob`, converters are not available as indirect children. Example:
```
In [3]: IJob.create('ASE')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 IJob.create('ASE')

File MDANSE/Core/SubclassFactory.py:174, in SubclassFactory.create(cls, name, *args, **kwargs)
    172     if len(closest) > 0:
    173         err_str += f" Did you mean: {closest[0]}?"
--> 174     raise ValueError(err_str)
    175 return specific_class(*args, **kwargs)

ValueError: Could not find ASE in IJob.
```

**Suggested fix**
We could have converter scripts use `Converter.create()`.

**Additional details**
N/A

ChiCheng45:
**Description of work**
Updated scripts so that they use Converter.create() for converter jobs.

**Fixes**
Fixes #922

**To test**
Using the GUI, generate a script that converts a trajectory and check that it runs correctly. Generate one for an analysis job and check that it runs correctly as well.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Converters/Converter.py
+++ b/MDANSE/Src/MDANSE/Framework/Converters/Converter.py
@@ -29,8 +29,10 @@ class Converter(IJob, metaclass=SubclassFactory):
     """Outputs a trajectory in the MDT format."""
 
     category = ("Converters",)
-
     ancestor = ["empty_data"]
+    runscript_import_line = (
+        "from MDANSE.Framework.Converters.Converter import Converter"
+    )
 
     @abstractmethod
     def run_step(self, index):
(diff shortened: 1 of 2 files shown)


oerc0122 (approved):
Looks reasonable to me! LGTM

PR #933, opened 2025-08-15
Make source PEP-8 naming compliant

oerc0122:
**Description of work**
- Run `sed -r -i ':loop;s/\b(_?[a-z][a-z_]*)([A-Z])/\1_\l\2/g;tloop' MDANSE/Src/**/*.py`
- Clean up resulting issues
- Manually use lsp replaces for GUI since PySide is not PEP8 compliant

**Fixes**
N/A

**To test**
Should be no functional changes.


MBartkowiakSTFC:
The only problem I found so far is that the `DistHistCutoffWidget` does not have valid starting values anymore. If you pick a trajectory with a unit cell and pick a job like `StaticStructureFactor`, the suggested r values will go from 0 to 0:
Screenshot 2025-08-27 at 08 59 22

The same problem does not occur in the main branch.

oerc0122:
> The only problem I found so far is that the `DistHistCutoffWidget` does not have valid starting values anymore. If you pick a trajectory with a unit cell and pick a job like `StaticStructureFactor`, the suggested r values will go from 0 to 0: [img] 
> The same problem does not occur in the main branch.

This was due to a `kwargs.get` getting `valueType`. Should be fixed.

MBartkowiakSTFC:
After some more testing I think that the GUI delegates have stopped working. They are subclasses of either `QItemDelegate` or `QStyledItemDelegate`, and their methods have to follow the `Qt` camelCase. The combo boxes in the plot details (e.g. plot axis selection), colour picker for the plotter and trajectory view, and input sanitisers for atom properties are all affected.
So far it seems that the files that need fixing are `MDANSE_GUI/Tabs/Views/Delegates.py` and `MDANSE_GUI/ElementsDatabaseEditor.py`.

MDANSE/Src/MDANSE/Framework/Configurators/MDAnalysisCoordinateFileConfigurator.py
```diff
@@ -15,7 +15,7 @@
 #
 from __future__ import annotations
 
-import MDAnalysis as mda
+import MDAnalysis as MDa
```
ChiCheng45: Is this a mistake? Why is this MDa with the M and D capitalised?

oerc0122: Because according to PEP-8, a capitalised import must remain capitalised on alias, however, going from Camel to upper-snake is not allowed. It's quite awkward and I couldn't think of a good alternative for an internal name. If you can think of something better, I'd appreciate it.

ChiCheng45: Fair enough, it doesn't matter really, but the all lowercase is what the MDAnalysis developers use. https://userguide.mdanalysis.org/stable/examples/quickstart.html.

oerc0122: Yeah, more than happy to revert it and add an exception

MDANSE/Src/MDANSE/Framework/Converters/MDAnalysis.py
```diff
@@ -229,7 +229,7 @@ def run_step(self, index: int):
 
             if hasattr(self.u.trajectory.ts, "forces"):
                 conf["gradients"] = self.u.trajectory.ts.forces * measure(
-                    1.0, "kJ/mol ang", equivalent=True
+                    1.0, "k_j/mol ang", equivalent=True
```
ChiCheng45: This change causes the MDAnalysis conversion to fail when forces exist.

```
MDANSE.Framework.Units.UnitError: The unit k_j is unknown
```

oerc0122: Ah, thought I'd caught all of those, good spot, thanks.

PR #934, opened 2025-08-27
Adjust job categories

### LINKED ISSUE #918: [ENHANCEMENT] Move trajectory jobs out of the analysis tree and converters up one level

- In the Actions tab, the trajectory jobs are under the analysis drop-down. They aren't analysis calculations, I think they should be moved out so they are not under the analysis drop-down. There should be an analysis and trajectory tree.

Image

- We could also update the converters jobs out one level since they are all in the same group.

- I also think that we should move averagestructure out of structure and into trajectory since it is not really an analysis calculation. It doesn't produce an MDA file.

MBartkowiakSTFC:
**Description of work**
Change the category tuples of analysis and converter jobs. This is meant to make the GUI trees easier to navigate.

closes #918 

**Fixes**
1. The "Converters" node in the converter tab is now hidden.
2. Converters are separated into "General" and "Specific".
3. The converter tree is fully expanded.
4. Analysis jobs which do not produce an .mda file are now in a "Trajectory" section, outside of "Analysis".

**To test**
Manual test of the GUI: Converter tab and analysis tab should be showing the new layout of available jobs.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Converters/ASE.py
+++ b/MDANSE/Src/MDANSE/Framework/Converters/ASE.py
@@ -54,6 +54,7 @@ class ASE(Converter):
     standard input file names.
     """
 
+    category = ("Converters", "General")
     label = "ASE"
 
     settings = collections.OrderedDict()
--- a/MDANSE/Src/MDANSE/Framework/Converters/Converter.py
+++ b/MDANSE/Src/MDANSE/Framework/Converters/Converter.py
@@ -28,7 +28,7 @@ from MDANSE.MLogging import LOG
 class Converter(IJob, metaclass=SubclassFactory):
     """Outputs a trajectory in the MDT format."""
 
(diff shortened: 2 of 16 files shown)


MBartkowiakSTFC:
I went slightly beyond the scope of #918 and separated the converters into "General" and "Specific". (This could have also been "External" and "MDANSE" but this is probably of less importance to the users.) Let me know if this was unnecessary.

oerc0122 (approved):
Seems reasonable to me, however, since not all of the items in the analysis trees are expanded, it strikes me as a little unintuitive that the `trajectory` options are expanded as thought to suggest these should be the first things to be aware of.

Happy to go in as-is, as this is an improvement if you disagree though.

MBartkowiakSTFC:
Thanks for the suggestion. I changed it so that the `Trajectory` runs are not expanded initially.

PR #937, opened 2025-08-28
Add warning to interpolation order input

### LINKED ISSUE #888: [ENHANCEMENT] Warn on interpolating positions when velocities are present
There is usually no advantage to calculating the velocities from atom positions when the MD engine velocities are provided in the trajectory. The GUI sets interpolation order to 0 (i.e. "use velocities, don't interpolate") if velocities are present. However, the user is still able to change the value. Also, a user could be re-using a script created for a different trajectory without noticing that the interpolation setting is not optimal.

Setting the interpolation order to a non-zero value for a trajectory with velocities included should produce a warning.
The warning should be handled by both the backend (for scripts) and the GUI.

We could also disable the interpolation altogether for trajectories with velocities.

**Additional context**
N/A

MBartkowiakSTFC:
**Description of work**
A warning is shown if the input trajectory contains velocities, but the interpolation order is set to a non-zero value.

closes #888 

**Fixes**
1. `InterpolationOrderConfigurator.warning_status` is set to a warning message if the input trajectory contains velocities and the interpolation order is not 0.
2. The initial text of the interpolation order input widget has been corrected to show "0th order - no interpolation" if the trajectory contains velocities.

**To test**
Load a trajectory with velocities into the GUI. Select an analysis type with interpolation order (e.g. density of states). Change interpolation order from 0 to other values and back.
The input widget should be highlighted on warning, and the warning text shown as tooltip and in the logger.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
@@ -55,6 +55,7 @@ class InterpolationOrderConfigurator(IntegerConfigurator):
         """
         if not self.update_needed(value):
             return
+        self.warning_status = ""
 
         frames_configurator = self.configurable[self.dependencies["frames"]]
         if not frames_configurator.valid:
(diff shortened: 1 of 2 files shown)


MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
```diff
@@ -67,28 +68,32 @@ def configure(self, value):
 
         IntegerConfigurator.configure(self, value)
 
-        if value == 0:
-            trajConfig = self.configurable[self.dependencies["trajectory"]]
+        trajConfig = self.configurable[self.dependencies["trajectory"]]
+        traj_has_velocities = "velocities" in trajConfig["instance"].variables()
```
oerc0122: Isn't there a `hasvariable` method?

oerc0122 (approved):

MDANSE/Src/MDANSE/Framework/Configurators/InterpolationOrderConfigurator.py
MBartkowiakSTFC: Good point. I switched this line to `Trajectory.has_variable()`.

PR #940, opened 2025-09-01
Fix single-index analysis

### LINKED ISSUE #939: [BUG] Analysis fails for selections with a single atom
**Description of the error**
`Trajectory.unique_elements` produces wrong results if a trajectory contains a single atom. It uses `operator.itemgetter` (https://docs.python.org/3/library/operator.html#operator.itemgetter) which in this case returns `tuple[str]` if more than one index has been selected, **but** it returns `str` if a single index has been selected. In that case it will find unique characters in the atom name, and not unique atom names.

It should have worked on a single unique atom names.

It tries to work on multiple atoms with single-character names.

**Suggested fix**
It would have been nice if `itemgetter` returned consistent types. Should we add checks for number of indices, or for the returned type?

**Additional details**
N/A

### LINKED ISSUE #943: [BUG] TrajectoryEditor fails for CentreOfMassesTrajectory output
**Description of the error**
A trajectory containing artificial atoms can be used as input for `TrajectoryEditor`, but the run fails when writing out the atom database. Here, I created a trajectory using `CenterOfMassesTrajectory`, and tried to reduce the number of frames.

Trajectories should be processed by the editor independent of what atoms they contain.

`TrajectoryEditor` fails after all the steps have been completed, with the error:
```
  File "Framework/Jobs/IJob.py", line 514, in run
    self.finalize()
  File "Framework/Jobs/TrajectoryEditor.py", line 254, in finalize
    self._output_trajectory.write_standard_atom_database()
  File "MolecularDynamics/Trajectory.py", line 917, in write_standard_atom_database
    property_dict = create_average_atom(atom_dict, database)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MolecularDynamics/Trajectory.py", line 636, in create_average_atom
    all_properties = database.properties_in_database
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'AtomsDatabase' object has no attribute 'properties_in_database'
```

**Suggested fix**
It should be straightforward to copy the atom properties from the input trajectory.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
In `Trajectory` and several analysis types we use `itemgetter` to pick items from a sequence of strings. This becomes a problem as `itemgetter` normally returns a `tuple[str]`, but in the case of a single index having been selected it returns a `str`, which is then typically split into single characters. This PR tries to eliminate this problem to allow the users run as many analysis types as possible, also with just a single atom selected.

closes #939
closes #943 

**Fixes**
1. The output of `itemgetter` is passed through `more_itertools.always_iterable` throughout the code.
2. `Trajectory.property` and `Trajectory.atoms` are used as property names to match `AtomsDatabase` property names.
3. `Configurable` mechanism of replacing trajectory input with in-memory trajectory objects has been changed to modify the class instance instead of the class definition.
4. Unit tests have been added to check that an analysis can run on a trajectory with a single atom. So far the tests do not use reference values, but it can be added later.

**To test**
1. Create a trajectory with a single atom, preferably with a long name. `CentreOfMassesTrajectory` can be used to convert molecules into atoms with long names.
2. Check that MeanSquareDisplacement, CoordinationNumber, Eccentricity etc. can be run on this trajectory.
3. Run `CentreOfMassesTrajectory` on a trajectory with molecules, and then run `TrajectoryEditor` on the resulting trajectory.


MDANSE/Src/MDANSE/Framework/Configurable.py
```diff
@@ -59,38 +59,35 @@ def __init__(self, settings=None, trajectory_input="mdanse"):
...
+            return ("MockTrajectoryConfigurator", {})
+        return None
 
     def build_configuration(self):
         from MDANSE.Framework.Configurators.IConfigurator import IConfigurator
 
         self._configuration.clear()
 
-        for name, (typ, kwds) in list(self.settings.items()):
+        for name, (orig_class_name, orig_kwds) in list(self.settings.items()):
```
oerc0122:
```suggestion
        for name, (orig_class_name, orig_kwds) in self.settings.items():
```

MDANSE/Src/MDANSE/Framework/Jobs/CroppedTrajectory.py
```diff
@@ -62,7 +64,7 @@ def initialize(self):
         atoms = self.trajectory.atom_types
         indices = self.trajectory.atom_indices
 
-        self._selectedAtoms = self.trajectory.selection_getter(atoms)
+        self._selectedAtoms = always_iterable(self.trajectory.selection_getter(atoms))
```
oerc0122, line 67: Instead of sticking this change everywhere, it would be possible in `Trajectory` to use [`make_decorator`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.make_decorator) with `always_iterable` to do this.

```python
## At the top of Trajectory
from more_itertools import make_decorator, always_iterable

alit = make_decorator(always_iterable)()
```

```python
## Trajectory l 254
        self.selection_getter = alit(itemgetter(*selected_indices))
```

```
>>> from more_itertools import always_iterable, make_decorator
>>> from operator import itemgetter
>>> x = [1, 2, 3, 4, 5, 6]
>>> alit = make_decorator(always_iterable, result_index=0)()
>>> q = alit(itemgetter(1))
>>> for i in q(x): print(i)
... 
2
>>> q = alit(itemgetter(1, 2, 3, 4))
>>> for i in q(x): print(i)
... 
2
3
4
5
```

oerc0122 (changes requested):
Most changes look good, but we can reduce the code noise with the `always_iterable` significantly

MDANSE/Src/MDANSE/Framework/Jobs/CroppedTrajectory.py
oerc0122: This does require we never use `selection_getter` in a non-iterable context (i.e. as a tuple/int)

MBartkowiakSTFC: I tried implementing this, but quite a few unit tests started failing with
`AttributeError: Can't pickle local object 'make_decorator.<locals>.decorator.<locals>.outer_wrapper.<locals>.inner_wrapper'`
so maybe our parallelisation scheme requires a different approach.

oerc0122: Alright, in that case, let's use it as-is.

oerc0122 (approved):

PR #945, opened 2025-09-02
Update AtomSelectionWidget on selection change

### LINKED ISSUE #887: [BUG] Jobs can be run when no atoms are selected.
**Description of the error**
Jobs can be run when no atoms are selected. We should block jobs from being run when no atoms are selected. 

Interestingly, the GUI does block the job being run after I have already run it. 

Image

MBartkowiakSTFC:
**Description of work**
The minimal change preventing the user from running an analysis on an empty selection.

closes #887

**Fixes**
`SelectionModel.selection_changed` signal has been connected to `AtomSelectionWidget.updateValue`.

**To test**
Load a trajectory, pick an analysis with atom selection, and invert selection. The GUI should prevent you from running the analysis.


oerc0122 (approved):
Does what it says on the tin

PR #949, opened 2025-09-04
Migrate to Python 3.10 officially

oerc0122:
**Description of work**
Python 3.9 will be deprecated come October. This is the minimal set of changes to appease `ruff` and deprecate 3.9.

**Fixes**
Replace `isinstance` tuples with `|`
Move `Callable` to `collections.abc`

**To test**
Standard tests


MBartkowiakSTFC (approved):
I think this should be fine. I removed the requirement for Python 3.9 tests to pass.

PR #951, opened 2025-09-04
Small subset of more-ruff-lint

oerc0122:
**Description of work**
Break the Ruff lint PR up into smaller components

**Fixes**
Enable ruff simplify

**To test**
Standard tests


MBartkowiakSTFC (approved):
I think these changes should be safe. I did not find any problems.

PR #957, opened 2025-09-06
Force NDTSF scaling factor to be real

### LINKED ISSUE #956: [BUG] NDTSF fails for some chemical elements
**Description of the error**
For some input trajectories, the `NeutronDynamicTotalStructureFactor` analysis will fail because the scaling factor will contain an imaginary component, while the output data type is float.

It seems that in our unit tests all the scaling factors are real. However, for some more complicated trajectories (like https://www.ccpbiosim.ac.uk/slim-md/slimmd-database/protein-membrane/beta2-receptor-in-the-empty-form-40-ns) the NDTSF analysis fails with this error:
```
 File "MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py", line 394, in finalize
    self._outputData["ndsf/f(q,t)_coh/total"][:] += (
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('complex128') to dtype('float64') with casting rule 'same_kind'
```

I expect that for now we will just have to use the real part of the scaling factors.

MBartkowiakSTFC:
**Description of work**
Explicitly use only the `real` part of the scaling factors in NeutronDynamicTotalStructureFactor analysis.

closes #956 

**Fixes**
1. Added `.real` to all the neutron-related scaling factors in the NDTSF `finalize()` method.

**To test**
Unit tests should pass.
Try any trajectory with C, H, N, O, P, S, Na, Cl elements. The NDTSF calculation should fail on the main branch and finish without error on this branch.
This https://www.ccpbiosim.ac.uk/slim-md/slimmd-database/protein-membrane/beta2-receptor-in-the-empty-form-40-ns or similar trajectories should be useful for triggering the error. (For the record, I removed water from the trajectory first using TrajectoryEditor, and only then ran the DCSF+DISF calculations.)

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
@@ -410,10 +410,10 @@ class NeutronDynamicTotalStructureFactor(IJob):
             bi = self.trajectory.get_atom_property(ele_i, "b_incoherent")
             self._outputData[f"ndsf/f(q,t)_inc/{label}"].scaling_factor *= (
                 bi**2 * number * norm_natoms
-            )
+            ).real
             self._outputData[f"ndsf/s(q,f)_inc/{label}"].scaling_factor *= (
                 bi**2 * number * norm_natoms
-            )
+            ).real
             self._outputData["ndsf/f(q,t)_inc/total"][:] += (
                 self._outputData[f"ndsf/f(q,t)_inc/{label}"][:]
                 * self._outputData[f"ndsf/f(q,t)_inc/{label}"].scaling_factor
--- a/MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/NeutronDynamicTotalStructureFactor.py
@@ -385,10 +385,10 @@ class NeutronDynamicTotalStructureFactor(IJob):
             pre_fac = 1 if label_i == label_j else 2
             self._outputData[f"ndsf/f(q,t)_coh/{pair_str}"].scaling_factor *= (
                 pre_fac * bi * bj * sqrt_cij
-            )
+            ).real
             self._outputData[f"ndsf/s(q,f)_coh/{pair_str}"].scaling_factor *= (
                 pre_fac * bi * bj * sqrt_cij
-            )
+            ).real
 
             self._outputData["ndsf/f(q,t)_coh/total"][:] += (
                 self._outputData[f"ndsf/f(q,t)_coh/{pair_str}"][:]


ChiCheng45:
Looks like its fixed now.

Can we update the documentation of what is supposed to happen with the complex weights? We can do this in another PR. I've lost track of what is supposed to happen. I thought that the imaginary part of the weights were meant to have been dropped when it was taken from the database so that the error in #956 should not have been possible. 

Note that

```
            self._outputData[f"ndsf/f(q,t)_coh/{pair_str}"].scaling_factor *= (
                pre_fac * bi * bj * sqrt_cij
            ).real
```

can be quite different to 

```
            self._outputData[f"ndsf/f(q,t)_coh/{pair_str}"].scaling_factor *= (
                pre_fac * bi.real * bj.real * sqrt_cij
            )
```
the second one was the one that I thought we were doing.

ChiCheng45:
Here I have made custom atoms types with massive complex components.

Screenshot 2025-09-09 at 16 59 50

I have then used these atoms types in the DFTB trajectory and calculated the NDTSF. Now if I compare DCSF from the DCSF calculation and the NDTSF calculation they are quite different.


Here is the OO DCSF from the DCSF and NDTSF calculations

Screenshot 2025-09-09 at 17 03 23
Screenshot 2025-09-09 at 17 03 32

Here it is again but with the scaling factors removed

Screenshot 2025-09-09 at 17 04 06

Screenshot 2025-09-09 at 17 04 15

ChiCheng45:
Sorry hit the wrong button wasn't meant to close it.

MBartkowiakSTFC:
There is definitely some room for improvement, and we could make our workflow clearer.

Normally, this happens:
1. `Trajectory.get_weights` is called just to get the value of the selected property for each atom type.
2. `get_weights` from `MDANSE.Mathematics.Arithmetic` is called to create the weight factors for the datasets. This immediately calls `adjust_weights`.
3. `adjust_weights` does `np.prod` on the properties, and _then_ it converts the product to `float64`. This is the same as doing `(b_i*b_j*...).real`.
4. `assign_weigths` is used to multiply the dataset scaling factors by the weights.

So, it looks like the changes in this PR are consistent with what we have been doing.

The problem is that `(b_i.real x b_j.real)` is not the same as `(b_i x b_j).real`, but also neither of them is the `b_i x b_j*` which the theory tells us to use. But since you mentioned this, `b_i.real x b_j.real` seems to be closer to the `b_i x b_j*` than the other one was.

Should we think again about using the complex conjugate? It would be good to sort this out before the release.

ChiCheng45:
Maybe we should just fix it so that complex scattering length is applied correctly.

```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)+ \beta_{B}^{*}\beta_{A}C_{AB}^{*}(q,t) \right] =  \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left[\beta_{A}^{*}\beta_{B} C_{AB}(q,t) \right]
```

I mentioned a lot about this in #740. There is also

```math
\tilde{F}_{AB}(q,t) = \frac{1}{2}(F_{AB}(q,t) + F_{AB}(-q,t)) = \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} \right\}  \mathrm{Re} \left\{ C_{AB}(q,t) \right\}
```

so a calculation with $q$ will actually be an average of $q$ and $-q$.

MBartkowiakSTFC:
OK, I'm fine to make it one of our goals to finally start using the complex scattering lengths correctly.

In the meantime, what about this PR? The difference between the NDTSF and DCSF that I get for made up atoms with large imaginary components is just a scaling factor:
dcsf
ndtsf

ChiCheng45:
> OK, I'm fine to make it one of our goals to finally start using the complex scattering lengths correctly.
> 
> In the meantime, what about this PR? The difference between the NDTSF and DCSF that I get for made up atoms with large imaginary components is just a scaling factor: dcsf ndtsf

How large? Because I did the same and got large differences see my post from earlier.

MBartkowiakSTFC:
Sorry, I will have to redo it tomorrow. It looks like my atom mapping in the converter, where I replaced normal atoms with the new ones, got completely ignored, and I still had normal elements in the output trajectory. I will have to put in an issue for that one as well.

MBartkowiakSTFC:
It still seems that the difference between the DCSF and NDTSF results is a constant factor. Unfortunately, this constant factor is sometimes negative, flipping the curves.
dcsf_fakeB
ndtsf_fakeB
I think this comes from the `normFactor` in `MDANSE.Mathematics.Arithmetic:adjust_weights`.

Here I used b_coh of O and H set to 6-5j and -4-9i, to make the imaginary part significant.

MBartkowiakSTFC:
This has been fixed in #963 instead.

PR #959, opened 2025-09-09
Fix IIR filter trajectory power spectrum graph

### LINKED ISSUE #955: [BUG] Filter designer displays incorrectly scaled spectrum for certain filter types
**Description of the error**
The filter designer widget graphs the spectra of filters and the trajectory from frequencies 0 to fs/2. 

For most of the available filters, this is default converted to rad/ps (or, angular frequency), but for a subset of them - IIR filters Notch, Peak and Comb - we plot them as a function of THz frequency.

The error observed is that while this frequency (x-axis) scaling is correct, the frequency curve (y-axis) seems to become incorrectly scaled in the process, causing it to spill of the plot and only retain a "zoomed in" subset of it.

I expected that when shifting filter type from non-IIR (i.e. Butterworth) to some IIR digital filter (one of Notch, Peak, Comb) the trajectory frequency response curve should remain unchanged, while the axis should be rescaled under conversion from rad ps^-1 to THz.

What is actually happening is that on changing to one of the problematic filter types, we see that the axis scales correctly, but the trajectory and filter frequency curves simultaneously - and incorrectly - scales so that is goes off the plot 
(it should remain fixed and look the same, since we are just converting units along the x-axis).

**Suggested fix**
Investigate the axes scaling that happens when one of the IIR filters (Notch, Peak, Comb) is selected.

**Additional details**

A) This is correct, all looks sensible - the frequency axis runs from 0 to (2pi * fs)/2 (where fs = 200 THz).
Image

B) These are wrong. The x-axis is correctly running from 0 to fs/2, but the curve has become distorted.
Image

Image

RobBuchananCompPhys:
**Description of work**
This is a fix for the filter designer issue described in #955, where the graphs for IIR filter (Notch, Peak, Comb) appear distorted and do not line up with the frequency axis.

Additionally, the code inside `set_trajectory_power_spectrum` method has been tidied up a little - some function calls have been inlined, and variable names simplified for clarity.

**Fixes**
Closes #955

Fixed by applying a conversion factor of 1/2pi to the trajectory power spectrum frequency axis if the filter is IIR (and therefore uses THz instead of rad/ps for display).

**To test**
Run standard unit tests

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
@@ -1192,7 +1192,7 @@ class FilterDesigner(QDialog):
         )
 
         if self.current_filter_units() == Filter.FrequencyUnits.CYCLIC:
-            freqs /= (2*np.pi)
+            freqs /= 2 * np.pi
 
         return (freqs, normalised, normalised * attenuation(freqs))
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
I tried it out. I think this fixed the problem.

PR #963, opened 2025-09-10
Complex weight fix: average q and -q

### LINKED ISSUE #956: [BUG] NDTSF fails for some chemical elements

ChiCheng45:
**Description of work**
Fixes the complex weight by applying,

```math
\tilde{F}_{AB}(q,t) = \frac{1}{2}(F_{AB}(q,t) + F_{AB}(-q,t)) = \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} \right\}  \mathrm{Re} \left\{ C_{AB}(q,t) \right\}
```

note that for single dimension results e.g. DOS the complex components of the weight are dropped unless it is squared which can happen when the b_coh or b_inc is used.

**Derivation of above**
In the theory of simple liquids, equation 7.1.9 for a generation correlation function $C_{AB}(t)$

```math
```
where $\epsilon_{A}$ and $\epsilon_{B}$ are the time-reversal signatures of $A$ and can be +1 or -1. So that for DCSF and related

```math
C_{AB}(q,t) = C_{AB}(q,-t) = C_{BA}^{*}(q,t)
```

since $A$ and $B$ will have the same sign for their time-reversal signatures. Also

```math
C_{AB}(-q, t) = C_{AB}^{*}(q, t) = C_{BA}(q, t).
```

We need to calculate

```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)+ \beta_{B}^{*}\beta_{A}C_{BA}(q,t) \right]
```

using the above expressions we should find that

```math
\tilde{F}_{AB}(q,t) = \frac{1}{2}(F_{AB}(q,t) + F_{AB}(-q,t)) = \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} \right\}  \mathrm{Re} \left\{ C_{AB}(q,t) \right\}
```

so that this fix means that whenever a $q$-vector is generated with a value $q$, the results returned by MDANSE will be averaged over $q$ and $-q$.

**Test Calculations**
Here is a comparison between 
```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \mathrm{Re}\left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)\right]
```

and

```math
\tilde{F}_{AB}(q,t) = \frac{1}{2}(F_{AB}(q,t) + F_{AB}(-q,t)) = \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} \right\}  \mathrm{Re} \left\{ C_{AB}(q,t) \right\}
```
using croco MD with the following atom types

image
image
image

For a small q-vectors, results can be a bit different. Here is "/dcsf/f(q,t)/C (copy)O (copy)" with 10 q-vectors and with $|q|=20$ and seed=1 (blue is the first equation)

image

here is 100

image

they are the same because I'm using lattice q-vectors and have sampled q and -q.

Here I do it again with $|q|=40$ and 10 q-vectors,
image

and 100 q-vectors

image

Basically, if we go for this correction, then the user might need to increase the number of q-vectors they use. Also, if we ever decide to allow the user to plot s(q,w) in terms of q_x, q_y, q_z, and w, we might need to take another look at this again since s(q,w) will be averaged over $q$ and $-q$.

**Fixes**
Fixes #956

**To test**
Run DCSF, DISF, and NDTSF and check results are consistent.

### DIFF
--- a/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
+++ b/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
@@ -62,7 +62,11 @@ def get_weights(
     )
     _, normFactor = adjust_weights(all_props, all_contents, n_atms, dim, conc_exp)
 
-    normalise = abs(normFactor) > 0.0  # if normFactor is 0, all weights are 0 too.
+    try:
+        # skip normalisation if weights are q-dependent e.g. x-ray SSF
+        len(normFactor)
+    except TypeError:
+        normalise = abs(normFactor) > 0.0  # if normFactor is 0, all weights are 0 too.
     if normalise:
         for k in weights:
             weights[k] /= normFactor
--- a/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
+++ b/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
@@ -62,8 +62,8 @@ def get_weights(
     )
     _, normFactor = adjust_weights(all_props, all_contents, n_atms, dim, conc_exp)
 
+    normalise = True
     try:
-        # skip normalisation if weights are q-dependent e.g. x-ray SSF
         len(normFactor)
     except TypeError:
         normalise = abs(normFactor) > 0.0  # if normFactor is 0, all weights are 0 too.
--- a/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
+++ b/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
@@ -105,11 +105,11 @@ def adjust_weights(
     weights = {}
 
     if dim == 1:
-        for el in props.keys():
+        for el in props:
             atom_conc_product = contents[el] / n_atms
             property_product = props[el]
             factor = atom_conc_product**conc_exp * property_product
-            weights[el] = factor
+            weights[(el, )] = factor
             normFactor += atom_conc_product * property_product
     elif dim == 2:
         cartesianProduct = itertools.product(props, repeat=2)
--- a/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
+++ b/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
@@ -109,9 +109,8 @@ def adjust_weights(
     weights = {}
 
     if dim == 1:
-        for el in props:
+        for el, property_product in props.items():
             atom_conc_product = contents[el] / n_atms
-            property_product = props[el]
             factor = atom_conc_product**conc_exp * property_product
             weights[(el,)] = factor
             normFactor += atom_conc_product * property_product
--- a/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
+++ b/MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
@@ -109,7 +109,7 @@ def adjust_weights(
             atom_conc_product = contents[el] / n_atms
             property_product = props[el]
             factor = atom_conc_product**conc_exp * property_product
-            weights[(el, )] = factor
+            weights[(el,)] = factor
             normFactor += atom_conc_product * property_product
     elif dim == 2:
         cartesianProduct = itertools.product(props, repeat=2)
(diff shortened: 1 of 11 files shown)


ChiCheng45:
@oerc0122 Do you know why the ruff is complaining?

MDANSE/Src/MDANSE/Mathematics/Arithmetic.py
```diff
@@ -109,24 +108,34 @@ def adjust_weights(
...
     weights = {}
 
-    cartesianProduct = itertools.product(props, repeat=dim)
-    for elements in cartesianProduct:
-        atom_conc_product = np.prod([contents[el] / n_atms for el in elements])
-        property_product = np.prod(np.array([props[el] for el in elements]), axis=0)
+    if dim == 1:
+        for el in props:
+            atom_conc_product = contents[el] / n_atms
+            property_product = props[el]
```
oerc0122: It's saying that you should iterate over `props.items()`
```suggestion
        for el, property_product in props.items():
            atom_conc_product = contents[el] / n_atms
```

ChiCheng45:
@MBartkowiakSTFC 

```math
F_{AB}(q,t) =  \frac{1}{N\sqrt{c_{A}c_{B}}} \mathrm{Re}\left[ \beta_{A}^{*}\beta_{B} C_{AB}(q,t)\right]
```

is the correct one to go for but 

```math
\tilde{F}_{AB}(q,t) = \frac{1}{2}(F_{AB}(q,t) + F_{AB}(-q,t)) = \frac{2}{N\sqrt{c_{A}c_{B}}} \mathrm{Re} \left\{\beta_{A}^{*}\beta_{B} \right\}  \mathrm{Re} \left\{ C_{AB}(q,t) \right\}
```

does give the same results for spherical lattice vector generation, given you used enough sampling so that $q$ and $-q$ was used. It is also a simpler change. Doing the first equation requires storing complex and changing the plotter to deal with complex values and scaling factors properly.

MBartkowiakSTFC:
For the record, the plotter currently has a mechanism where, if you load a complex dataset, it splits it into two float datasets, Re(x) and Im(x). Which is not the best solution, but at makes it possible to plot them. Still, I think the simpler solution will be fine, and way better than what we've had before.

ChiCheng45:
> For the record, the plotter currently has a mechanism where, if you load a complex dataset, it splits it into two float datasets, Re(x) and Im(x). Which is not the best solution, but at makes it possible to plot them. Still, I think the simpler solution will be fine, and way better than what we've had before.

Sure, but it would need to be updated so that it will plot `Re(data * scaling_factor)` rather than plotting real and imaginary components separately. Both the data and the scaling factor are complex.

MBartkowiakSTFC (approved):
I tried it out on several trajectories with real and artificial atoms. Looks good.

PR #965, opened 2025-09-10
Skip normalisation if weights are q-dependent

### LINKED ISSUE #964: [BUG] Weight normalisation with q-dependent weights.
**Description of the error**
The x-ray static structure factor uses weights which are dependent on the value of $q$ and are normalised. However it looks like the weights are normalised for each value of $q$, that means that normalisation factor is also q-dependent.

**Suggested fix**
Any job which has weights which are q-dependent should not be normalised?

ChiCheng45:
**Description of work**
On protos, the x-ray static structure factor job normalisation factor was q-dependent so that each value of q in the x-ray static structure factor was normalised with a different value. I'm not sure this is correct, perhaps it is. I need to go through the theory to check this but here is a PR which turns the normalisation off.

Here is a before and after using the large test dl poly water trajectory we have.

Before
Screenshot 2025-09-10 at 17 21 37

After
Screenshot 2025-09-10 at 17 09 36

**Fixes**
Fixes #964

**To test**
Check x-ray static structure factor from MDANSE with literature results.


MBartkowiakSTFC:
I will have to get more papers dealing with it to check what the standard convention is for x-ray results. So far I found one paper where the S(Q) still reaches 1 for large Q. At the same time, the error bars for that part of the curve are huge.

Maybe on Friday I can access more papers, since so far even PRB does not let me read anything.

gonzalezma:
Here you have a nice paper reviewing the neutron and x-ray structure factors for water and the different normalisations typically employed in the literature:
[soper13_water_structure_review.pdf](https://github.com/user-attachments/files/22270206/soper13_water_structure_review.pdf)

I did not re-read it again, but as you can see there is at least 4 different types of possible ways of normalising D(Q) for the x-rays. In any case, there is no issue in having a Q-dependent normalisation factor (see e.g. eq. 43). Basically, if you turn off the normalisation you will get closer to the raw data (Fig. 8), while the previous normalisation scheme used for the x-ray SSF calculation will give Dx(Q)+1, if I am not wrong (using normalisation II in Soper's paper I think, but I am not 100% sure).

MBartkowiakSTFC:
@ChiCheng45 Independent of which way this PR goes, please think about updating the documentation section of XSSF in `scattering.rst` to explain how the results are calculated.

ChiCheng45:
Let's close this for now and look at it again in the future.

PR #966, opened 2025-09-10
Converter atom mapping

### LINKED ISSUE #960: [BUG] DFTB converter ignores atom mapping
**Description of the error**
At the moment, the DFTB converter offers `atom_mapping` as one of the inputs. However, the element types input there are ignored and the resulting trajectory still contains the original atoms.

The converter should replace the chemical elements from the input files with those from the `atom_mapping` input.

The atom types were not changed. The output trajectory header still contains the input of `atom_mapping`, showing that it was set to replace atom types.

**Suggested fix**
1. Make sure the atoms get replaced,
2. Write a unit test specifically for testing `atom_mapping` for all the converters that support it,
3. Fix any other converter that may be found not to change atom types correctly.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
An attempt to identify and fix the converters which offer `atom_aliases` as input, but ignore the input values.

closes #960

**Fixes**
1. New unit test for converters have been added. For each converter, it tries to replace every atom in the trajectory with an Os atom, then loads the trajectory from file and check the chemical composition.
2. ASE converter has been corrected to use `atom_aliases`.
3. `XTDFileConfigurator.build_chemical_system` has been corrected to use `atom_aliases`.

**To test**
All unit tests, including the new ones, should pass.
Try converting the DFTB water trajectory and change the atom types in the converter input.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/XTDFileConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/XTDFileConfigurator.py
@@ -152,12 +152,10 @@ class XTDFileConfigurator(FileWithAtomDataConfigurator):
         coordinates = np.array(
             [atom["xyz"] for atom in self._atoms.values()], dtype=np.float64
         )
-        element_list = []
-
-        for info in self._atoms.values():
-            symbol = info["element"]
-            element = get_element_from_mapping(aliases, symbol, type=info["atom_name"])
-            element_list.append(element)
+        element_list = [
+            get_element_from_mapping(aliases, info["element"], type=info["atom_name"])
+            for info in self._atoms.values()
+        ]
         name_list = [atom["atom_name"] for atom in self._atoms.values()]
         unique_labels = set(name_list)
         label_dict = {label: [] for label in unique_labels}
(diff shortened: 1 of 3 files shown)


MDANSE/Src/MDANSE/Framework/Configurators/XTDFileConfigurator.py
```diff
@@ -152,7 +152,12 @@ def build_chemical_system(self, aliases):
         coordinates = np.array(
             [atom["xyz"] for atom in self._atoms.values()], dtype=np.float64
         )
-        element_list = [atom["element"] for atom in self._atoms.values()]
+        element_list = []
+
+        for info in self._atoms.values():
+            symbol = info["element"]
+            element = get_element_from_mapping(aliases, symbol, type=info["atom_name"])
+            element_list.append(element)
```
oerc0122:
```suggestion
        element_list = [
            get_element_from_mapping(aliases, info["element"], type=info["atom_name"])
            for info in self._atoms.values()
        ]
```

oerc0122 (approved):
Looks good, just a minor simplification

PR #967, opened 2025-09-11
Make AtomLabel usefully printable

oerc0122:
**Description of work**
Very minor "fix" found useful in debugging

**Fixes**
Adds `__repr__` to `AtomLabel` so it can be helpfully printed.

**To test**
N/A

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/AtomMapping/atom_mapping.py
+++ b/MDANSE/Src/MDANSE/Framework/AtomMapping/atom_mapping.py
@@ -86,6 +86,14 @@ class AtomLabel:
         """
         return hash((self.atm_label, self.grp_label, self.mass))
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC:
There is a difference between the way `AtomLabel` works and what `AtomLabel.__repr__` suggests it does at the moment.

I can do:
`al = AtomLabel("Xx", mass="0.505", molecule="H1/2")`
and then the `__repr__` of `al` is
`AtomLabel(atom='Fx', group='mass=0.505;molecule=H1/2')`

It is already useful information, but should we aim for the `al.__repr__` to return `AtomLabel("Xx", mass="0.505", molecule="H1/2")` in this situation?

MBartkowiakSTFC (approved):
Thanks for the changes! Looks good.

PR #968, opened 2025-09-12
Use general json loader for simpler handling

oerc0122:
**Description of work**
Implement a json_loader which handles `str`, `Path`, and `dict` seamlessly.

**Fixes**
N/A

**To test**
Should be no changes.


MBartkowiakSTFC (approved):
I tried it out. Looks good.

PR #953, opened 2025-09-04
Add ruff furb

oerc0122:
**Description of work**
Enable RUF and FURB checks

**Fixes**
Fix ruff issues raised by checks.

**To test**
Standard tests

PR #972, opened 2025-09-15
Pin ruff version to prevent new errors in old PRs

oerc0122:
**Description of work**
Prevent issue of newer ruff versions adding checks which invalidate "good" PRs.

**Fixes**
Pin ruff versions in lint

**To test**
N/A

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,6 +15,7 @@ jobs:
         with:
           src: "./MDANSE/Src"
           args: "format --check"
+          version: "0.13.0"
 
   lint_check_ruff:
     runs-on: ubuntu-latest
@@ -24,6 +25,7 @@ jobs:
         with:
           src: "./MDANSE/Src"
           args: "check"
+          version: "0.13.0"
 
   lint_format_ruff_gui:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC:
Have we had problems with different ruff versions? Or are there some planned changes in ruff?

oerc0122:
The bump to 0.13.0 introduced some new FURB errors on #953 . This will cause issues with PRs in future containing unrelated changes just to satisfy ruff and this will prevent that in future.

MBartkowiakSTFC:
So, to make sure I got this one right: ruff 0.13 is the latest one, and we are pinning the version number to this one. Does this mean that the code that is compatible with ruff 0.13 is not compatible with previous versions, and vice versa?

I just think I missed some detail here, because the PR description says `Prevent issue of newer ruff versions adding checks which invalidate "good" PRs`, but we are apparently pinning the `ruff` version to `0.13` which is the newest one.

oerc0122:
I pinned to `0.13` as `0.12` -> `0.13` were broken (and then fixed) in the `FURB` PR (#953). Since the new issue came with the addition of `FURB` and `RUF` linters, our main branch is still valid with `0.13`.

MBartkowiakSTFC (approved):
Thanks for explaining. Let's merge it.

Issue #971, opened 2025-09-15
[ENHANCEMENT] Badges/metrics for the project page


MBartkowiakSTFC:
**Is your feature request related to a problem? Please describe.**
We have been asked to include some metrics on how many times MDANSE has been downloaded.

**Describe the solution you'd like**
We could add badges to the MDANSE page to show the number of PyPI downloads for MDANSE and MDANSE_GUI.

**Describe alternatives you've considered**
There is already some information in the 'Insights' section.

**Additional context**
N/A

MBartkowiakSTFC:
closed by #973

PR #973, opened 2025-09-15
Add badges for relevant stats

oerc0122:
**Description of work**
Add badges to MDANSE. 

Also add proper version labels such that the Python version badge becomes useful.

**Fixes**
#971

**To test**
See: https://github.com/oerc0122/MDANSE/tree/add-badges?tab=readme-ov-file


MBartkowiakSTFC (approved):
Thanks for the changes, looks good!

Issue #944, opened 2025-09-02
[ENHANCEMENT] A separate converter for XYZ files


MBartkowiakSTFC:
**Is your feature request related to a problem? Please describe.**
Normally, we use the ASE converter to read the `.extxyz` files. However, some software (DFTB+) seems to use XYZ format with additional columns without writing the header explaining the contents. Other software packages like CP2K can dump multiple XYZ files, using one for positions, another for velocities and yet another for forces.

We should consider creating a general XYZ converter, which can extract information from XYZ files, and allows the user to fill in the missing details.

**Describe the solution you'd like**
The XYZ converter should allow to read the following variables:
1. positions,
2. velocities,
3. forces,
4. charges,
5. unit cell (with an optional manual input for fixed cells, or input from a non-xyz file).

For each of these, it should be possible to specify:
- which file they are in,
- which columns they are in,
- what physical unit it is.

This way we can work both with one file with multiple columns (DFTB-style), or with multiple files (CP2K-style).

**Describe alternatives you've considered**
We could add more and more inputs and special cases to the existing converters, but I would prefer not to.

**Additional context**
N/A

PR #975, opened 2025-09-16
Fix CP2K parser and add separate basic xyz parser

oerc0122:
**Description of work**
Implement and use newer XYZ parser for CP2K avoiding issue with file-length mismatches with old one.

**Fixes**
Addresses #944 
Fixes @ChiCheng45 reported data issues.

**To test**
Standard tests.


ChiCheng45 (commented):
Works great. Just the ruff warnings and the xyz.py file is missing the MDANSE header, otherwise can approve..

ChiCheng45 (approved):

PR #980, opened 2025-09-17
Enable comprehesions and quotes checking

oerc0122:
**Description of work**
Enable ruff comprehensions rules and quotes (no quotes changes) and fix result.

**Fixes**
N/A

**To test**
Should be no changes


MBartkowiakSTFC (approved):
I tried it out, seems to work the same as before.

PR #981, opened 2025-09-17
Rework parsers

oerc0122:
**Description of work**
Move `Parsers` out of `Configurators` and use them as needed.

Separation should increase maintainability.

This opens up future opportunities to streamline and unify converter jobs, as well as parallelise them.

**Fixes**
- Updates `FileWithAtomDataConfigurator` to take a mandatory `parser` argument to convert the file on use. For old behaviour this should be set to `self.parse` (c.f. MDAnalysis/MDTraj topologies)
- Add new parsers and use in `Converter` jobs
- Remove old configurators
- Update `IConfigurator` to have explicit arguments rather than anonymous kwargs.
- MDAnalysis and MDTraj converter left as-is due to complexity of parser.

**To test**
Standard tests.


MBartkowiakSTFC:
At the moment, the file input widgets fail to initialise correctly in these converters:
- ASE
- CASTEP
- CP2K
- DFTB
- DL_POLY
- Forcite
- LAMMPS (configuration file only)
- VASP

The GUI creates a backup widget with text field input only, and the specialised widget with the "Browse" button is not created for these converters.

oerc0122:
> At the moment, the file input widgets fail to initialise correctly in these converters:
> 
> * ASE
> * CASTEP
> * CP2K
> * DFTB
> * DL_POLY
> * Forcite
> * LAMMPS (configuration file only)
> * VASP
> 
> The GUI creates a backup widget with text field input only, and the specialised widget with the "Browse" button is not created for these converters.

Should be fixed. I didn't notice in my quick test since it produced the text-box and I didn't notice the lack of button

MBartkowiakSTFC:
> Should be fixed. I didn't notice in my quick test since it produced the text-box and I didn't notice the lack of button

We can come up with better way of handling these problems on the GUI side. Having a backup widget was useful in the early development stages of the GUI, when some of the widgets still had to be written. At the current stage we could already decide to be more restrictive and have the GUI fail completely instead, or at least output a big and obvious error message.

oerc0122:
> > Should be fixed. I didn't notice in my quick test since it produced the text-box and I didn't notice the lack of button
> 
> We can come up with better way of handling these problems on the GUI side. Having a backup widget was useful in the early development stages of the GUI, when some of the widgets still had to be written. At the current stage we could already decide to be more restrictive and have the GUI fail completely instead, or at least output a big and obvious error message.

Ideally, we should develop some GUI tests which work. Trying to virtually build all of the tabs for every job type, etc. and search for GUI warnings about bad Widgets.

MBartkowiakSTFC:
So far I have had problems with only two things in MDANSE_GUI:
1. CP2K unit cell parser. The GUI throws an error saying "'str' object has no attribute 'open'" (`CP2K_Cell.py`, line 42),
2. Forcite converter, MDANSE-Examples/Data/Trajectories/Forcite/nylon66_rho100_500K_v300K.xtd file. `xtd.py` line 108 `Key error: "Connects"`. (The nylon example causes the problem, the other trajectory works.)

MBartkowiakSTFC (approved):
I did not find any more problems. Looks good.

ChiCheng45 (approved):
Looks good, I don't see any issues on my end.

PR #982, opened 2025-09-17
Update docs so that complex scattering lengths are used

ChiCheng45:
**Description of work**
Updated documentation so that it is consistent with the recent update made in #963. Daggers are added to complex scattering lengths in some equations. An explaination about the averaging over $q$ and $-q$ is added to the weights section. Explicitly written where the real part of some function are taken in some equations. Feel free to edit and commit over this PR.

### DIFF
--- a/Doc/pages/atom_selection.rst
+++ b/Doc/pages/atom_selection.rst
@@ -182,12 +182,18 @@ with selection is
 .. math::
    :label: selection1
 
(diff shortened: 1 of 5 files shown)


MBartkowiakSTFC (approved):
Looks good, as long as we agree to use `\dagger` as the symbol for complex conjugate. I added a line to "symbols and notations" to say specifically that b_coh^{\dagger} is the c.c. If you agree with my changes, you are welcome to merge it.

PR #983, opened 2025-09-17
Fixed GUI periodic table information display on hover

ChiCheng45:
**Description of work**
Fixed GUI periodic table information display on hover.

image

**Fixes**
Periodic table element information not displayed on hover.

**To test**
Start the GUI and open the periodic table. Hover the mouse over some elements and check that the element information is displayed.


MBartkowiakSTFC (approved):
Thanks, this fixed it.

PR #985, opened 2025-09-18
Run IJob.check_status before IJob.initialize

MBartkowiakSTFC:
**Description of work**
It seems that a lot of jobs with bad inputs fail during `IJob.initialize` and do not generate a meaningful error, because the bad error parameters are only reported in `IJob.check_status`. This PR makes sure that `check_status` is run before `initialize`.

**Fixes**
1. Swapped 'IJob.initialize()` with `IJob.check_status()` in `IJob.run`.

**To test**
Save any analysis script and change the parameters in the script to some wrong values. Run the script and check if the error output matches the wrong input variable.


ChiCheng45 (approved):
Looks good, I see correct error messages now.

PR #987, opened 2025-09-22
Disable save button on invalid Action script

### LINKED ISSUE #976: [BUG] Disallow save on invalid script
**Description of the error**
If a set of parameters are invalid we disallow running the job, but still allow people to save.

While an invalid script is visually flagged, given the save is still available some people may ignore errors and try to fix them in post.

Invalid scripts are saveable.

**Suggested fix**
Disable the save button on error?

**Additional details**
N/A

RobBuchananCompPhys:
**Description of work**
When a script is invalid, but Run and "Save as Script" are disabled.

**Fixes**
Closes #976 

**To test**
Run GUI, create invalid script via Action input field.


MBartkowiakSTFC (approved):
I tried it for both the converters and analysis tabs. Looks good.

PR #977, opened 2025-09-16
Improve Q vector generation mechanism

MBartkowiakSTFC:
**Description of work**
Includes off-diagonal components of HKL indices when finding the extent of HKL values to be used in vector generation.

closes #974

**Fixes**
1. `SphericalLatticeQVectors` uses all vector components when checking the HKL range to cover.
2. Q vector generators calculate HKL values of vectors if a unit cell has been provided, even if the vector generator does not use the unit cell definition to generate vectors.
3. `IQVectors.create()` now takes a `UnitCell` instead of a `Configuration` as optional input.
4. Unit tests have been added to check if spherical vectors are generated the same for equivalent unit cell definitions.

**To test**
Unit tests should all pass.

PR #990, opened 2025-09-23
Visualise Q vector spread

MBartkowiakSTFC:
**Description of work**
While we already save q-vector information, there is no easy way of viewing a summary of the information.

This PR adds an option of creating a plot from any .mda file that contains a group called "vector_generator".
Screenshot 2025-09-23 at 14 00 58

At the moment, two data sets are put in the same plot:
1. number of vectors per shell,
2. histogram of |q| for each shell.
Screenshot 2025-09-23 at 14 07 02

**Fixes**
1. Added functions for summarising q vector information,
2. Added new plotting option to the plot selector context menu,
3. Enabled `SingleDataset` creation from a manually supplied data array, instead of loading from an .mda file.

**To test**
Run GUI unit tests.
Start the gui and load an .mda file of DCSF or DISF results. Right-click the file in Plot Creator and choose "Vector summary". A new plot should be created.


ChiCheng45:
Looks good, should be useful. Just one thing I think it might be useful to have the q-vector summary plotted as a histogram with stacked bars instead https://matplotlib.org/stable/gallery/statistics/histogram_multihist.html#sphx-glr-gallery-statistics-histogram-multihist-py. 

I think the binning with the line plot makes it difficult to see what the actually minimum value in the each q-vector shell is.
Screenshot 2025-09-24 at 10 41 11

I used the spherical q vector generator with width 2 and a shell spacing of 2. With stacked histogram I should also see a constant value across the histogram which is difficult to determine with the line plot.

MBartkowiakSTFC:
We can make a plot like this:
vector_count_plot
It is interesting if you want to see the total population of the q-vector space. It is not great if you wanted to check how well the shells are centered on the nominal value of |q|. 

A less extreme example is here, with a lower value of width:
narrower

ChiCheng45:
> We can make a plot like this: vector_count_plot It is interesting if you want to see the total population of the q-vector space. It is not great if you wanted to check how well the shells are centered on the nominal value of |q|.
> 
> A less extreme example is here, with a lower value of width: narrower

True, I suppose you could plot just one of them at a time. Anyway, I can see the expected results for the SphericalQVectors. 

image

The numbers in each bin are nearly the same across the range |q|. The edges are lower because the bins start from a value smaller than the smallest value of |q| set. Can you update it so the bins start from the smallest value of |q| and end with the largest from the qvector setting?


Also, here is a 0.1 0.1 0.1 nm unit cell with the SpericalLatticeQVector generator, so the problem we were talking about will be exaggerated. 

image

I used shells = (10, 20, 2), n_vectors=1000, and width=2. So with the red shell, it is weighted to the larger values for |q| for that bin. I think the problem we see here is that with n_vectors=1000, we use all the vectors that are available for each distance of |q|, and for some values of |q|, there will just be more points.

MBartkowiakSTFC:
> Also, here is a 0.1 0.1 0.1 nm unit cell with the SpericalLatticeQVector generator, so the problem we were talking about will be exaggerated.
>
> I used shells = (10, 20, 2), n_vectors=1000, and width=2. So with the red shell, it is weighted to the larger values for |q| for that bin. I think the problem we see here is that with n_vectors=1000, we use all the vectors that are available for each distance of |q|, and for some values of |q|, there will just be more points.

When I tried reproducing this result, I got no vectors at all with the parameters above. Main branch tells me it can't be done, and my new branch generated one shell with one vector, 0 0 0. (Which was not meant to happen and I will have to correct this.)

ChiCheng45:
> > Also, here is a 0.1 0.1 0.1 nm unit cell with the SpericalLatticeQVector generator, so the problem we were talking about will be exaggerated.
> > I used shells = (10, 20, 2), n_vectors=1000, and width=2. So with the red shell, it is weighted to the larger values for |q| for that bin. I think the problem we see here is that with n_vectors=1000, we use all the vectors that are available for each distance of |q|, and for some values of |q|, there will just be more points.
> 
> When I tried reproducing this result, I got no vectors at all with the parameters above. Main branch tells me it can't be done, and my new branch generated one shell with one vector, 0 0 0. (Which was not meant to happen and I will have to correct this.)

I just checked, and I think I must've used the trajectory with the normal unit cell with 1.44, 1.44, 1.44 nm, the dftb water trajectory. Sorry about that.

MBartkowiakSTFC:
This is an example of what happens for `SphericalLatticeQVectors` with the latest version of the vector statistics plotter:

Screenshot 2025-09-26 at 15 31 56

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -134,6 +157,66 @@ def __init__(self, name: str, source: h5py.File | None, linestyle: str = "-"):
 
         self.create_axes_tags(self._axes_tag, source)
 
+    def init_manually(
```
oerc0122: Given the naming of other parts does this want to be called something like `configure`?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -82,12 +82,19 @@ def get_mpl_colours():
 class SingleDataset:
     """Manages a plottable data set from an .mda file."""
 
-    def __init__(self, name: str, source: h5py.File | None, linestyle: str = "-"):
+    def __init__(
+        self,
+        name: str,
+        source: h5py.File | None,
+        linestyle: str = "-",
+        marker: str | None = None,
```
oerc0122, line 91: I wonder if we add a `**kwargs` to this and pass it through to a `configure` function at the end. 

This `configure` function could be a `functools.singledispatchmethod` dispatching on whether `data` is an `h5py.File` or a `np.ndarray` and calling the appropriate sub-functions.

This may be a bit too much of a rewrite for this PR, however, and it's important to get the functionality in soon.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def slider_labels(self) -> list[str]:
+        """Return labels to show that sliders are not used."""
+        return ["Inactive", "Inactive"]
+
+    def slider_limits(self) -> list[str]:
+        """Return generic slider limit values."""
+        return self._number_of_sliders * [[-1.0, 1.0, 0.01]]
+
+    def clear(self, figure: Figure = None):
```
oerc0122:
```suggestion
    def clear(self, figure: Figure | None = None):
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+        """Clear the figure."""
+        target = self._figure if figure is None else figure
+        if target is None:
+            return
+        target.clear()
+
+    def get_figure(self, figure: Figure = None):
```
oerc0122:
```suggestion
    def get_figure(self, figure: Figure | None = None):
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+
+        """
+        super().change_normalisation(new_value)
+
+    def check_curve_lengths(self):
+        """Find the maximum number of elements in the x axes of the plot data."""
+        self.curve_length_limit = 0
+        for num, _ in enumerate(self._active_curves):
+            xdata = self._backup_curves[num][0]
+            self.curve_length_limit = max(self.curve_length_limit, len(xdata))
```
oerc0122:
```suggestion
        self.curve_length_limit = max(
            len(self._backup_curves[num][0])
            for num, _ in enumerate(self._active_curves)
        )
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def plot(
+        self,
+        plotting_context: PlottingContext,
+        figure: Figure = None,
+        update_only=False,
+        toolbar=None,
```
oerc0122:
```suggestion
        figure: Figure | None = None,
        update_only: bool = False,
        toolbar: type | None = None,
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+        self._active_curves = []
+        self._backup_curves = []
+        self._normalisation_errors = []
+        self._axes = []
+        self.apply_settings(plotting_context)
+        x_axis_labels = []
+        if plotting_context.set_axes() is None:
+            LOG.debug("Axis check failed.")
+            return
+        if len(plotting_context.datasets()) == 0:
```
oerc0122:
```suggestion
        if not plotting_context.datasets():
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+            try:
+                best_unit, best_axis = (
+                    dataset._axes_units[databundle.main_axis],
+                    databundle.main_axis,
+                )
+            except KeyError:
+                best_unit, best_axis = dataset.longest_axis()
+            plotlabel = databundle.legend_label
+            x_axis_labels.append(dataset.x_axis_label(best_axis))
+            if dataset._name == "Available vectors":
```
oerc0122, line 129: This would probably be best as a `match` now we've moved to 3.10. Not terribly important though.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Vectors.py
```diff
@@ -0,0 +1,247 @@
...
+                            label=plotlabel + ":" + dataset._curve_labels[key],
+                            bottom=bottom,
+                            width=width,
+                            edgecolor="black",
+                        )
+                        bottom += value
+                    except ValueError:
+                        LOG.error(f"Plotting failed for {plotlabel} using {best_axis}")
+                        LOG.error(f"x_axis={dataset._axes[best_axis]}")
+                        LOG.error(f"values={value}")
```
oerc0122, line 215: Strictly logs should use the form:
```python
LOG.error(f"Plotting failed for %s using %s", plotlabel, best_axis)
```
Because of lazy evaluation of strings for where logs aren't issued. 

Again, not terribly important.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
```diff
@@ -15,16 +15,112 @@
...
+    -------
+    tuple[SingleDataset, SingleDataset]
+        1D array of vector count vs. |q|, 2D histogram of q vector counts per shell.
+    """
+    parent_dset = file[main_dstet]
+    qvals = parent_dset["q"][:]
+    nshells = len(qvals)
+    modq_per_shell = [shell_to_modq(n, parent_dset) for n in range(nshells)]
+    qmin, qmax = np.min(modq_per_shell[0]), np.max(modq_per_shell[nshells - 1])
+    q_step = np.mean(np.abs(qvals[1:] - qvals[:-1]))
```
oerc0122:
```suggestion
    q_step = np.mean(np.abs(np.diff(qvals)))
```
[`np.diff`](https://numpy.org/doc/stable/reference/generated/numpy.diff.html#numpy.diff)

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
```diff
@@ -15,16 +15,112 @@
...
+    bin_step = max(
+        0.4 * np.min([np.std(one_shell) for one_shell in modq_per_shell]), 0.05 * q_step
+    )
+    common_bins = np.arange(max(0.0, qmin), qmax + 1.1 * bin_step, bin_step)
+    qmod_histograms = [np.histogram(qmods, common_bins)[0] for qmods in modq_per_shell]
+    xvals = (common_bins[:-1] + common_bins[1:]) / 2
```
oerc0122:
```suggestion
    xvals = -np.diff(common_bins) / 2
```

oerc0122 (commented):
As mentioned in meeting, would be nice to have a preview button on the GUI panel.

MBartkowiakSTFC:
`QVectorsWidget` now offers a preview option, which pops up a limited plotting widget showing the vector statistics plots.

ChiCheng45 (commented):
Looks good so far. Just a few things from me.

Preview vector can be opened even for invalid q_vector settings.

Image

The population histogram in the plotter is zoomed out for some reason.

Image

This is fine in the q vector previewer.

The grid and legend can take a while to update. It's fine but if there is an easy optimisation would be good to have.

Image

MBartkowiakSTFC:
I think I fixed all the problems, except for the histogram x-axis range being too large in some cases. It seems that the files generated with #977 don't normally create that problem in any case. Do you think this PR may be good enough for now?

ChiCheng45:
> I think I fixed all the problems, except for the histogram x-axis range being too large in some cases. It seems that the files generated with #977 don't normally create that problem in any case. Do you think this PR may be good enough for now?

Yes looks good. Just a few more thing I spotted. When right click plotting q-vectors, the file information on the table shows no file but I think you could put the file of the MDA file here right?

Screenshot 2025-10-22 at 16 48 22

Finally I also found some issues when a trajectory without unit cell data is used with the lattice based q-vector generators. But I think this issue is due to #1042 rather than this PR.
vector data

MBartkowiakSTFC:
@ChiCheng45 I think I fixed both problems (file name in plot details and the error handling for trajectories without the unit cell). What do you think?

ChiCheng45 (approved):
Yes, looks great. I don't see any more issues.

PR #991, opened 2025-09-25
Add atom mapping to converters which use PDB files.

### LINKED ISSUE #333: [ENHANCEMENT] Uniform converter behaviour
There is different amount of information provided about the system by different MD engines, which requires some converters to compensate for the missing bits with additional code. Specifically, some engines assign arbitrary labels to atoms, making it difficult to assign correct chemical elements to objects in the trajectory. The way this is handled should be made uniform between converters.

Every converter should have the following features:
- give a preview of atom types it will load,
- optionally, guess the atom types based on mass,
- optionally, change the atom types **before** the conversion,
- scan for molecules/bonding information during conversion.

At the moment, two converters (LAMMPS and ImprovedASE) have chemical element recognition based on atom mass, and DL_POLY uses molecules.json for atom type lookup. Other converters rely on accurate input from the MD files.

**Additional context**
The converters will always create trajectories with all the atoms inside the box, at the expense of some molecules not being contiguous. If an analysis type needs contiguous molecules, it needs to contain code that will ensure this condition is met.

### LINKED ISSUE #978: [BUG] Minimal PDB file reader does not recognise some atom types
**Description of the error**
There are some atoms types which are not recognised by the minimal PDB reader e.g. SOD for sodium which is quite commonly used. Previously an alternatives key existed in the atom databased which stored alternative names for each atom type e.g. SOD for Na and was used by the PDB reader.

PDB reader should recognise some common aliases for some atom types.

**Suggested fix**
Add atom mapping to converter which use the PDB reader and add an atom type alias mechanism.

ChiCheng45:
**Description of work**
Added atom mapping to converters that use PDB files. PDB files' elements are determined from atom names and are parsed before atom mapping is done to ensure that things like CA get converted to carbon and CAL to calcium. Some common atom name mapping, like SOD to Na, are taken from https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/guesser/tables.py.

**Fixes**
Fixes #978
Closes #333

**To test**
Run trajectory conversion with the GROMACS or the DCD converter with different atom mapping and check results follow expectations.


MBartkowiakSTFC:
I think that this PR will also close #333.

MBartkowiakSTFC (approved):
Looks good. The output trajectories contain the atoms I selected in atom mapping.

PR #992, opened 2025-09-25
render math in job tab info

RobBuchananCompPhys:
**Description of work**
Currently, the Job tab info does not render the embedded LaTex expression strings, which can look less informative and cluttered. This PR proposes an enhancement in which images (rendered in a simple `matplotlib` figure) of the expression are inserted into the html of the Job tab, allowing mathematical expressions to be displayed in the text.

**Fixes**
Before:
Screenshot 2025-09-26 094524

After:
Screenshot 2025-09-26 094921

**To test**

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import base64
 import io
 import re
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
@@ -113,15 +113,28 @@ class MathRenderer:
         return f"${text}$"
 
     @staticmethod
-    def render(expression: str) -> None:
+    def render(expression: str, dark: bool = False) -> None:
         # Create a figure containing the rendered LaTex expression
         fig, ax = plt.subplots(figsize=(0.01, 0.01))
         ax.axis("off")
-        fig.text(0, 0, MathRenderer.mask(expression), fontsize=7)
+        fig.text(
+            0,
+            0,
+            MathRenderer.mask(expression),
+            fontsize=7,
+            color="white" if dark else "black",
+        )
 
         # Save the image as bytes
         buffer = io.BytesIO()
-        plt.savefig(buffer, format="png", bbox_inches="tight", pad_inches=0.1, dpi=150)
+        plt.savefig(
+            buffer,
+            format="png",
+            bbox_inches="tight",
+            pad_inches=0.1,
+            dpi=150,
+            transparent=dark,
+        )
         plt.close(fig)
         buffer.seek(0)
         image = base64.b64encode(buffer.read()).decode("utf-8")
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
@@ -2,7 +2,6 @@ import re
 import io
 import base64
 from matplotlib import pyplot as plt
-from typing import List, Any
 
 
 class MathRenderer:
@@ -23,12 +22,12 @@ class MathRenderer:
     def ignore(text: str) -> bool:
         return text in MathRenderer.ignores
 
-    def scan(self) -> List[Any]:
+    def scan(self) -> dict[str, bool]:
         # Use regex matching to find expressions
         pattern = r"(:math:`.*?`)"
         substrings = re.split(pattern, self.raw_text)
 
-        scanned = []
+        scanned = dict()
         for s in substrings:
             if not s:
                 # This is not a string - skip
@@ -45,10 +44,10 @@ class MathRenderer:
             if s.startswith(":math:`") and s.endswith("`"):
                 # This is a raw LaTex expression string - instantiate object to identify it as such
                 expr = s[len(":math:`") : -1]
-                scanned.append(MathExpression(expr))
+                scanned.update({expr: True})
             else:
                 # Normal text string
-                scanned.append(s)
+                scanned.update({s: False})
 
         return scanned
 
@@ -85,11 +84,3 @@ class MathRenderer:
     @classmethod
     def from_cache(cls, key):
         return cls.cache[key]
-
-
-class MathExpression:
-    def __init__(self, expr: str):
-        self.expr = expr
-
-    def get(self) -> str:
-        return self.expr
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
@@ -22,7 +22,7 @@ class MathRenderer:
 
     # Line breaks used
     BREAK = r"<br\s*/?>"
-    DOUBLE_BREAK = f"{BREAK}\s*{BREAK}"
+    DOUBLE_BREAK = rf"{BREAK}\s*{BREAK}"
 
     def __init__(self, text: str, dark: bool = False) -> None:
         self.raw_text = text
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
Looks like it's almost there. A few possible points that could still be worked on:

1. If an equation is not inline, it is not picked up:

Screenshot 2025-09-26 at 10 58 32


2. The ignored expressions still affect the line breaks:

Screenshot 2025-09-26 at 10 58 13

3. In principle, it _may_ be worthwhile to try to inline the images. With some tweaking, I got to the stage where it looked like this:
Screenshot 2025-09-26 at 11 18 37
But if the vertical alignment cannot be corrected easily and the whole thing looks to awkward, then it should be fine to stick to the current version with a line break per maths symbol.

ChiCheng45 (commented):
Just a few things.

Image

White background with dark mode theme, and the inline equations' font size is very large. For the white background, I guess we could make the text white and the background transparent. The font size issue might be trickier. I wonder if we should latex the whole thing?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @staticmethod
+    def containsBlockExpression(text: str) -> bool:
+        return text.startswith(".. math:")
+
+    @staticmethod
+    def containsInlineExpressions(text: str) -> bool:
+        pattern = r"(:math:`.*?`)"
+        substrings = re.split(pattern, text)
+        return len(substrings) > 1
```
oerc0122:
```suggestion
        return re.search( r"(:math:`.*?`)", text) is not None
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
```diff
@@ -0,0 +1,143 @@
+from __future__ import annotations
+
+import base64
+import io
+import re
+
+from matplotlib import pyplot as plt
+
+
+class MathRenderer:
```
oerc0122, line 10: Really, this would be an ideal use-case of 3.14's [t-strings](https://docs.python.org/3.14/library/string.templatelib.html#template-strings)

RobBuchananCompPhys:
> Just a few things.
> Image
> 
> White background with dark mode theme, and the inline equations' font size is very large. For the white background, I guess we could make the text white and the background transparent. The font size issue might be trickier. I wonder if we should latex the whole thing?

Good question - we may be able to get away with transparency and white font when in dark mode.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
```diff
@@ -0,0 +1,143 @@
...
+        substrings = re.split(pattern, text)
+        scanned = []
+        for s in substrings:
+            if s.startswith(":math:"):
+                expr = s[len(":math:`") : -1]
+                scanned.append((expr, True))
+            else:
+                scanned.append((s, False))
+
+        return scanned
```
oerc0122: This is a somewhat strange way to do this. The usual way would be to use Regular expression groups.

```suggestion
    def processInlineExpressions(text: str) -> list[tuple[str, bool]]:
        pattern = r"(:math:`(.*?)`)"  # <-- Should really be class-level constant
        last = 0
        for match in re.finditer(pattern, text):
             scanned.append((text[last:match.start], False))
             scanned.append((match.group(1), True))
             last = match.end
        scanned.append((text[last:], False))
        
        return scanned
```

Although I'm still not sure whether you really want to capture the non-converting bits or whether you just want to substitute the matches in-place for their image equivalents, which simplifies the logic massively.

```suggestion
    def processInlineExpressions(self, text: str) -> None:
        pattern = r"(:math:`(.*?)`)"  # <-- Should really be class-level constant

        for match in re.finditer(pattern, text):
             text.replace(match[0], self.render(match[1]))
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
```diff
@@ -0,0 +1,143 @@
...
+            group.append(s)
+
+        group = [s.replace(r"<br />", "") for s in group]
+        total = "".join(group)
+        if total.startswith(r"<br />") and total.endswith(r"<br />"):
+            result = total.split(r"<br />")[1]
+        else:
+            result = total.split(r"<br /><br />")[0]
+
+        return [(result, True)], n + len(group) + 1
```
oerc0122: Again, you can use pattern matching here.

```suggestion
    def processMultiLineBlockExpression(
        self, strings: str
    ) -> None:
        pattern = r".. math::\s*\n(?:\s*\n)?(\s+)(.+)\n(\1.+\n)*"  # <- Should be class level variable
        
        for match in re.finditer(pattern, text):
            if match[3] is not None:
                body = "".join(more_itertools.prepend(match[2], match[3].splitlines()))
            else:
                body = match[2]
            text.replace(match[0], self.render(body))
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/MathRenderer.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @staticmethod
+    def containsInlineExpressions(text: str) -> bool:
+        pattern = r"(:math:`.*?`)"
+        substrings = re.split(pattern, text)
+        return len(substrings) > 1
+
+    @staticmethod
+    def processBlockExpression(text: str) -> list[tuple[str, bool]]:
+        return [(text[len(".. math:: ") :], True)]
```
oerc0122: As above, we can use pattern matching:
```suggestion
    def processBlockExpression(text: str) -> list[tuple[str, bool]]:
        pattern = ".. math::\s*(.+)"
        
        for match in re.finditer(pattern, text):
            text.replace(match[0], self.render(match[1]))
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/TextInfo.py
```diff
@@ -47,3 +49,53 @@ def filter(self, some_text: str, line_break="<br />"):
...
+        for token, is_expression in scanned:
+            if is_expression and not MathRenderer.cached(token):
+                renderer.render(
+                    token,
+                    QApplication.styleHints().colorScheme() == Qt.ColorScheme.Dark,
+                )
+
+        return scanned
+
+    def filter(self, some_text: str, line_break="<br />"):
```
oerc0122: `filter` is a Python builtin and shouldn't be used as a name

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/TextInfo.py
```diff
@@ -47,3 +49,53 @@ def filter(self, some_text: str, line_break="<br />"):
...
+        filtered = super().filter(some_text, line_break)
+        scanned = self.scan(filtered)
+
+        html_substrings = []
+        for token, is_expression in scanned:
+            if is_expression:
+                image = MathRenderer.from_cache(token)
+                if len(token) < 10:
```
oerc0122: Bear in mind that latex tokens may not correspond to large images:

`\underbrace{\mathbf{x}} = \mathbf{3}`

$\underbrace{\mathbf{x}} = \mathbf{3}$

oerc0122 (commented):
I think the MathRenderer can be simplified by using REs to their full extent. Though you may have to be careful and make a copy of the text since `replace` may disrupt `finditer`.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/TextInfo.py
RobBuchananCompPhys: Yep, good point - this method name will also need to be changed in the overridden function (`TextInfo.filter`)

ChiCheng45:
I don't think this needs to be fix here but something I discover in #1038 which also will happen for the equation images.

auto-dark-light-new

ChiCheng45:
> I don't think this needs to be fix here but something I discover in #1038 which also will happen for the equation images.
> 

I've been able to fix some dark light mode system change GUI issues. See #1062 might be worth trying to add a similar fix here I think.

ChiCheng45:
A few more things. Some text has been dropped. The "for a set of atoms" is missing. Also the equations are on their own line, I think it would be nice if you could inline them.

image

image

here is the text on protos.

image

@RobBuchananCompPhys Can you resolve the merge conflict, I can have a look to see if dark/light mode change can be fixed for the equation images.

PR #998, opened 2025-10-01
Axis preview update fix

### LINKED ISSUE #994: [BUG] Energy axis preview does not update
**Description of the error**
The predicted values of energy in the output files should depend on the correlation frames setting in the analysis, but at the moment changing the correlation frames setting does not affect the energy.

Example screenshots:

Image

Image

Every time we change the input of the correlation frames `Configurator`, the predicted energy axis of the results should be updated in the preview.

The predicted energy axis corresponds to (0, MAX_FRAMES, 1, MAX_FRAMES/2) independent of the actual numbers.

**Suggested fix**
Find out why the energy axis does not get recalculated.

**Additional details**
Other axes (e.g. time) still get updated in the preview, so the problem is specific to energy.

ChiCheng45:
**Description of work**
Fixes the configurators so that if a configurator is updated, other configurators that dependent on it will also be updated.

**Fixes**
Fixes #994

**To test**
Check that the axes preview updates correctly for a few different analysis jobs.


MBartkowiakSTFC (approved):
It works now. Also, I think this mechanism will be useful for fixing other problems later (#1002)

PR #999, opened 2025-10-01
Remove last glob-based file loading from GUI

oerc0122:
**Description of work**
Replace globbing with explicit imports.

**Fixes**
Turn globs in `__init__.py` in MDANSE_GUI into explicit `import`s.

**To test**
Standard GUI tests should work.


MBartkowiakSTFC (approved):

PR #1003, opened 2025-10-01
Correct atom transmutation configurator error status

### LINKED ISSUE #988: [BUG] Atom transmutation widget does not clear errors
**Description of the error**
If I manually edit the text input of `AtomTransmutationWidget`, it starts showing an error (which is expected) but the error does not disappear, even if I reset the to the original value:

Original input, considered valid:
Image

The same input, but now marked as wrong:
Image

If the `{}` was a valid value in the beginning, it should be also valid later.

Once an error status has been set on atom transmutation, it does not go away.

**Suggested fix**
Maybe a `self.error_status = "OK"` is missing in the `AtomTransmutationConfigurator.configure`?

**Additional details**
N/A

ChiCheng45:
**Description of work**
Set error status to OK when no transmutations are carried out

**Fixes**
Fixes #988

**To test**
In the GUI manually edit the atom transmutation line and check that the error clears correctly.


MBartkowiakSTFC (approved):
Looks good.

PR #1005, opened 2025-10-02
Use an enum for grouping levels

oerc0122:
**Description of work**
Replace strings with `Enum`.

**Fixes**
N/A

**To test**
Standard tests


MBartkowiakSTFC (approved):
Do you think it would be good to have a docstring for UCEnum, just to leave a hint as to why we need it? I'll leave it up to you to decide.

It works, everything seems fine.

MBartkowiakSTFC (approved):
Thanks for changes, looks good.

PR #1010, opened 2025-10-03
Some GUI adjustments.

ChiCheng45:
**Description of work**

- I have moved the text info box on the running jobs tab to the bottom left. This is done so it is consistent with the converter, trajectory, and action tabs. Added header and footer to all text info boxes.

image

- The output prediction box is alway updated so that its height it big enough so that we don't need to use a scrollbar.

image

- The plotter on the plot tab is streched is it is larger. The difference is most apparent when you maximise the MDANSE window.

image

**To test**
Test the the plotter works correctly,the analysis job output prediction box updates correctly, and the running job info text information updates correctly.


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
```diff
@@ -419,6 +420,17 @@ def show_output_prediction(self):
                 else:
                     text += f"<p>[{array[0]}, {array[1]}, {array[2]}, ..., {array[-1]}] ({new_unit})</p>"
             self._preview_box.setHtml(text)
+            # need to use singleshot to ensure we get the right height
+            QTimer.singleShot(
```
MBartkowiakSTFC, line 424: It is not clear to me why the rescaling of the text box has to be run via the QTimer. Could you explain this briefly?

BTW it works fine, and the GUI looks good.

ChiCheng45: Without the singleshot, self._preview_box.document().size().height() returned the previous height. The singleshot ensure it get updated and the correct height is returned.

MBartkowiakSTFC (approved):
Thanks, looks good.

PR #1012, opened 2025-10-07
Fix results preview box size update

ChiCheng45:
**Description of work**
The preview box size didn't update correctly when the trajectory was loaded with an analysis job already selected. This PR fixes this issue.

**To test**
Check that the results preview box size updates correctly.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
@@ -15,7 +15,6 @@
 #
 from __future__ import annotations
 
-import math
 import traceback
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC:
If we wanted to be pedantic, there is still one way to make it do something unintended. It the GUI window is narrow enough, the text box will get a horizontal scroll bar, and it will block part of the text:
Screenshot 2025-10-07 at 11 21 12

Do you think this needs to be addressed too?

ChiCheng45:
> u think this needs to be addressed too?



> If we wanted to be pedantic, there is still one way to make it do something unintended. It the GUI window is narrow enough, the text box will get a horizontal scroll bar, and it will block part of the text: Screenshot 2025-10-07 at 11 21 12
> 
> Do you think this needs to be addressed too?

Thanks for spotting that. I've decided to change the results preview to use Qlabel. A scroll bar shouldn't appear anymore. Can you test on mac?

MBartkowiakSTFC:
Screenshot 2025-10-07 at 13 42 14

It fixed the problem with the scroll bars, but now the preview widget does not get cleaned up correctly and empty preview widgets accumulate in the action widget.

ChiCheng45:
> Screenshot 2025-10-07 at 13 42 14
> It fixed the problem with the scroll bars, but now the preview widget does not get cleaned up correctly and empty preview widgets accumulate in the action widget.

Thanks, should be fixed now.

MBartkowiakSTFC (approved):
I think this solved the problem. Looks good.

PR #1015, opened 2025-10-07
Add grouping to RMSF calculation

### LINKED ISSUE #1006: [ENHANCEMENT] Label molecules in Root Mean Square Fluctuation
At the moment the labels of RMSF analysis output are largely meaningless.

Here is an example of the analysis of a MOF loaded with gas molecules:

Image

Image

There is no way of telling which part of the system produced each point of the results.

It would be good to add partial datasets, grouped by element or by molecule label.
Additionally, molecule grouping analysis should output the indices of individual atoms that contributed to each data point.

The indices of atoms composing a molecule can be saved as datasets, or could be dataset attributes.

**Additional context**
Do we need to give users some way to match a data point to an object in the trajectory 3D view?

### LINKED ISSUE #1014: [BUG] RMSF each molecule inconsistent with COM trajectory and RMSF each atom
**Description of the error**
Running the RMSF with each molecule setting should give you the RMSF of the center of mass of each molecule. However, if you run RMSF with each molecule setting and then run RMSF with each atom setting for a trajectory generated using the centre of masses conversion, you can get different results. 

With the DFTB trajectory, you get the same results because there are no jumps across the periodic boundaries. But with this water ethanol trajectory, which does have jumps across the periodic boundary, I get different results.

DFTB water trajectory RMSF each molecule and DFTB COM RMSF each atom
Image

Image

water ethanol trajectory RMSF each molecule and water ethanol COM RMSF each atom. Here the blue is the "each molecule" with the full trajectory.

Image

Image

ChiCheng45:
**Description of work**
RMSF is now changed so that only the results per atom are calculated. Now results, are grouped by element and can also be grouped by molecule in a way similar to all other MDANSE jobs. Indexes of each atom are maintained after grouping. Documentation has been updated.

Previously, each molecule would calculate the COM for each molecule and use that to calculate the RMSF. This is removed so that user should run COM trajectory job then the RMSF calculation to get this as for all other jobs in MDANSE.

Here are the results for a water ethanol system with atom and molecule grouping.

image

Here are the results for ethanols elements. I use the X offset in the plotter to separate the elements.

image

Here is the RMSF for all ethanols elements and all waters elements.

image

**Fixes**
Fixes #1014
Closes #1006

**To test**
Test the RMSF calculation with atom and molecule grouping.


ChiCheng45:
@MBartkowiakSTFC can you test this out on the MOF structure you had in #1006. Let me know what you think.

MBartkowiakSTFC:
Here is the results with the new implementation of grouping. I overplotted the results grouped by molecule with those grouped by atom.

rmsf_test_new

And here are some magnified details:

rmsf_test_new_details

MBartkowiakSTFC (approved):
Looks good, I did not find any problems with the results.

PR #1019, opened 2025-10-08
Block jobs when transmutation cause weight with nan values to be used.

### LINKED ISSUE #1002: [BUG] Jobs can be started with invalid weights by transmuting atoms
**Description of the error**
Normally, the `WeigthsConfigurator` is supposed to produce an error if an atom in the trajectory has invalid values (e.g. `nan`) for the selected weights.

At the moment it is possible to pick `b_coherent` and transmute an atom to another element (e.g. `At`) with `nan` value of `b_coherent`.

The `Configurators` should detect the problem and the GUI should prevent the user from running the job.

The job can be started, and it fails during initialisation.

**Suggested fix**
Both atom selection and atom transmutation information should be made available to `WeightsConfigurator`, so the invalid weight values cannot be introduced using transmutation, and so they can be removed by adjusting atom selection.

**Additional details**
N/A

ChiCheng45:
**Description of work**
Fixed the weight configurator and widget so that jobs get blocked whenever the atom selection or transmutation causes weight with nan values to be used.

**Fixes**
Fixes #1002

**To test**
Check that the weight widgets blocks jobs from running when weight with nans values are selected. Should work with both atom selection and transmutation.


MBartkowiakSTFC (approved):

I was worried at first that removing the `cached_property` decorators could affect the performance of the code, but it seems that most of these properties are accessed once per analysis anyway.

ChiCheng45:
> I did not find any problems. Looks good.
> 
> I was worried at first that removing the `cached_property` decorators could affect the performance of the code, but it seems that most of these properties are accessed once per analysis anyway.

Yeah, I found that cached_property needed to be removed because most of them depended on the atom selection or transmutation, which may change after caching.

PR #1021, opened 2025-10-09
Delete unused code

MBartkowiakSTFC:
**Description of work**
There are still some files in the code that can be removed without any loss of functionality compared to the current main branch.

Some of them were prototypes of new solutions which were not continued later. Others are converters and analysis types which do not work are the moment and have been disabled.

**Fixes**
Removed the following:
- ImprovedASE converter,
- RigidBodyTrajectory, GlobalMotionFilteredTrajectory, UnfoldedTrajectory, CroppedTrajectory,
- OrderParameter and McStasVirtualInstrument,
- NeutronInstrument classes,
- StringWidget, InputDirectoryWidget, OutputDirectoryWidget (in the GUI).

**To test**
Check if you think that some of these could still be needed.


ChiCheng45:
What's the plan exactly? Are we going to remove all unused code or just some of it? Because I think there is more unused code that I'm unsure whether you'd want to delete or not.

MBartkowiakSTFC:
The other candidates for getting removed from the code would be `GeneralAutoCorrelationFunction` and `RotationAutocorrelation`. But for these two I still see a possibility of getting them enabled again.

The intention of this PR was to remove code based on a reasonable prediction of if we ever intend to work on the code again. If it has not been used, nobody is asking for it to work or we have other methods of dealing with the same problem, then it probably makes sense to remove it. For some of the analysis types I propose to remove, I found more papers describing the method than papers actually using it.

Still, this initial draft is based only on my opinion. I'm happy to remove more code, or keep some of the code if you think that we may still realistically use it.

Reasons to deleting things so far:
1. All the `NeutronInstrument` classes: we will probably use ResIns instead.
2. `GlobalMotionFilteredTrajectory` and `RigidBodyTrajectory`: they would need to be completely re-written, when I tried it out I had molecules suddenly flip by 180 degrees even for simple cases, and we have `TrajectoryFilter`.
3. `CroppedTrajectory`: we just use `TrajectoryEditor` instead.
4. `UnfoldedTrajectory`: if we want it, it should be an option in `TrajectoryEditor`.
5. `ImprovedASEConverter`: we now have working examples of `MDAnalysisConverter` and `MDTrajConverter` which do what this one was supposed to. If anything, it would make sense to extend `ASEConverter` to work like the other two (i.e. with optional extra input files).
6. `McStasVirtualInstrument`: we should just have a tool that turns the DISF results into a McStas sample definition, and leave the rest to the users.
7. `OrderParameter`: it does not work at the moment, the code needs re-writing, not sure who needs it. But we don't really have a replacement for it, so we could still try to get it back into shape.

ChiCheng45:
> The other candidates for getting removed from the code would be `GeneralAutoCorrelationFunction` and `RotationAutocorrelation`. But for these two I still see a possibility of getting them enabled again.
> 
> The intention of this PR was to remove code based on a reasonable prediction of if we ever intend to work on the code again. If it has not been used, nobody is asking for it to work or we have other methods of dealing with the same problem, then it probably makes sense to remove it. For some of the analysis types I propose to remove, I found more papers describing the method than papers actually using it.
> 
> Still, this initial draft is based only on my opinion. I'm happy to remove more code, or keep some of the code if you think that we may still realistically use it.
> 
> Reasons to deleting things so far:
> 
> 1. All the `NeutronInstrument` classes: we will probably use ResIns instead.
> 2. `GlobalMotionFilteredTrajectory` and `RigidBodyTrajectory`: they would need to be completely re-written, when I tried it out I had molecules suddenly flip by 180 degrees even for simple cases, and we have `TrajectoryFilter`.
> 3. `CroppedTrajectory`: we just use `TrajectoryEditor` instead.
> 4. `UnfoldedTrajectory`: if we want it, it should be an option in `TrajectoryEditor`.
> 5. `ImprovedASEConverter`: we now have working examples of `MDAnalysisConverter` and `MDTrajConverter` which do what this one was supposed to. If anything, it would make sense to extend `ASEConverter` to work like the other two (i.e. with optional extra input files).
> 6. `McStasVirtualInstrument`: we should just have a tool that turns the DISF results into a McStas sample definition, and leave the rest to the users.
> 7. `OrderParameter`: it does not work at the moment, the code needs re-writing, not sure who needs it. But we don't really have a replacement for it, so we could still try to get it back into shape.

Sure, but I was wondering about stuff like the code in

MDANSE_GUI/Src/MDANSE_GUI/DataViewModel/ActionsHolder.py
MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Dummy.py
MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/DummyWidget.py

which, as far as I can tell, are unused. I think there is more.

MBartkowiakSTFC:
> 
> Sure, but I was wondering about stuff like the code in
> 
> MDANSE_GUI/Src/MDANSE_GUI/DataViewModel/ActionsHolder.py MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Dummy.py MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/DummyWidget.py
> 
> which, as far as I can tell, are unused. I think there is more.

Yes, let's remove them too.

ChiCheng45:
I've deleted some more unused code, as well as the GACF job.

MDANSE/Tests/UnitTests/Analysis/test_mcstas.py
oerc0122, line 1: Suspect this whole file can go?

oerc0122 (approved):
Just one more to delete?

PR #1023, opened 2025-10-13
fix: Replace faulty separator

### LINKED ISSUE #986: [BUG] Wrong handling of output file names with commas
**Description of the error**
The following happens when running an analysis with output file name "some.output"
1. MDANSE creates a file named "some.mda",
2. MDANSE_GUI starts protecting "some.output.mda",
3. The analysis finishes without errors,
4. The GUI fails to load the results,
5. The protected file name "some.output.mda" is never released.

Due to this bug it is possible to have two analysis runs writing out to the same file name, by starting two separate runs with output names "some.output" and "some.thingelse", as they will both write into "some.mda".

RobBuchananCompPhys:
**Description of work**
Closes #986 
 
**Fixes**
A list of fixes.

**To test**

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import time
 import traceback
+from pathlib import Path
 from itertools import count
 from logging import Handler
 from logging.handlers import QueueListener
@@ -26,6 +27,7 @@ from typing import Any, NamedTuple
 from qtpy.QtCore import QMutex, QObject, Qt, QThread, QTimer, Signal, Slot
 from qtpy.QtGui import QStandardItem, QStandardItemModel
 
+from MDANSE.Framework.Formats import MDAFormat
 from MDANSE.Framework.Converters import Converter
 from MDANSE.Framework.Jobs.JobStatus import JobInfo, JobStates
 from MDANSE.MLogging import FMT, LOG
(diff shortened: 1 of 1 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
```diff
@@ -316,6 +318,21 @@ def __init__(self, parent: QObject | None = None):
     def reportError(self, err: str):
         LOG.error(err)
 
+    @staticmethod
+    def safe_mda_file_name(filename: str = "") -> str:
+        path = Path(filename)
+        name = path.name
+        parts = name.split('.')
```
oerc0122, line 325: What are you trying to do here? Are you trying to get [`Path.suffixes`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffixes)?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
```diff
@@ -316,6 +318,21 @@ def __init__(self, parent: QObject | None = None):
     def reportError(self, err: str):
         LOG.error(err)
 
+    @staticmethod
+    def safe_mda_file_name(filename: str = "") -> str:
```
oerc0122, line 322: Docstring

Do you actually want a default here? The default will be meaningless and fail or return a folder name which could be worse.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobHolder.py
```diff
@@ -316,6 +318,21 @@ def __init__(self, parent: QObject | None = None):
...
+        path = Path(filename)
+        name = path.name
+        parts = name.split('.')
+
+        if (len(parts) > 1):
+            if (parts[-1] == MDAFormat.extension.strip(".")):
+                name = "_".join(parts[0:-1])
+            else:
+                name = "_".join(parts)
+                name += MDAFormat.extension
```
oerc0122, line 332: I don't understand the logic here. If it ends with extension, remove the extension, otherwise add the extension?

oerc0122 (changes requested):

MBartkowiakSTFC:
At the moment, if I name my output `this.will.work.now`, the analysis job ends up writing to `this_will_work_now.mda`. This is better than what the main branch does (since the main branch would create an output called `this.mda`). However, there is still the problem that the GUI still allows me to re-use the name `this.will.work.now`.

Ideally, we should find a way to give the user what they asked for. My preferred outcome would be to end up with a file `this.will.work.now.mda` without replacing any characters, and this would probably make it easier to keep track of which file name we are still allowed to use.

MBartkowiakSTFC:
This will be done in #1071

Issue #724, opened 2025-03-25
[ENHANCEMENT] Migrate unit handling to `pint`


oerc0122:
**Is your feature request related to a problem? Please describe.**
Unit handling is currently handled by custom (somewhat restricted) unit lists.

By migrating to [`pint`](https://pint.readthedocs.io/en/stable/) we would gain acces to a rich set of units with dimensionality checks and other benefits.

In the case of the `json` "registries", it would be possible to add an extended `json` parser to handle a `units` element in the dictionaries and return the appropriate `Quantity` objects.

Due to `pint`'s issues with spectroscopic units (i.e. the conflict between `nm` / `cm -1` and `eV` being fundamentally dimensionally different), we could use something similar (exactly) to [Euphonic's spectroscopy unit registry](https://github.com/pace-neutrons/Euphonic/blob/master/euphonic/ureg.py) (with appropriate acknowledgement).

**Describe the solution you'd like**
`pint` as the main unit handler.

**Describe alternatives you've considered**
N/A

**Additional context**
c.f. #698

PR #1024, opened 2025-10-13
Add tab for viewing properties of `Trajectory`s in GUI

### LINKED ISSUE #962: [ENHANCEMENT] Better access to trajectory contents in the GUI
Trajectories are currently able to store atom property information. It is possible to check these values in the CLI, but not in the GUI.

An optional atom property viewer could be added as a separate dialog. At the moment we have a context menu in the trajectory tab which allows us to delete trajectories. We could add an entry there for viewing trajectory contents.

We could pack more and more information into `TrajectoryInfo` in the trajectory tab. But then we will need to structure the information and add a search option.

**Additional context**
N/A

oerc0122:
**Description of work**
Add a tab to the trajectory tab options which displays properties of trajectory.

**Fixes**
Fixes #962 

- Makes `MolecularViewer` not collapsible.
- Makes `Controls` tab expandable.
- `PropertyViewer` is made inactive when tab is deselected (but not when hidden...)
- `PropertyViewer` only updates when frame changes if (potentially) frame-dependent property

**To test**
Open trajectories and look at the properties.

### DIFF
--- a/MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
+++ b/MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
@@ -207,7 +207,7 @@ class H5MDTrajectory(TrajectoryFile):
         """Mapping of property labels to units."""
         return ChainMap(
             self._units,
-            {"b_incoherent": "Ang", "b_coherent": "Ang"},
+            {"b_incoherent": "ang", "b_coherent": "ang"},
             ATOMS_DATABASE.units,
         )
 
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
@@ -38,6 +38,7 @@ from qtpy.QtWidgets import (
 )
 
 from MDANSE_GUI.MolecularViewer.MolecularViewer import MolecularViewer
+from MDANSE_GUI.MolecularViewer.PropertyWidget import PropertyWidget
 from MDANSE_GUI.MolecularViewer.TraceWidget import TraceWidget
 from MDANSE_GUI.Tabs.Views.Delegates import ColourPicker, RadiusSpinBox
 
@@ -189,6 +190,7 @@ class ViewerControls(QWidget):
         layout = QVBoxLayout(base)
         base.setLayout(layout)
         self._side_base.addTab(base, "Controls")
+
         # colour changes
         wrapper0 = QGroupBox("Colour settings", base)
         layout0 = QHBoxLayout(wrapper0)
@@ -199,6 +201,7 @@ class ViewerControls(QWidget):
         layout0.addWidget(bkg_button)
         layout0.addWidget(proj_button)
         layout.addWidget(wrapper0)
+
         # the table of chemical elements
         wrapper1 = QGroupBox("Atom properties", base)
         layout1 = QHBoxLayout(wrapper1)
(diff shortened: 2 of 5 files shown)


MBartkowiakSTFC:
One thing I spotted that still needs fixing:

The second atom gets renamed if you view atom charges. The first frame shows this:

Screenshot 2025-10-14 at 09 51 33

If you then change frames, you get:

Screenshot 2025-10-14 at 09 51 39

ChiCheng45 (commented):
A few requests.

Unit cell vertical header should be a, b, c.

Image

The raw_velocitiy and raw_coordinates start index starts from zero, but I think this needs to be for everything else to be consistent.

Image

There is a blank selection in the atoms.

Image

Can we sort the drop-down?

Image

Add units to vertical header? E.g. would be nice to see xs_absorption / nm2.

Image

Url and other string properties show a -1.

Image

Otherwise, looking good.

oerc0122:
Should all be fixed with the exception of `str` properties being `-1.0`, c.f #1027

MBartkowiakSTFC:
This is looking good, but now we are uncovering another layer of mess which I created. (Sorry.) That is, the units in the built-in atom database in the trajectory file do not match the units in the central database.

When I first added the atom database to the trajectory, there was no unit information in the _central_ database, and the choice of units we inherited in MDANSE was quite arbitrary. (Normally neutron scattering length is in fm, the MDANSE internal unit of length is nm, so the MDANE database stored neutron scattering length in Angstrom.)

When I replaced the b_coh and b_inc values in the central database with those in fm, I added the unit information to the central database to make sure that it is clear what the values are. But since the old trajectories had no unit information AND the values in the old units, I decided to keep using the old units internally. While this made sure that we could reproduce the old results and use the old trajectory files from before the change, in the long run this is creating confusion.

The current situation is this:
1. The `mdanse element` CLI command shows the unit of b_coh and b_inc as `fm` for the central database, but has all the units shows as `None` for the values from trajectory file. In the case of oxygen, we have the output of
`mdanse element O`
with the line
`b_coherent                     (5.803-5.28364849833148e-08j)        fm`
But if we read the value from a trajectory file using `mdanse element -t some_trajectory.mdt`, it will show
`b_coherent                            5.8030000000000005e-05      None`
which is also not right, because for the new files the unit of b_coh is saved, and it says `ang`.
2. In the GUI, the atom property preview will show the value in the trajectory as 5.803e-05 fm, since fm is the unit in the central database.

Ideally, the GUI should check if there is unit information in the trajectory, and use the unit definition from there. But you are welcome to suggest a better long-term solution.

oerc0122:
I don't really see a simple solution to conflicting information, but the most obvious answer is we warn on "old" trajectories, 2.0.0 is API breaking so we're allowed to do that. 

In the long term #724

MBartkowiakSTFC:
We are not likely to get _conflicting_ information. There are two easy scenarios:
1. Values from central database and units from central database -> straightforward.
2. Values from trajectory and units from trajectory -> either change the unit label or recalculate to match the units in the central database.

And most likely we have one more tricky scenario, where we have values in the trajectory and no unit information provided. But then it is easy to assume that the units are the old ones and NOT those in the central database. To be safe, we could apply some simple heuristics, like `if abs(b_coh) < 1e-2: unit = 'ang' else unit= 'fm'`...

But ideally we would have had just one easy scenario and nothing else.

MBartkowiakSTFC:
Could this be a working solution?
1. We agree that all the atom properties should be shown in the GUI in the same units as in the central database.
2. If the trajectory database contains the unit information, we compare units and convert where necessary.
3. If the trajectory has no unit information, we assume that the old units are being used, convert the values accordingly AND produce a warning about the unit information not being present in the trajectory file.

MBartkowiakSTFC (approved):
I think this works. I tried many trajectories, old and new, and I got values of neutron scattering length that matched the displayed units.

Issue #946, opened 2025-09-02
[ENHANCEMENT] Unify writing of dicts


oerc0122:
**Is your feature request related to a problem? Please describe.**
There are lots of instances where dict pairs are written. Could add standard function to `IOUtils` to dump dicts nicely reducing code duplication. 

Places:
- MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
- MDANSE/Src/MDANSE/Framework/Jobs/IJob.py

PR #1025, opened 2025-10-14
Update the MDANSE save scripts

ChiCheng45:
**Description of work**
Updated the save scripts so that they are formatted with black, long JSON strings are saved as multiline strings, and are formatted. Basically, this should stop the MDANSE save script from writing very long lines.

Updated the MDANSE trajectory configurator label. Fixed the trajectory editor save script.

**Fixes**
Trajectory editor saved scripts used a numpy array, but did not include the numpy import in the script. Updated the unit cell configurator so it uses a list of floats. 

**To test**
Save scripts for trajectory and analysis jobs and run them. Test the save script for the trajectory editor.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -304,8 +304,8 @@ class IJob(Configurable, metaclass=SubclassFactory):
             if (
                 isinstance(v, str)
                 and len(str_v) > 40
-                and str_v[0] == "{"
-                and str_v[-1] == "}"
+                and str_v.startswith("{")
+                and str_v.endswith("}")
             ):
                 # if it's a long json string then try to make a multiline
                 # string and format it
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -324,14 +324,16 @@ class IJob(Configurable, metaclass=SubclassFactory):
             )
 
         with open(jobFile, "w") as f:
-            f.write(RUNSCRIPT.format(
-                executable=sys.executable,
-                import_line=cls.runscript_import_line,
-                param_str=param_str,
-                parent=cls.runscript_import_line.split(" ")[-1],
-                var_name=cls.__name__.lower(),
-                job_name=cls.__name__,
-            ))
+            f.write(
+                RUNSCRIPT.format(
+                    executable=sys.executable,
+                    import_line=cls.runscript_import_line,
+                    param_str=param_str,
+                    parent=cls.runscript_import_line.split(" ")[-1],
+                    var_name=cls.__name__.lower(),
+                    job_name=cls.__name__,
+                )
+            )
 
         os.chmod(jobFile, stat.S_IRWXU)
 
--- a/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
@@ -320,14 +320,16 @@ class IJob(Configurable, metaclass=SubclassFactory):
             )
 
         with open(jobFile, "w") as f:
-            f.write(RUNSCRIPT.format(
-                executable=sys.executable,
-                import_line=cls.runscript_import_line,
-                param_str=param_str,
-                parent=cls.runscript_import_line.split(" ")[-1],
-                var_name=cls.__name__.lower(),
-                job_name=cls.__name__,
-            ))
+            f.write(
+                RUNSCRIPT.format(
+                    executable=sys.executable,
+                    import_line=cls.runscript_import_line,
+                    param_str=param_str,
+                    parent=cls.runscript_import_line.split(" ")[-1],
+                    var_name=cls.__name__.lower(),
+                    job_name=cls.__name__,
+                )
+            )
 
         os.chmod(jobFile, stat.S_IRWXU)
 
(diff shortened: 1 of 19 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -32,6 +33,7 @@
 from pathlib import Path
 from typing import Any
 
+import black
```
oerc0122: I don't really think we want to necessarily `black` format the outputs, it's a little overkill and an extra dependency. 

What I would recommend instead is making the file into a separate template file and reading that in and re-dumping it. Formatting that will be easier.

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -296,22 +298,39 @@ def save(
...
+                # if it's a long json string then try to make a multiline
+                # string and format it
+                try:
+                    json_data = json.loads(str_v)
+                    repr_v = '"""' + json.dumps(json_data, indent=4) + '"""'
+                    repr_v = repr_v.replace("\n", "\n    ")
+                except json.decoder.JSONDecodeError:
+                    repr_v = repr(v)
+            else:
+                repr_v = repr(v)
```
oerc0122: Would prefer to see this drawn out into a function. Probably in `IO.IOUtils`

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -296,22 +298,39 @@ def save(
...
-            for k, (v, label) in parameters.items()
-        )
+        param_str = ""
+        for k, (v, label) in parameters.items():
+            str_v = str(v)
+            if (
+                isinstance(v, str)
+                and len(str_v) > 40
+                and str_v[0] == "{"
+                and str_v[-1] == "}"
```
```suggestion
                and str_v.startswith("{")
                and str_v.endswith("}")
```

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -296,22 +298,39 @@ def save(
...
+            ):
+                # if it's a long json string then try to make a multiline
+                # string and format it
+                try:
+                    json_data = json.loads(str_v)
+                    repr_v = '"""' + json.dumps(json_data, indent=4) + '"""'
+                    repr_v = repr_v.replace("\n", "\n    ")
```
oerc0122: What is this replace doing?

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -296,22 +298,39 @@ def save(
...
+                isinstance(v, str)
+                and len(str_v) > 40
+                and str_v[0] == "{"
+                and str_v[-1] == "}"
+            ):
+                # if it's a long json string then try to make a multiline
+                # string and format it
+                try:
+                    json_data = json.loads(str_v)
+                    repr_v = '"""' + json.dumps(json_data, indent=4) + '"""'
```
```suggestion
                    repr_v = f'"""{json.dumps(json_data, indent=4)}"""'
```

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
ChiCheng45: How would you do that exactly?

oerc0122: Saving the template as a text file and if it's changed substituting in some values and running `ruff format`.

I think that except for the JSON strings nothing will be terribly long and disrupt the formatting.

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
ChiCheng45: We need an indent after the first new line because the parameter dict block is indented.

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
ChiCheng45: Sure, but doesn't that mean we swap the dependency here of black with ruff? Seems more complicated because I will need to call subprocess ruff?

oerc0122: I mean don't run it every time we dump. I think that's overkill. 

If it's slightly poorly formatted then it's easy enough for a user to do that or deal with it.

MBartkowiakSTFC: Just to add an opinion to the discussion: having black as an extra dependency did not really bother me much, and it did not slow us down either.

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
ChiCheng45: Can you explain? I can see some point in moving the runscript template and runscript writing to IO.IOUtils, but not so much the above lines of code, as I don't think it will be used anywhere else except here.

oerc0122: I think we have a couple of places where we need to do this (#946). Although I've just remembered 
https://docs.python.org/3/library/pprint.html#pprint.pformat
which might be a better way to do it.

ChiCheng45: I use pprint for the dict, list, and tuple datatypes. How exactly do you want to see the move of the above to IO.IOUtils? Do you just want a function there because I don't know how the above could be generalized without overcomplicating the code, and then used in MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py.

oerc0122: As discussed, I agree that this shouldn't go into `IO.IOUtils`, I do still think the code should be pulled out into a module-private function such as `_format_params`.

Then I'm fairly sure this is ready to go.

oerc0122 (approved):
Looks good. Sorry for the hassle.

PR #1027, opened 2025-10-14
Fix getting string properties from newer Trajectories

oerc0122:
**Description of work**
Currently string properties from newer Trajectory files return `-1.0` which is not right.

**Fixes**
If a string property is requested, revert to `ATOMS_DATABASE`.

**To test**
Ensure that getting string properties (e.g. `'url'`) from new trajectories works.


MBartkowiakSTFC (approved):
I checked that it works. For an atom that exists in the trajectory and NOT in the central database, the string properties are now shown as "None", which makes more sense than "-1.0". Looks good.

PR #1028, opened 2025-10-14
Report trajectory loading failure in the GUI

### LINKED ISSUE #1022: [BUG] Loading invalid file stalls "loading"
**Description of the error**
If you attempt to load an invalid file the task will fail, but the menu will continue to read "loading ..."

Either the trajectory should be deleted from the list or the label should change to "Failed"

Perpetual "loading ..."

**Suggested fix**
N/A

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Gives more feedback to the user if an invalid file has been picked as a trajectory.

closes #1022

**Fixes**
- invalid items in TrajectoryModel are now disabled,
- replaced the True/False 'is_ready' status of trajectories in the model with an enum,
- files that failed to load are labelled in the trajectory list and can be deleted,
- a QProxyStyle prevents the combo box in JobTab from resizing to the length of the longest text string.

**To test**
Load all kinds of files as trajectories. See which failed and which can be used. Please check if the `Action` in JobTab updates correctly when switching from one trajectory to another.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
@@ -15,15 +15,13 @@
 #
 from __future__ import annotations
 
-import os
 from functools import partial
-from pathlib import PurePath
 
-from qtpy.QtCore import Slot
-from qtpy.QtWidgets import QComboBox, QLabel, QWidget
+from qtpy.QtCore import Qt, Slot
+from qtpy.QtGui import QFontMetrics
+from qtpy.QtWidgets import QApplication, QComboBox, QLabel, QProxyStyle, QStyle, QWidget
 
 from MDANSE.MLogging import LOG
-from MDANSE_GUI.InputWidgets.MoleculeWidget import MoleculeWidget
 from MDANSE_GUI.Session.Session import Session
 from MDANSE_GUI.Tabs.GeneralTab import GeneralTab
 from MDANSE_GUI.Tabs.Layouts.MultiPanel import MultiPanel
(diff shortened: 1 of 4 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -102,6 +131,7 @@ def launch_loader(
             None, filename, index, display_item=display_item, target_label=target_label
         )
         thread.results.connect(self.accept_results)
+        thread.failure.connect(self.accept_failure)
```
oerc0122, line 134: A mantra for MDANSE?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -142,6 +181,7 @@ def removeRow(self, row: int, parent: QModelIndex | None = None) -> bool:
         retcode = super().removeRow(row, parent)
         instance = self._trajectory_instances.pop(node_number)
         filename = self._trajectory_paths.pop(node_number)
+        _ = self._trajectory_status.pop(node_number)
```
```suggestion
        self._trajectory_status.pop(node_number)
```
Don't need to return, although a perhaps more canonical way would be:

```suggestion
        del self._trajectory_status[node_number]
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
```diff
@@ -66,6 +78,10 @@ def __init__(
         self._instrument_index = -1
         self._trajectory_combo = QComboBox()
         self._trajectory_combo.setEditable(False)
+        self._combo_style = ComboStyle(
+            QApplication.style().name(), metrics=self._trajectory_combo.fontMetrics()
+        )
```
oerc0122: Does this need to be on the `JobTab` object?

oerc0122 (approved):

PR #1029, opened 2025-10-15
Improve the atom trace

### LINKED ISSUE #932: Atom trace plot needs to be improved what should we do?
Currently, the atom trace plots the positions of the atom's trajectory. But from my testing, it is difficult to make it plot something that could be used in a publication.

Image

Here are some suggestions I have come up with.

Here, I make a small change where I interpolate the trajectory; this can clean up the current implementation a bit.

Image

Here is another where I place Gaussian functions on each coordinate along the trajectory and then evaluate the values with a meshgrid. I plot isovalues of 20, 40, and 60% for two hydrogen atoms in the dftb water trajectory.

Image

This is sort of like a probability density; the higher values of the isovalue show where the atom spent the most time.

Here is one that is useful if the user is only interested in the trace, not the probability density.

Image

Here, I use a vtk tubefilter and also interpolate the trajectory.

Here are some atom trace settings we might want to add.
- Run atom trace over specific frames only.
- Allow for multiple atoms to be selected.

ChiCheng45:
**Description of work**
Changed the atom trace so it is now calculated using gaussians and we now plot isovalue rather than the percentile as it seems to make better plots.

Screenshot 2025-10-15 at 17 31 18

There are two new options and one updated.

- Trajectory samples: how many times the trajectory is sampled. With make the atom trace will be more continuous, useful when the trajectory used larger time steps. With lower values atom trace will be computed a bit more quickly and use less memory.
- Position smearing: Smears the gaussians out, with larger smearing atom trace will appear more like a probability density while with lower it will be more like a trace of the trajectory.
- Grid step now replaces the fine sampling but allows more precise specification of the grid that is sampled.

Here is an example of the atom trace of oxygen with trace percent of 0.30, 0.50, 0.70, and 0.90. 
Screenshot 2025-10-15 at 17 38 05

Now I reduce the smearing to from 1 to 0.1, increase the grid step to 0.005 and plot the trace isovalue to 0.10.
Screenshot 2025-10-15 at 17 41 09

Here is something in-between using a different atom with smearing of 0.50, grid step of 0.01 and the trace plotted at 0.10.
Screenshot 2025-10-15 at 17 47 48

Here is an example with a longer trajectory. 0.10 smearing, 0.01 grid step, and 0.01 isovalue.

image

here is the same atom with 1.00 smearing, 0.02 grid step and 0.05 isovalue

image

**Fixes**
Closes #932 although it may still be worth adding the tube filter after this gets merged.

**To test**
Test the atom trace with a few systems and check it works as expected.


MBartkowiakSTFC (approved):
This is certainly an improvement over the current atom trace functionality.

With the current default values I frequently get the impression that something is wrong, because the atom I selected is outside of the isosurface (for the first frame). But the result actually makes sense, and I can confirm that everything is fine by changing the parameters.

Overall, I think that this PR is good to go, but it would be good to follow it up with the following work:
1. Add to the documentation the explanation you already included in this PR description,
2. Could we have some pre-defined parameter sets that the users could pick? So people can switch quickly between the exacty trace and probability plots? (And then they could still customise by manually adjusting the settings.)

ChiCheng45:
> This is certainly an improvement over the current atom trace functionality.
> 
> With the current default values I frequently get the impression that something is wrong, because the atom I selected is outside of the isosurface (for the first frame). But the result actually makes sense, and I can confirm that everything is fine by changing the parameters.
> 
> Overall, I think that this PR is good to go, but it would be good to follow it up with the following work:
> 
> 1. Add to the documentation the explanation you already included in this PR description,
> 2. Could we have some pre-defined parameter sets that the users could pick? So people can switch quickly between the exacty trace and probability plots? (And then they could still customise by manually adjusting the settings.)

Great thanks. Yes, these are some good ideas. I have created #1033 to keep track of it.

PR #1035, opened 2025-10-20
Extract properties from `Trajectory`s for `unit_info` more consistently

oerc0122:
**Description of work**
- Add a `units` property to `Trajectory`, `MdanseTrajectory` and `H5MDTrajectory` which try units defined in file, followed by `ATOMS_DATABASE` (which is a `defaultdict` and thus is total).
- General clean up of atom_info

**Fixes**
Fixes needed for #1024 

**To test**
Standard tests should pass as should `mdanse element -t ... []`

### DIFF
--- a/MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
+++ b/MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
@@ -15,6 +15,8 @@
 #
 from __future__ import annotations
 
+from collections import ChainMap, defaultdict
+from collections.abc import Iterable, Mapping
 from pathlib import Path
 
 import h5py
@@ -502,6 +504,23 @@ class MdanseTrajectory(TrajectoryFile):
         return [
             key for key in self._h5_file["/atom_database"] if "property_" not in key
         ]
+
+    @property
+    def units(self) -> Mapping[str, str]:
+        """Mapping of property labels to units."""
+        if "atom_database/property_units" not in self._h5_file:
+            return ATOMS_DATABASE.units
+
+        return ChainMap(
+            dict(
+                zip(
+                    self._h5_file["/atom_database/property_labels"].asstr(),
+                    self._h5_file["/atom_database/property_units"].asstr(),
+                    strict=True,
+                ),
+            ),
+            ATOMS_DATABASE.units,
+        )
 
     def properties(self) -> list[str]:
         """Return the list of all the properties in the trajectory's database.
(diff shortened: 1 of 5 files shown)


MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
```diff
@@ -500,6 +501,23 @@ def atoms_in_database(self) -> list[str]:
             key for key in self._h5_file["/atom_database"] if "property_" not in key
         ]
 
+    @property
+    def units(self) -> Mapping[str, str]:
+        """Mapping of property labels to units."""
+        if "atom_database" not in self._h5_file:
```
MBartkowiakSTFC: Here it would make sense to be more specific and do
`if "atom_database/property_units" not in self._h5_file:`,
just in case somebody was using a trajectory from the beta version, where the atom database already existed, but it did not have its own dataset for units (because the units were the same as in the central database).

MBartkowiakSTFC:
At the moment, for the **central database** I get values like
`b_coh 6.646 fm`
and for the newest trajectories I get
`b_coh 6.646e-05 ang`.
For the in-between trajectories (which have the values but do not have the units) I get
`b_coh 6.646e-05 fm`

A good example in our tests is `Converted/four_molecules.mdt`.

Was this supposed to happen, and is the plan to add the unit conversion in another PR?

oerc0122:
> At the moment, for the **central database** I get values like `b_coh 6.646 fm` and for the newest trajectories I get `b_coh 6.646e-05 ang`. For the in-between trajectories (which have the values but do not have the units) I get `b_coh 6.646e-05 fm`
> 
> A good example in our tests is `Converted/four_molecules.mdt`.
> 
> Was this supposed to happen, and is the plan to add the unit conversion in another PR?

I think the problem is with no knowing what units things are in. It believes that if the units are missing, they're in the units from the ATOMS_DATABASE, which I don't think is unreasonable. Do we think people are likely to have "in-between" trajectories?

MBartkowiakSTFC:
> 
> I think the problem is with no knowing what units things are in. It believes that if the units are missing, they're in the units from the ATOMS_DATABASE, which I don't think is unreasonable. Do we think people are likely to have "in-between" trajectories?

We started writing unit information explicitly at the same time as we first switched from Angstroms to femtometres for b_coh and b_inc. So, it should normally be safe to assume that if no unit information is provided, then the b_coh and b_inc are both in Angstroms.

To the best of my knowledge, there could be trajectories that have no unit information and have neutron scattering lengths in Angstroms, but we never produced trajectories that have no unit information and have neutron scattering lengths in femtometres.

MBartkowiakSTFC (approved):
Looks good, the values and units match now.

PR #1037, opened 2025-10-21
Update molecule viewer and SMARTs pattern matching

### LINKED ISSUE #286: [ENHANCEMENT] Add bonding and atom information to MDANSE

### LINKED ISSUE #1030: [BUG] Molecule preview window can be opened multiple times.
**Description of the error**
The molecule preview window can be opened multiple times. Update it so the preview closes when the button is pressed again.

**Additional details**
Image

ChiCheng45:
**Description of work**

- The molecule viewer closes when the button is pressed more than once, in line with other helper dialogs. 
- Bond types are determined using the rdDetermineBonds.DetermineBondOrders in rdkit.
- Molecule viewer now shows the correct bond types.
- SMARTs pattern matching is now working without the need for ~ bond wildcards. SMARTS pattern matching ignore any bonds to dummy atoms.

**Molecule viewer**
Before
image

After
image

Unfortunately rdDetermineBonds.DetermineBondOrders has some issues dealing with large molecules. Currently, we skip large molecules to avoid long load times. I think there could be a way to reduce the computation time for larger molecules. I have reached out to the RDKit teams to see what they think [#7922](https://github.com/rdkit/rdkit/issues/7922). 

A few things I've noticed which I have left unchanged from the protos (screenshot are from protos).

- Sometimes atoms may be grouped into a molecule but are not bonded.
Screenshot 2025-10-22 at 09 56 04

- Bonds to dummy atoms will be drawn.
Screenshot 2025-10-22 at 09 56 56

**Fixes**
Closes #286
Fixes #1030
SMARTS pattern should now work with small/medium molecules.
Molecule viewer should display the correct bond types for small/medium molecules. 

**To test**
Check that the molecule viewer no longer duplicates the window when clicked multiple times. Try out different pattern matches in the atom selection.

Take a look at https://www.daylight.com/dayhtml_tutorials/languages/smarts/smarts_examples.html and let me know if there are any functional groups that would be worth adding.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -69,7 +69,7 @@ class ChemicalSystem:
         # the rdkit atom indexes to the ones in MDANSE
         self._rdkit_map = {}
         self._rdkit_map_inv = {}
-        self._rdkit_dummy_atms = []
+        self._rdkit_dummy_atms = set()
 
         self._unique_elements = set()
 
@@ -104,11 +104,11 @@ class ChemicalSystem:
         if name_list is not None:
             self._atom_names = [str(x) for x in name_list]
 
-        self._rdkit_dummy_atms = [
+        self._rdkit_dummy_atms = {
             atom.GetIdx()
             for atom in self.rdkit_mol.GetAtoms()
             if atom.GetAtomicNum() == 0
-        ]
+        }
 
     def add_atom(self, atm_num: int) -> int:
         rdkit_atm = Chem.Atom(atm_num) if atm_num is not None else Chem.Atom(0)
@@ -128,7 +128,9 @@ class ChemicalSystem:
                 # dummy atoms.
                 self.rdkit_mol.AddBond(i, j, Chem.rdchem.BondType.UNSPECIFIED)
 
-    def set_bond_orders(self, coords: np.ndarray, *, max_iters: int = 1000, max_natms: int = 100):
+    def set_bond_orders(
+        self, coords: np.ndarray, *, max_iters: int = 1000, max_natms: int = 100
+    ):
         """Set the bond types for the bonds in the rdkit_mol.
 
         Parameters
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -172,7 +172,9 @@ class ChemicalSystem:
         # deal with them.
         for cluster_name in self._clusters:
             for idx, cluster in enumerate(self._clusters[cluster_name]):
-                cluster_no_dummes = [i for i in cluster if i not in self._rdkit_dummy_atms]
+                cluster_no_dummes = [
+                    i for i in cluster if i not in self._rdkit_dummy_atms
+                ]
 
                 atm_nums = [atom_numbers[i] for i in cluster_no_dummes]
                 unsupported = [
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -26,8 +26,11 @@ import networkx as nx
 import numpy as np
 from more_itertools import padded
 from rdkit import Chem
+from rdkit.Chem import rdDetermineBonds
+from rdkit.Geometry import Point3D
 
 from MDANSE.Chemistry import ATOMS_DATABASE
+from MDANSE.Framework.Units import measure
 from MDANSE.MLogging import LOG
 
 
@@ -105,6 +108,25 @@ class ChemicalSystem:
             self.rdkit_mol.AddBond(
                 int(pair[0]), int(pair[1]), Chem.rdchem.BondType.UNSPECIFIED
             )
+
+    def set_bond_orders(self, coords: np.ndarray):
+        """Set the bond types for the bonds in the rdkit_mol.
+
+        Parameters
+        ----------
+        coords : np.ndarray
+            The coordinates of the system to determine the bond types
+            from.
+        """
+        coord_ang = coords * measure(1.0, "nm").toval("ang")
+        conf = Chem.Conformer(coord_ang.shape[0])
+        for i, (x, y, z) in enumerate(coord_ang):
+            conf.SetAtomPosition(i, Point3D(x, y, z))
+        self.rdkit_mol.AddConformer(conf)
+        try:
+            rdDetermineBonds.DetermineBondOrders(self.rdkit_mol, charge=0)
+        except Exception as e:
+            LOG.error(f"Error determining bond order: {e}")
 
     def add_labels(self, label_dict: dict[str, list[int]]):
         for key, item in label_dict.items():
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -25,7 +25,7 @@ from typing import Any, SupportsInt
 import h5py
 import networkx as nx
 import numpy as np
-from more_itertools import padded
+from more_itertools import padded, quantify
 from rdkit import Chem
 from rdkit.Chem import rdDetermineBonds
 from rdkit.Geometry import Point3D
@@ -194,10 +194,12 @@ class ChemicalSystem:
                     continue
 
                 atm_num_valences = []
-                for i, j in enumerate(cluster_no_dummies):
+                for atm_num, j in zip(atm_nums, cluster_no_dummies, strict=False):
                     num_bonds = self.rdkit_mol.GetAtomWithIdx(j).GetDegree()
                     atm_num_valences.append(
-                        len([k for k in atomic_valences[atm_nums[i]] if num_bonds <= k])
+                        quantify(
+                            atomic_valences[atm_num], pred=lambda k: num_bonds <= k
+                        )
                     )
 
                 num_valences_combo = np.prod(atm_num_valences, dtype=object)
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -173,11 +173,11 @@ class ChemicalSystem:
         # deal with them.
         for cluster_name in self._clusters:
             for idx, cluster in enumerate(self._clusters[cluster_name]):
-                cluster_no_dummes = [
+                cluster_no_dummies = [
                     i for i in cluster if i not in self._rdkit_dummy_atms
                 ]
 
-                atm_nums = [atom_numbers[i] for i in cluster_no_dummes]
+                atm_nums = [atom_numbers[i] for i in cluster_no_dummies]
                 unsupported = [
                     element_list[idx]
                     for idx, atom in enumerate(atm_nums)
@@ -194,7 +194,7 @@ class ChemicalSystem:
                     continue
 
                 atm_num_valences = []
-                for i, j in enumerate(cluster_no_dummes):
+                for i, j in enumerate(cluster_no_dummies):
                     num_bonds = self.rdkit_mol.GetAtomWithIdx(j).GetDegree()
                     atm_num_valences.append(
                         len([k for k in atomic_valences[atm_nums[i]] if num_bonds <= k])
@@ -214,11 +214,11 @@ class ChemicalSystem:
                 mapping = {}
 
                 submolecule = Chem.RWMol()
-                for i in cluster_no_dummes:
+                for i in cluster_no_dummies:
                     new_idx = submolecule.AddAtom(self.rdkit_mol.GetAtomWithIdx(i))
                     mapping[i] = new_idx
                 bond_idxs = []
-                for i, j in it.combinations(cluster_no_dummes, 2):
+                for i, j in it.combinations(cluster_no_dummies, 2):
                     bond = self.rdkit_mol.GetBondBetweenAtoms(i, j)
                     if bond is None:
                         continue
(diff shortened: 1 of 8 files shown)


MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -93,6 +104,11 @@ def initialise_atoms(
         if name_list is not None:
             self._atom_names = [str(x) for x in name_list]
 
+        self._rdkit_dummy_atms = []
+        for atom in self.rdkit_mol.GetAtoms():
+            if atom.GetAtomicNum() == 0:
+                self._rdkit_dummy_atms.append(atom.GetIdx())
```
oerc0122:
```suggestion
        self._rdkit_dummy_atms = [
            atom.GetIdx()
            for atom in self.rdkit_mol.GetAtoms()
            if atom.GetAtomicNum() == 0
        ]
```

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -102,9 +118,153 @@ def add_atom(self, atm_num: int) -> int:
...
+            from.
+        max_valence_combo: int
+            The maximum number of valence combinations.
+        """
+
+        # number of possible valences that an atom could have e.g. nitrogen
+        # has two possible valences 3 or 4 when determining the bond types,
+        # DetermineBondOrders will check all possible combinations of valences
+        # see https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/DetermineBonds/DetermineBonds.cpp
+        atomic_valences = {
```
oerc0122: Does this want to go somewhere atoms database-y?

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -102,9 +118,153 @@ def add_atom(self, atm_num: int) -> int:
...
+
+        # Calling rdDetermineBonds.DetermineBondOrders for a large system
+        # is quite computationally expensive. It's better to call this
+        # function for unique molecules and then copy over all the bond
+        # types for all others. We also need to remove dummy atom before
+        # using rdDetermineBonds.DetermineBondOrders because it can't
+        # deal with them.
+        for cluster_name in self._clusters:
+            for idx, cluster in enumerate(self._clusters[cluster_name]):
+                cluster_no_dummes = [
```
oerc0122:
```suggestion
                cluster_no_dummies = [
```

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -102,9 +118,153 @@ def add_atom(self, atm_num: int) -> int:
...
+                    i for i in cluster if i not in self._rdkit_dummy_atms
+                ]
+
+                atm_nums = [atom_numbers[i] for i in cluster_no_dummes]
+                unsupported = [
+                    element_list[i]
+                    for i, j in enumerate(atm_nums)
+                    if j not in atomic_valences
+                ]
```
oerc0122:
```suggestion
                unsupported = [
                    element_list[idx]
                    for idx, atom in enumerate(atm_nums)
                    if atom not in atomic_valences
                ]
```

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -102,9 +118,153 @@ def add_atom(self, atm_num: int) -> int:
...
+                        f"use bond type wildcard ~ to match bonds in this molecule."
+                    )
+                    continue
+
+                atm_num_valences = []
+                for i, j in enumerate(cluster_no_dummes):
+                    num_bonds = self.rdkit_mol.GetAtomWithIdx(j).GetDegree()
+                    atm_num_valences.append(
+                        len([k for k in atomic_valences[atm_nums[i]] if num_bonds <= k])
+                    )
```
oerc0122:
```suggestion
                for atm_num, j in zip(atom_nums, cluster_no_dummes, strict=False):
                    num_bonds = self.rdkit_mol.GetAtomWithIdx(j).GetDegree()
                    atm_num_valences.append(
                        more_itertools.quantify(
                            atomic_valences[atom_num],
                            pred=lambda k: num_bonds <= k
                        )
                    )
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/MoleculePreviewWidget.py
```diff
@@ -94,14 +94,16 @@ def prepare_rdkit_molecule(self) -> chem.RWMol | None:
...
         for bond in self.mol_info["bond_list"]:
-            new_pair = mapping[bond[0]], mapping[bond[1]]
-            bond_index = submolecule.AddBond(new_pair[0], new_pair[1])
-            self.bond_mapping[bond_index] = new_pair
+            i, j = mapping[bond[0]], mapping[bond[1]]
+            rdkit_bond = large_molecule.GetBondBetweenAtoms(bond[0], bond[1])
+            if rdkit_bond is None:
+                submolecule.AddBond(i, j, chem.rdchem.BondType.UNSPECIFIED)
+            else:
+                submolecule.AddBond(i, j, rdkit_bond.GetBondType())
```
oerc0122:
```suggestion
            bond_type = rdkit_bond.GetBondType() if rdkit_bond is not None else chem.rdchem.BondType.UNSPECIFIED
            submolecule.AddBond(i, j, bond_type)
```

oerc0122 (approved):
Looks good, just a couple of potential stylistic things.

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
ChiCheng45: Not right now, this was mostly added to determine whether there would be a lot of valence combinations and to avoid calling the computationally expensive rdkit code. I think there is a fix to their algorithm https://github.com/rdkit/rdkit/issues/7922, hopefully it could be implemented and this bit could be removed.

MBartkowiakSTFC:
I tried several trajectories, but for every single one so far I have been getting:
```
2025-10-22 12:16:05,814 - INFO - process[21847] - ChemicalSystem 246 - Determining bond orders for molecule H2_O1 with index 0
2025-10-22 12:16:05,817 - ERROR - process[21847] - ChemicalSystem 256 - Error determining bond orders for molecule H2_O1 with index 0: Python argument types in
    rdkit.Chem.rdDetermineBonds.DetermineBondOrders(RWMol)
did not match C++ signature:
    DetermineBondOrders(RDKit::ROMol {lvalue} mol, int charge=0, bool allowChargedFragments=True, bool embedChiral=True, bool useAtomMap=False). SMARTS pattern matching will not work as expected. Bond types set to UNSPECIFIED use bond type wildcard ~ to match bonds in this molecule.
```

ChiCheng45:
> I tried several trajectories, but for every single one so far I have been getting:
> 
> ```
> 2025-10-22 12:16:05,814 - INFO - process[21847] - ChemicalSystem 246 - Determining bond orders for molecule H2_O1 with index 0
> 2025-10-22 12:16:05,817 - ERROR - process[21847] - ChemicalSystem 256 - Error determining bond orders for molecule H2_O1 with index 0: Python argument types in
>     rdkit.Chem.rdDetermineBonds.DetermineBondOrders(RWMol)
> did not match C++ signature:
>     DetermineBondOrders(RDKit::ROMol {lvalue} mol, int charge=0, bool allowChargedFragments=True, bool embedChiral=True, bool useAtomMap=False). SMARTS pattern matching will not work as expected. Bond types set to UNSPECIFIED use bond type wildcard ~ to match bonds in this molecule.
> ```

What version of rdkit are you using? Here is the one I am using.

rdkit                            2025.3.5         pypi_0                   pypi

MBartkowiakSTFC:
> 
> What version of remit are you using? Here is the one I am using.
> 
> rdkit 2025.3.5 pypi_0 pypi

Mine was 2024.9.5 and I thought it would have got updated when I did `pip install --upgrade` on MDANSE itself, but in the end I had to upgrade it manually. Now I have rdkit 2025.9.1 and the bond detection works.

MBartkowiakSTFC (commented):
Looks good, definitely works better than before.

If you wanted to add some other pre-defined SMARTS strings to the current set, my preference would be:
1. Carboxylic acid
`[CX3](=O)[OX2H1]`
2. Unfused benzene ring
`[cR1]1[cR1][cR1][cR1][cR1][cR1]1`

Should we have a lower limit on the version of `rdkit` in our `pyproject.toml`? You could set it to the version you are using now, since we know that it works, and so does the latest one.

Also, interesting to see if the RdKit authors accept your suggested changes.

oerc0122:
> Looks good, definitely works better than before.
> 
> If you wanted to add some other pre-defined SMARTS strings to the current set, my preference would be:
> 
> 1. Carboxylic acid
>    `[CX3](=O)[OX2H1]`
> 2. Unfused benzene ring
>    `[cR1]1[cR1][cR1][cR1][cR1][cR1]1`
> 
> Should we have a lower limit on the version of `rdkit` in our `pyproject.toml`? You could set it to the version you are using now, since we know that it works, and so does the latest one.
> 
> Also, interesting to see if the RdKit authors accept your suggested changes.

Do we at some point want some sort of setting to allow users to define their own custom groups? Definitely not this PR, but might be worth an issue.

MBartkowiakSTFC:
> 
> Do we at some point want some sort of setting to allow users to define their own custom groups? Definitely not this PR, but might be worth an issue.

I added this as issue #1040

oerc0122 (approved):

MBartkowiakSTFC:
Will this close #286?

ChiCheng45:
> Will this close #286?

Yes this will close #286.

ChiCheng45:
> Looks good, definitely works better than before.
> 
> If you wanted to add some other pre-defined SMARTS strings to the current set, my preference would be:
> 
> 1. Carboxylic acid
>    `[CX3](=O)[OX2H1]`
> 2. Unfused benzene ring
>    `[cR1]1[cR1][cR1][cR1][cR1][cR1]1`
> 
> Should we have a lower limit on the version of `rdkit` in our `pyproject.toml`? You could set it to the version you are using now, since we know that it works, and so does the latest one.
> 
> Also, interesting to see if the RdKit authors accept your suggested changes.

I've added both groups above. I've changed them slightly so that the carboxylic acid group also matches the hydrogen.

image

And the unfused benzene matches all hydrogens are bonded to it. The SMARTS pattern is "\[cR1r6\]([H]).[cR1r6]".

image

MBartkowiakSTFC (commented):
At the moment I have been testing it on the trajectories from SlimMD, and they all have large molecules for which rdkit fails. Then I can still create a selection using SMARTS, but I need to insert `~` every time in the strings. Do you think we need to add some extra hints for the users who are not familiar with SMARTS, to show them where these should be added? You already mention it in the log message. We could add an example to the current list of pre-defined SMARTS strings, or add some note in the documentation, if you think it is meaningful.

ChiCheng45:
> At the moment I have been testing it on the trajectories from SlimMD, and they all have large molecules for which rdkit fails. Then I can still create a selection using SMARTS, but I need to insert `~` every time in the strings. Do you think we need to add some extra hints for the users who are not familiar with SMARTS, to show them where these should be added? You already mention it in the log message. We could add an example to the current list of pre-defined SMARTS strings, or add some note in the documentation, if you think it is meaningful.

I've updated to warning log and updated to documentation. Also, I think we could increase the max number of atoms that the bond type determination but I wanted to make sure that it never causes any performance issues. Perhaps, we should add some sort of setting to allow the user to increase that max number of atoms for bond determination, or move it so that it is done in the trajectory editor or something else.

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -102,9 +119,122 @@ def add_atom(self, atm_num: int) -> int:
...
+            i, j = int(pair[0]), int(pair[1])
+            if i not in self._rdkit_dummy_atms and j not in self._rdkit_dummy_atms:
+                # Ignore the bonds to any dummy atoms. This means that
+                # the SMARTS pattern match will ignore any bonds to
+                # dummy atoms. E.g. [OX2] will match oxygen atoms connected
+                # to two other real atoms even if it is also connected to
+                # dummy atoms.
+                self.rdkit_mol.AddBond(i, j, Chem.rdchem.BondType.UNSPECIFIED)
+
+    def set_bond_orders(self, coords: np.ndarray, max_iters=1000, max_natms=100):
```
oerc0122:
```suggestion
    def set_bond_orders(self, coords: np.ndarray, *, max_iters: int = 1000, max_natms: int = 100):
```

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -93,6 +104,12 @@ def initialise_atoms(
         if name_list is not None:
             self._atom_names = [str(x) for x in name_list]
 
+        self._rdkit_dummy_atms = [
```
oerc0122: If we're only ever checking `__contains__` (`in`) do we want this to be a set rather than a list (constant time vs. linear)

MDANSE_GUI/Src/MDANSE_GUI/Widgets/SelectionWidgets.py
```diff
@@ -496,7 +499,7 @@ def add_specific_widgets(self):
         layout.addWidget(self.selection_field)
         self.selection_field.addItems(self.pattern_dictionary.keys())
         layout.addWidget(QLabel("pattern:"))
-        self.input_field = QLineEdit("", self)
+        self.input_field = QLineEdit(next(iter(self.pattern_dictionary.values())), self)
```
```suggestion
        self.input_field = QLineEdit(more_itertools.first(self.pattern_dictionary.values()), self)
```

oerc0122 (approved):

MBartkowiakSTFC (approved):
Thanks for the changes, looks good.

PR #1043, opened 2025-10-23
Undo and redo atom selection operations

MBartkowiakSTFC:
**Description of work**
At the moment the only solution to applying a wrong atom selection operation is to reset the selection and start over. This PR adds a simple undo/redo mechanism using `QUndoStack`.

**Fixes**
1. The last atom selection can always be undone using the 'Undo' button. All the undone operations can be re-applied by clicking 'Redo'.
2. Fixed the problem with loading a selection from a file. (At the moment, it fails if the selection dialog had not been opened at least once prior to loading.)

**To test**
Create complicated atom selections, and check that each step can be correctly undone and reapplied using the buttons. Include manual selection in the tests. Try it also for atom transmutation and partial charges.


ChiCheng45:
Looks good, just a few things. Here is one bug I found that causes a crash.

- make a normal selection
- undo the selection
- make a selection with the 3D view but do not click confirm
- click redo

Animation5

In this example, you can undo the initial select all operations so that it is slightly odd after manual selection.

- undo all operations
- make a manual selection
- click confirm

Animation6

MBartkowiakSTFC:
Is it better now? I think I fixed both of them, in a way. Now you can't undo the first operation, and manual selection switches off the redo button.

ChiCheng45 (approved):
Looks good.

PR #1044, opened 2025-10-23
Plot settings for 2D datasets in Single mode

### LINKED ISSUE #569: [ENHANCEMENT] Update the single plotting so that it uses the linestyle and marker setting
**Description of the error**
The single plot style doesn't use the linestyle or marker setting, should we update the single plot to use them?

image

MBartkowiakSTFC:
**Description of work**
At the moment it is not possible to customise the appearance of curves created from 2D datasets in `Single` plot mode. The main argument for ignoring the settings has been that it would make multiple curves from a single dataset appear exactly the same.

This PR makes 2D datasets use the plot settings (colour, line style and marker), introducing a colour gradient in each group of curves.

closes #569 

Example plot:
sample_plot_multicurve

**Fixes**
1. Single plotter uses plot settings to change the appearance of curves.
2. The colour chosen by the user is used as a starting value, and then gradually changed.

**To test**
Please check if the plots are legible.
Other ideas for calculating the colour increment in a dataset are welcome.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Single.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Single.py
@@ -224,6 +224,11 @@ class Single(Plotter):
                             linestyle=databundle.line_style,
                             color=tuple(main_colour),
                         )
+                        try:
+                            temp.set_marker(databundle.marker)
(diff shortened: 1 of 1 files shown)


ChiCheng45 (approved):
Looks good. Plots look great.

PR #1045, opened 2025-10-24
Descriptors rework

oerc0122:
**Description of work**
Complete rework of MDANSE infrastructure to replace parameter dictionaries with [`descriptor`](https://docs.python.org/3/howto/descriptor.html) `Parameter`s with validators. 

**Fixes**
- Moves the emphasis of data validation out of the GUI
- Unifies the validation between CLI and GUI
- More Pythonic variable access
- Easier to define and write extensible objects and functions
- Reduces number of unique objects and modules

**To test**
This is a complete rework and will require significant and thorough testing.


ChiCheng45 (commented):
Issues I have come across so far in the GUI.

- Converter and analysis jobs fails with similar errors `    raise ConfigError(f"Value ({value}) is not a valid Path.") from error
MDANSE.Framework.Parameters.Parameters.ConfigError: Value (<object object at 0x000001E60D32BEF0>) is not a valid Path.`.
- MDAnalysis converter does not allow the job to be run when only the topology file is used. It should be able to do this.

Image

- MDAnalysis converter does not allow the job to be run, even when both topology and coordinate files are input.

Image

- Similar to the above for the MDTraj converter.
- The output files box is missing on the average structure job.

Image

- Default Q-vectors settings are not the same as protos.
- Changing the instrument in the GUI does not update the q_vector settings for scattering jobs.

Image

- Preview output axes not being displayed for all analysis jobs
Image

- The interpolation order widget is not updating correctly; it stays in the error state even when the settings are correct and vice versa.

Animation40

- The optional checkbox is slightly confusing. A tick suggests that it would be used.

image
image

- Save as script does not copy the parameters that were used in the GUI e.g., the atom selection setting

- Trajectory filter widget is blank.

image

- The molecule name, frame, and other widgets are not updating on trajectory change.

Animation41

- MDANSE job logs are missing in the running jobs tab.
Animation43

PR #1046, opened 2025-10-28
Extend output prediction

### LINKED ISSUE #1009: [BUG] Some jobs don't use the output prediction properly
**Description of the error**
Some analysis job in MDANSE do not show the output prediction correctly. Here are some examples.

- AverageStructure: shows an output prediction of times.
- EISF: shows times.
- NDTSF: shows nothing.
- SLDP: shows times.
- SSF: shows times.
- SFFSF: shows nothing.
- XRaySSF: shows times.

I think the output prediction should be updated so that it shows the variables of the main results. E.g. the main results of SLDP is a plot with main axis of r (distance). The main results of DCSF has a plot with main axis of q (inverse distance) and omega (frequency).

MBartkowiakSTFC:
**Description of work**
An attempt to show the analysis output prediction in a more consistent way.

Example for van Hove function, distinct part. Current main branch:
Screenshot 2025-10-28 at 08 45 15

This branch:
Screenshot 2025-10-28 at 08 46 07

closes #1009

**Fixes**
1. Analysis jobs that output .mda files (and do NOT use other .mda files as input) show the prediction of the output axes.
2. Only the axes relating to the _main_ result of the analysis are shown. (For DOS there is no prediction of the time axis, which relates to VACF, etc.)
3. Optional labels can be added to configurators for output prediction. Instead of `r_values` I use `Interatomic distance`, etc.

**To test**
Check if the output prediction gets updated correctly for different trajectories and analysis types.


MDANSE/Src/MDANSE/Framework/Configurators/GridStepConfigurator.py
```diff
@@ -0,0 +1,90 @@
...
+from MDANSE.MolecularDynamics.Trajectory import Trajectory
+
+
+class GridStepConfigurator(FloatConfigurator):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self["grid"] = []
+        self.prediction_key = "grid"
+        self.prediction_unit = "nm"
+        self.prediction_label = "Real space bin centres"
```
oerc0122: I think this might be best expressed universally as a `typing.NamedTuple`

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -124,6 +125,7 @@ def __init__(
 
         """
         self.name = name
+        self.prediction_label = kwargs.get("prediction_label", name)
```
oerc0122: Prefer full args over `**args`

```
        show_prediction: bool = False,
        prediction_label: str = "",
```

```suggestion
        self.prediction_label = prediction_label or name
```

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -330,3 +334,19 @@ def check_dependencies(self, configured: list[str] | None = None) -> bool:
...
+            for key in self.prediction_keys:
+                yield (key, self[key], self.prediction_unit)
+        elif self.prediction_key in self:
+            yield (
+                self.prediction_label,
+                self[self.prediction_key],
+                self.prediction_unit,
+            )
+        else:
+            yield None
```
```suggestion
        if not self.show_prediction or not self.is_configured() or not self.valid:
            return

        if hasattr(self, "prediction_keys"):
            for key in self.prediction_keys:
                yield (key, self[key], self.prediction_unit)
        elif self.prediction_key in self:
            yield (
                self.prediction_label,
                self[self.prediction_key],
                self.prediction_unit,
            )
```

MDANSE/Src/MDANSE/Framework/Jobs/AreaPerMolecule.py
```diff
@@ -50,7 +50,7 @@ class AreaPerMolecule(IJob):
     settings["trajectory"] = ("HDFTrajectoryConfigurator", {})
     settings["frames"] = (
         "FramesConfigurator",
-        {"dependencies": {"trajectory": "trajectory"}},
+        {"dependencies": {"trajectory": "trajectory"}, "show_prediction": True},
```
oerc0122: I do wonder if a better way to handle this would be a classarg on the job listing the settings which should "show prediction" (maybe just a classarg tuple e.g. here `PREDICTORS = ("frames",)`). 

That, to me, puts the duty on the job rather than the parameter, even if the parameter is ultimately what's called to get the data.

MDANSE/Src/MDANSE/Framework/Jobs/IJob.py
```diff
@@ -307,14 +307,19 @@ def set_up_trajectory(self):
...
+        axes = []
         for configurator in self._configuration.values():
-            preview_method = getattr(configurator, "preview_output_axis", None)
-            if callable(preview_method):
-                axis, unit = preview_method()
-                if axis is not None:
-                    axes[unit] = axis
+            for prediction in configurator.preview_output_axis():
+                if prediction is not None:
+                    axes.append(prediction)
```
oerc0122, line 324: With the change suggested above, this would become:

```suggestion
        axes = []
        for predictor in self.PREDICTORS:
            configurator = self._configuration[predictor]
            for prediction in configurator.preview_output_axis():
                if prediction is not None:
                    axes.append(prediction)
```

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -337,16 +342,14 @@ def check_dependencies(self, configured: list[str] | None = None) -> bool:
...
+                yield (key, self[key], self.prediction.unit)
+        elif self.prediction.key in self:
             yield (
-                self.prediction_label,
-                self[self.prediction_key],
-                self.prediction_unit,
+                self.prediction.label,
+                self[self.prediction.key],
+                self.prediction.unit,
             )
```
oerc0122, line 375:
```suggestion
            yield self.prediction
```
This is already a tuple-like entity.

MDANSE/Src/MDANSE/Framework/Jobs/XRayStaticStructureFactor.py
```diff
@@ -67,6 +67,7 @@ class XRayStaticStructureFactor(DistanceHistogram):
         "Analysis",
         "Scattering",
     )
+    PREDICTORS = ("q_values",)
```
oerc0122, line 70: Is `r_values` no longer a predictor?

MBartkowiakSTFC: That's right. Yesterday we agreed that we should only output the prediction of the axes of the main result. In x-ray static structure factor, the result is given as `S(q)`, so only the q values are used for output prediction.

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
MBartkowiakSTFC: `yield self.prediction` will give us ("label", "key", "unit"). The current version gives ("label", Sequence[float], "unit"). The `key` is used to pick one of the many dictionary entries in the `Configurator` which contains numbers we can use.

oerc0122: Ah, I see, `key` is a key in the `IJob` dictionary. A little confusing, might be worth a quick comment.

MDANSE/Src/MDANSE/Framework/Configurators/IConfigurator.py
```diff
@@ -78,6 +79,13 @@ def __str__(self) -> str:
...
+    """Strings used for showing the predicted output range in the GUI.
+
+    key - The dictionary key of where the Sequence[float] is stored in the Configurator.
+    label - An explanation for the user of what physical property the numbers represent.
+    unit - The internal unit of MDANSE of the number array. Used for converting to GUI units.
+    """
+
     key: str
     label: str
     unit: str = "none"
```
oerc0122, line 91: It is strictly possible to document the parameters directly, we'd need to double check the sphinx format we use supports it though.
```suggestion
    """
    
    #: The dictionary key of where the Sequence[float] 
    #: is stored in the Configurator.
    key: str
    #: An explanation for the user of what physical 
    #: property the numbers represent.
    label: str
    #: The internal unit of MDANSE of the number array.
    #: Used for converting to GUI units.
    unit: str = "none"
```

oerc0122 (approved):
I think it looks good as-is.

PR #1050, opened 2025-10-29
Close matplotlib setting editor when MDANSE closes and update marker settings

ChiCheng45:
**Description of work**
Small fixes to the GUI related to the plotting tab. A None setting was missing from the markers.

Screenshot 2025-10-29 at 17 21 53

Matplotlib settings windows would remain open even if MDANSE is closed.

**Fixes**
Maplotlib setting window did not close when manse is closed.
Added None setting to the plotting markers selection and is set to the default. It looks like it was missing.

**To test**
Check that the matplotlib setting windows closes when MDANSE is closed. Check that the markers work as expected.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
@@ -63,7 +63,7 @@ def get_mpl_markers():
         key: mark
         for key, mark in MarkerStyle.markers.items()
         if isinstance(key, str)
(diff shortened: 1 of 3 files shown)


oerc0122 (changes requested):
Looks good. Though "None" doesn't appear in the combobox for the marker after I've changed it.

ChiCheng45:
> Looks good. Though "None" doesn't appear in the combobox for the marker after I've changed it.

Thanks, this should be fixed now.

oerc0122 (approved):
Looks fixed. Let's get this merged.

PR #1051, opened 2025-10-30
Calculate all components of the DOS, PPS, VCF, and PCF

ChiCheng45:
**Description of work**
Calculates the individual components (xx, xy, xz, yy, yz, zz, and isotropic (xx + yy + zz) / 3) of the DOS, VCF (velocity auto- and cross-correlation functions), PPS, and PCF (positional auto- and cross-correlation functions).

There was a lot of duplicated code between the VACF, PACF, DOS, and PPS code from protos. I've generalised the code a bit to reduce the duplication.

**To test**
Check that the isoptropic results from the DOS, PPS, VCF and PCF are unchanged from protos calculations.


MBartkowiakSTFC:
Just a question: before the changes, Position Power Spectrum was using `weights=atomic_weight` (mainly to be directly comparable to the Density of States), and now it is using `weights=equal`. Did we want this to be the case?

ChiCheng45:
> Just a question: before the changes, Position Power Spectrum was using `weights=atomic_weight` (mainly to be directly comparable to the Density of States), and now it is using `weights=equal`. Did we want this to be the case?

I changed it back so that it was consistent with previous version. Could change it back.

MBartkowiakSTFC:
As far as I am concerned, you can make both DOS and PPS use `atomic_weights` to stay consistent with protos.

MBartkowiakSTFC (approved):
I did not find any problems. The isotropic results are reproduced correctly, cross terms are close to zero for normal systems. Also, reduced code duplication is also a plus. Looks good.

PR #1053, opened 2025-10-30
Escape html in the trajectory and plotting info

### LINKED ISSUE #1034: [BUG] Text field in Plot Creator and molecule names
**Description of the error**
Since our QTextEdit widgets have HTML enabled (mainly for the hyperlinks to work), some molecule names have side effects.

This is the information about the dataset `/sas/taken/<H2>/H`
Image

The `<H2>` molecule name should be shown as normal text.

The molecule label `<H2>` is interpreted as HTML header tag.

**Suggested fix**
Not sure yet. There should be some tag to stop the text from being interpreted as HTML.

**Additional details**
N/A

ChiCheng45:
**Description of work**
Escape the HTML in the trajectory and plotting info text boxes. This should resolve cases when custom atom types or molecules have names that are also HTML valid syntax.

**Fixes**
Fixes #1034

**To test**
Load a trajectory and check that the trajectory information text box looks fine. Load an analysis results and check that the plotting information text box looks fine.


MBartkowiakSTFC:
This is much better, but I still get funny output in the Logger tab when I select a dataset for plotting:
Screenshot 2025-11-03 at 10 13 28

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/PlotDataView.py
```diff
@@ -255,13 +257,20 @@ def on_select_dataset(self, index):
             packet = text, mda_data_structure.file
         self.dataset_selected.emit(packet)
         if hasattr(mda_data_structure, "_metadata"):
-            self.item_details.emit(mda_data_structure._metadata)
+            self.item_details.emit(
```
oerc0122, line 260: Rather than escaping on `emit` would it not be less impact to escape on the `@Slot` function? Do we ever pass deliberately marked up text?

ChiCheng45: There are some instances where TextInfo is passed html, so it I can't just escape html there.

https://github.com/ISISNeutronMuon/MDANSE/blob/29324d4738d00f720c0de54d79fa592f482145da/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/ActionsTree.py#L55-L63

https://github.com/ISISNeutronMuon/MDANSE/blob/29324d4738d00f720c0de54d79fa592f482145da/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/JobTree.py#L112-L115

ChiCheng45:
> This is much better, but I still get funny output in the Logger tab when I select a dataset for plotting: Screenshot 2025-11-03 at 10 13 28

Thanks for spotting. I think this should be fixed now.

MBartkowiakSTFC:
Now I am just going to unreasonable lengths to create a problem:
Screenshot 2025-11-03 at 11 49 43

Do you think that we need to do anything about users who deliberately include HTML tags in their trajectory file names? You are welcome to fix this one as well. Or we can decide to fix it in another PR. In any case, I do not think that any sensible user would do this to a file name...

ChiCheng45:
> Now I am just going to unreasonable lengths to create a problem: Screenshot 2025-11-03 at 11 49 43
> 
> Do you think that we need to do anything about users who deliberately include HTML tags in their trajectory file names? You are welcome to fix this one as well. Or we can decide to fix it in another PR. In any case, I do not think that any sensible user would do this to a file name...

I guess it's possible that they could do this. I just need add another html.escape, so we might as well do it. Can you think of any other places which this could pop up?

MBartkowiakSTFC:
> 
> I guess it's possible that they could do this. I just need add another html.escape, so we might as well do it. Can you think of any other places which this could pop up?

I'm out of ideas. I tried a few other things, but your changes covered them already.

MBartkowiakSTFC (approved):
I could not find any new ways of making text appear wrong in the GUI. Looks good. Thanks for the changes.

PR #1054, opened 2025-11-03
Add recently used files option to g UI

IfyyAgu:
**Description of work**
Add an option in the GUI to open recently used files. 
The File menu will have a new section. 
There will be a way to select a recently opened trajectory file, which will then be opened in the trajectory tab. 
There will be a way to select a recently opened results file (.mda), which will then be opened in the plot creator tab.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Session/RecentFiles.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Session/RecentFiles.py
@@ -1,0 +1,65 @@
+#    This file is part of MDANSE_GUI.
+#
+#    MDANSE_GUI is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation, either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from MDANSE.MLogging import LOG
+
+
+class RecentFiles:
+    """Stores recently used file names in a JSON file."""
+
+    def __init__(self, file_path: Path, max_file_count: int, placeholder_text: str):
+        self.json_file_path = file_path
+        self.max_files = max_file_count
+        self.placeholder_text = placeholder_text
+        self.create_initial_file()
+
+    def create_initial_file(self):
+        """Initialise the JSON file with a placeholder text string."""
+        if not self.json_file_path.exists():
+            with self.json_file_path.open("w", encoding="utf-8") as file:
+                json.dump([self.placeholder_text], file, indent=4)
+
+    def store_recently_used_filename(self, loaded_file: str):
+        """Updating recent file list with successfully loaded file
+
+        Parameters
+        ----------
+        loaded_file : str
+            Path of the successfully loaded file
+        """
+        # if the recent file exists and not empty
+        if self.json_file_path.is_file():
+            try:
+                with self.json_file_path.open(encoding="utf-8") as file:
+                    recent_files = json.load(file)
+            except json.decoder.JSONDecodeError as err:
+                LOG.error("Invalid recent files json %s: %s", self.json_file_path, err)
+                recent_files = []
+        else:
+            recent_files = []
+
+        if loaded_file in recent_files:
+            recent_files.remove(loaded_file)
+        recent_files.append(loaded_file)
+
+        recent_files = recent_files[-self.max_files :]
+
+        with self.json_file_path.open("w", encoding="utf-8") as file:
+            json.dump(recent_files, file, indent=4)
--- a/MDANSE_GUI/Src/MDANSE_GUI/Session/RecentFiles.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Session/RecentFiles.py
@@ -45,15 +45,14 @@ class RecentFiles:
             Path of the successfully loaded file
         """
         # if the recent file exists and not empty
+        recent_files = []
+
         if self.json_file_path.is_file():
             try:
                 with self.json_file_path.open(encoding="utf-8") as file:
                     recent_files = json.load(file)
             except json.decoder.JSONDecodeError as err:
                 LOG.error("Invalid recent files json %s: %s", self.json_file_path, err)
-                recent_files = []
-        else:
-            recent_files = []
 
         if loaded_file in recent_files:
             recent_files.remove(loaded_file)
--- a/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
@@ -189,15 +189,16 @@ class TabbedWindow(QMainWindow):
         file_group = menubar.addMenu("File")
         settings_group = menubar.addMenu("Settings")
         help_group = menubar.addMenu("Help")
-        self.exitAct = QAction("Exit", parent=menubar)
-        self.exitAct.triggered.connect(self.shut_down)
-        file_group.addAction(self.exitAct)
         self.recent_trajectory_fileAct = QMenu("Open Recent Trajectories File", parent=menubar)
         self.recent_trajectory_fileAct.aboutToShow.connect(self.populate_recent_trajectory_menu)
         file_group.addMenu(self.recent_trajectory_fileAct)
         self.recent_plot_selection_fileAct = QMenu("Open Recent Plot Selection File", parent=menubar)
         self.recent_plot_selection_fileAct.aboutToShow.connect(self.populate_recent_plot_selection_menu)
         file_group.addMenu(self.recent_plot_selection_fileAct)
+        file_group.addSeparator()
+        self.exitAct = QAction("Exit", parent=menubar)
+        self.exitAct.triggered.connect(self.shut_down)
+        file_group.addAction(self.exitAct)
         self.settingsAct = QAction("User Settings", parent=menubar)
         self.settingsAct.triggered.connect(self.launchSettingsEditor)
         settings_group.addAction(self.settingsAct)
@@ -231,24 +232,16 @@ class TabbedWindow(QMainWindow):
                 if not isinstance(data, list):
                     raise ValueError("Recent trajectory file is not a list")
                 data = list(reversed(data))
-                #print("Reopen closed trajectory files:")
                 for file in data:
-                    print(file)
                     action = QAction(file, self.recent_trajectory_fileAct)
                     action.triggered.connect(lambda checked=False, fp=file: self.open_recent_trajectory_file(fp))
                     self.recent_trajectory_fileAct.addAction(action)
         else:
-            # QMessageBox.information(
-            #     self,
-            #     "No Recent Trajectory File",
-            #     f"No recent trajectory file found at {filename}.",
-            # )
             return []
 
     @Slot()
-    #create a signal that will send a single path to the trajectory tab to load recent files
     def open_recent_trajectory_file(self, file: str):
-        #emit signal with the file name as the argument
+        """Emit signal to the trajectory tab to load the file with the file path as the argument."""
         self.signal_recent_trajectory_file.emit(file)
 
     def populate_recent_plot_selection_menu(self, filename=PlotSelectionTab.DEFAULT_JSON_PATH):
@@ -260,18 +253,15 @@ class TabbedWindow(QMainWindow):
                 if not isinstance(data, list):
                     raise ValueError("Recent plot selection file is not a list")
                 data = list(reversed(data))
-                #print("Reopen closed plot selection files:")
                 for file in data:
-                    print(file)
                     action = QAction(file, self.recent_plot_selection_fileAct)
                     action.triggered.connect(lambda checked=False, fp=file: self.open_recent_plot_selection_file(fp))
                     self.recent_plot_selection_fileAct.addAction(action)
         else:
             return []
     @Slot()
-    #create a signal that will send a single path to the plot selection tab to load recent files
     def open_recent_plot_selection_file(self, file: str):
-        #emit signal with the file name as the argument
+        """ Emit signal to the plot selection tab to load the file with the file path as the argument."""
         self.signal_recent_plot_selection_file.emit(file)
 
     def version_information(self):
--- a/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
@@ -19,6 +19,7 @@ import json
 import os
 from collections import defaultdict
 from importlib import metadata
+from pathlib import Path
 
 from qtpy.QtCore import QMessageLogger, QSize, Qt, QTimer, QUrl, Signal, Slot
 from qtpy.QtGui import QDesktopServices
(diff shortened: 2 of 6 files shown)


MBartkowiakSTFC:
I have two comments for now:

1. Please rebase your branch to make sure that all the recent changes to `protos` branch have been included and merge conflicts resolved.
2. Could you make the saving process more selective, so it stores the trajectory as recently used only if it had been loaded _successfully_?

IfyyAgu:
I have included the changes you suggested including rebase.

MBartkowiakSTFC:
Sorry for not writing it explicitly before: do you think you could also make the `.mda` files appear on the list only if they were loaded successfully?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -185,6 +204,42 @@ def add_file(self, filename: str):
...
+        Str
+            Recently loaded plot selection file path.
+        """
+        filename = self.DEFAULT_JSON_PATH
+        max_num_files = self.MAX_NUMBER_RECENT_FILES
+        if os.path.exists(filename) and os.path.getsize(filename) > 0:
+            with open(filename) as f:
+                recent_files = json.load(f)
+        else:
+            recent_files = []
```
oerc0122:
```suggestion
        filepath = Path(filename)
        if filepath.is_file():
            with filepath.open(encoding="utf-8") as file:
                recent_files = json.load(file)
        else:
            recent_files = []
```
Prefer using `Path`.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -185,6 +204,42 @@ def add_file(self, filename: str):
...
+
+        if plot_filepath in recent_files:
+            recent_files.remove(plot_filepath)
+        recent_files.append(plot_filepath)
```
oerc0122: What is this block supposed to be doing? Reducing it to its last element?
```suggestion
        recent_files = recent_files[-1:]
```
Though I'm not sure why you'd want this?
Are you trying to reverse the array, but have dropped the other outside the loop? `reversed(recent_files)` would do that.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -185,6 +204,42 @@ def add_file(self, filename: str):
...
+
+        if len(recent_files) > max_num_files:
+            delete_number_of_files = len(recent_files) - max_num_files
+            while delete_number_of_files > 0:
+                recent_files.pop(0)
+                delete_number_of_files -= 1
```
oerc0122:
```suggestion
        recent_files = recent_files[:max_num_files]
```
No reason to have these loops also not sure why you're popping from the front? If you want that you may need to reverse the array or have some different logic.

However, in future, if you want to iterate over a range use `range`, `while` should only be used for an unknown length of loop:


```suggestion
        if len(recent_files) > max_num_files:
            delete_number_of_files = len(recent_files) - max_num_files
            for _ in range(delete_number_of_files):
                recent_files.pop(0)
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -185,6 +204,42 @@ def add_file(self, filename: str):
...
+
+        with open(filename, "w") as f:
+            json.dump(recent_files, f, indent=4)
```
oerc0122:
```suggestion
        with filepath.open(, "w", encoding="utf-8") as file:
            json.dump(recent_files, file, indent=4)
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -152,8 +160,33 @@ def accept_results(self, result_tuple) -> None:
...
+        recent_files.append(traj_filepath)
+
+        if len(recent_files) > max_num_files:
+            delete_number_of_files = len(recent_files) - max_num_files
+            while delete_number_of_files > 0:
+                recent_files.pop(0)
+                delete_number_of_files -= 1
+
+        with open(filename, "w") as f:
+            json.dump(recent_files, f, indent=4)
```
oerc0122: Given this is a recurring theme, perhaps it wants to be a function somewhere else and just called in these to reduce duplicated code.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/PlotSelectionTab.py
```diff
@@ -15,6 +15,7 @@
 #
 from __future__ import annotations
 
+import json
```
oerc0122: Unused?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/TrajectoryTab.py
```diff
@@ -60,9 +60,13 @@ def load_trajectories(self):
             str(self.get_path("trajectory")),
             "HDF5 files, MDANSE or H5MD format (*.mdt *.h5);;H5MD files (*.h5);;All files (*)",
         )
+
+        loaded_files = []  # list to store loaded files for recent files tracking
+
         for fname in fnames[0]:
             self.load_trajectory(PurePath(fname))
             last_path = str(PurePath(os.path.split(fname)[0]))
+            loaded_files.append(str(PurePath(fname)))
```
oerc0122:
```suggestion
        loaded_files = []  # list to store loaded files for recent files tracking

        for fname in fnames[0]:
            filepath = Path(fname)
            self.load_trajectory(filepath)
            last_path = str(filepath.parent)
            loaded_files.append(str(filepath))
```
Using `Path`s effectively will be easier.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/TrajectoryTab.py
```diff
@@ -60,9 +60,13 @@ def load_trajectories(self):
...
         if fnames[0]:
             self.set_path("trajectory", str(PurePath(last_path)))
```
oerc0122: This is already a string? 

```suggestion
            self.set_path("trajectory", last_path)
```

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
```diff
@@ -203,6 +232,58 @@ def shut_down(self):
...
+                for file in data:
+                    action = QAction(file, self.recent_trajectory_fileAct)
+                    action.triggered.connect(
+                        lambda checked=False, fp=file: self.open_recent_trajectory_file(
+                            fp
+                        )
+                    )
+                    self.recent_trajectory_fileAct.addAction(action)
+        else:
+            return []
```
oerc0122: Prefer minimal nesting and indentation


```suggestion
        filepath = Path(filename)

        if not filepath.is_file():
            return []

        with filename.open(encoding="utf-8") as file:
            data = json.load(file)
            for file in data[::-1]:
                action = QAction(file, self.recent_trajectory_fileAct)
                action.triggered.connect(
                    lambda checked=False, fp=file: self.open_recent_trajectory_file(
                        fp
                    )
                )
                self.recent_trajectory_fileAct.addAction(action)
```

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
```diff
@@ -203,6 +232,58 @@ def shut_down(self):
...
+
+    @Slot()
+    def open_recent_trajectory_file(self, file: str):
+        """Emit signal to the trajectory tab to load the file with the file path as the argument."""
+        self.signal_recent_trajectory_file.emit(file)
+
+    def populate_recent_plot_selection_menu(
```
oerc0122, line 334: What's the difference between this and the other method above? If it's just the file-name and the target you may want to look into using [`partial`](https://docs.python.org/3/library/functools.html#functools.partial) and avoid some code duplication.

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
```diff
@@ -203,6 +232,58 @@ def shut_down(self):
...
+                        fp=file: self.open_recent_plot_selection_file(fp)
+                    )
+                    self.recent_plot_selection_fileAct.addAction(action)
+        else:
+            return []
+
+    @Slot()
+    def open_recent_plot_selection_file(self, file: str):
+        """Emit signal to the plot selection tab to load the file with the file path as the argument."""
+        self.signal_recent_plot_selection_file.emit(file)
```
oerc0122, line 350: Likewise here

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
IfyyAgu: This checks if the loaded file is in the recent file list. If it's present then remove it and add it again so it's in order. Just trying to avoid duplications.

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
IfyyAgu: This is emitting signal to plot selection tabs while the other is for trajectory tab

MBartkowiakSTFC:
On my computer, if the list of recently used files was empty on GUI startup, the items in the recently used file menu would not appear, even though new file names had been added to the JSON files. The second time I ran the GUI everything was fine and the list of recently used files would be updated every time I loaded a file. I did not find the reason for it, so I just added an placeholder item to make sure that the list is never empty, and this worked around the problem.

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
MBartkowiakSTFC: The only thing I am not sure about is how well the use of `partial` would agree with Qt's signal/slot mechanism. We have already had problems with `lambda`s  used as slots (e.g. in the context of TrajectoryFilter), so maybe for now we can play it safe and just have explicitly defined methods here.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -154,14 +165,26 @@ class PlotDataModel(QStandardItemModel):
...
     all_elements = Signal(object)
+    finished_loading = Signal(int)
+    recent_files_update = Signal(str)
 
     def __init__(self, parent: QObject = None):
         super().__init__(parent=parent)
         self.mutex = QMutex()
         self._nodes = {}
         self._next_number = 0
+        if not self.DEFAULT_JSON_PATH.exists():
```
oerc0122: If it does exist, it silently does nothing? Is this what we want?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -176,7 +199,7 @@ def add_file(self, filename: str):
         try:
             new_datafile = MDADataStructure(filename)
         except Exception as e:
-            LOG.error(f"Invalid: {e!s}")
+            LOG.error(f"Invalid: {e}")
```
oerc0122: Strictly this should be
```suggestion
            LOG.error(f"Invalid: %s", e)
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -185,6 +208,22 @@ def add_file(self, filename: str):
             self._next_number += 1
             self.appendRow(new_item)
             new_item.populate(new_datafile._file)
+            self.finished_loading.emit(self._next_number - 1)
+            self.recent_files_update.emit(filename)
```
oerc0122: I think it would make more sense to move the `self._next_number` to the end of the block rather than doing a `-1`

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -185,6 +208,22 @@ def add_file(self, filename: str):
...
+
+    @Slot(str)
+    def store_mda_file(self, plot_filepath: str):
+        """Store the input file path in a JSON file.
+
+        Parameters
+        ----------
+        plot_filepath : str
+            _description_
```
oerc0122: This needs filling in

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -31,10 +33,43 @@
...
+    loaded_file : str
+        Path of the successfully loaded file
+    """
+    filepath = Path(recent_file_path)
+    # if the recent file exists and not empty
+    if filepath.is_file() and filepath.stat().st_size > 0:
+        with filepath.open(encoding="utf-8") as file:
+            recent_files = json.load(file)
+    else:
+        recent_files = []
```
oerc0122: May be better duck typed.


```suggestion
    if filepath.is_file():
        try:
            with filepath.open(encoding="utf-8") as file:
                recent_files = json.load(file)
         except json.decoder.JSONDecodeError as err:
             log.error("Invalid recent files json at %s.", filepath) from err
             recent_files = []
    else:
        recent_files = []
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -90,10 +125,15 @@ class TrajectoryModel(QStandardItemModel):
     """Like GeneralModel, but should implement trajectory
     loading in the background."""
 
+    DEFAULT_JSON_PATH = PLATFORM.application_directory() / "recent_trajectory_file.json"
+    MAX_NUMBER_RECENT_FILES = 10  # maximum number of recent files to store
+    PLACEHOLDER_STRING = "Recently used trajectory files (.mdt, .h5)"
```
oerc0122, line 97: This duplicated code and set of constants makes it seem like this recently used files handling might be best delegated to a class attached to these objects; with the appropriate methods (like `store_recently_used_filename` above), dumping placeholders on init, returning appropriately filtered lists, etc.

This could sit on the class where it might be configured, for example.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/TrajectoryTab.py
```diff
@@ -60,16 +60,20 @@ def load_trajectories(self):
...
+            self.load_trajectory(filepath)
+            last_path = str(filepath.parent)
+            loaded_files.append(str(filepath))
+            self.set_path("trajectory", last_path)
             self._session.save()
 
     @Slot(str)
     def load_trajectory(self, some_fname: str):
-        fname = str(PurePath(some_fname))
+        fname = str(Path(some_fname))
```
oerc0122: I don't understand why this line is here or what it's doing. Later it's only treated like a string, and I'm not sure the Path/str loop will do anything.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
MBartkowiakSTFC: In the current implementation of the code, the GUI checks the contents of the JSON file every time we select the 'recently used files' menu, but not before. A weird problem (at least on my machine) caused the menu not to appear at all throughout the GUI session if the JSON file was nonexistent on startup. This code is just meant to make sure that a file with this name exists, which should be the case after the user worked with this version of the code once.

MDANSE_GUI/Src/MDANSE_GUI/Session/RecentFiles.py
```diff
@@ -0,0 +1,65 @@
...
+        # if the recent file exists and not empty
+        if self.json_file_path.is_file():
+            try:
+                with self.json_file_path.open(encoding="utf-8") as file:
+                    recent_files = json.load(file)
+            except json.decoder.JSONDecodeError as err:
+                LOG.error("Invalid recent files json %s: %s", self.json_file_path, err)
+                recent_files = []
+        else:
+            recent_files = []
```
oerc0122: Could reduce duplication.

```suggestion
        recent_files = []

        if self.json_file_path.is_file():
            try:
                with self.json_file_path.open(encoding="utf-8") as file:
                    recent_files = json.load(file)
            except json.decoder.JSONDecodeError as err:
                LOG.error("Invalid recent files json %s: %s", self.json_file_path, err)
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlotDataModel.py
```diff
@@ -174,17 +173,17 @@ class PlotDataModel(QStandardItemModel):
...
         self._next_number = 0
-        if not self.DEFAULT_JSON_PATH.exists():
-            with self.DEFAULT_JSON_PATH.open("w", encoding="utf-8") as file:
-                json.dump([self.PLACEHOLDER_STRING], file, indent=4)
-        self.recent_files_update.connect(self.store_mda_file)
+        self.recent_files = RecentFiles(
+            self.DEFAULT_JSON_PATH,
+            self.MAX_NUMBER_RECENT_FILES,
+            self.PLACEHOLDER_STRING,
+        )
```
oerc0122: This should be on the class, not the instance, i.e. before the init with the signals.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -144,10 +110,11 @@ def __init__(self, parent: QObject = None):
...
         self._next_number = itertools.count()
-        if not self.DEFAULT_JSON_PATH.exists():
-            with self.DEFAULT_JSON_PATH.open("w", encoding="utf-8") as file:
-                json.dump([self.PLACEHOLDER_STRING], file, indent=4)
-        self.recent_files_update.connect(self.store_mdt_file)
+        self.recent_files = RecentFiles(
+            self.DEFAULT_JSON_PATH,
+            self.MAX_NUMBER_RECENT_FILES,
+            self.PLACEHOLDER_STRING,
+        )
```
oerc0122: Again, this should be a ClassVar.

oerc0122 (commented):
Looks good, just an optional thing and some duplication reduction.

oerc0122 (approved):
Looks ok to me.

PR #1056, opened 2025-11-04
Add block_signals context manager to manage signals

oerc0122:
**Description of work**
Add context manager to safely block signals and always re-enable.

**Fixes**
N/A

**To test**
GUI should work as normal.


MBartkowiakSTFC (approved):

PR #1057, opened 2025-11-07
GUI selection undo update

### LINKED ISSUE #1020: [BUG] Selection in progress and the run button
**Description of the error**
When you have a selection in progress and have not confirmed the selection either by closing the helper or hitting the confirm selection button then the configurator is not updated. This means that apparently invalid selection can be used to run the job and the run button can be pressed but doesn't run any job.

**Suggested fix**
Stop the user from leaving the helper windows when a manual selection is underway.

**Additional details**

Image

ChiCheng45:
**Description of work**
The 3D view atom selection has been updated. After #1043 the 3D viewer selection undo button is now redundant since it also undos the entire 3D view selection. In this PR I have changed it so that the 3D view undo, undoes each atom selection at a time. #1020 Is fixed by confirming the 3D viewer selection whenever the user selects another window.

**Fixes**
Fixes #1020

**To test**
Test the 3D atom selection. The manual selection should be confirmed when the atom selection window deactivates. Check that the 3D view gui buttons confirm and undo grey out correctly.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/AtomSelectionWidget.py
@@ -550,6 +550,7 @@ class SelectionHelper(QDialog):
         if len(self.selection_model._clicked_atoms) == 1:
             self.disable_gui_selection_buttons()
         self.selection_model.atom_clicked_undo()
+        self.update_selection_textbox()
 
     @Slot()
     def confirm_manual_selection(self):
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
It's almost there. Two things:
1. When I undo the manual selection step by step, the 3D view selection gets updated, but the text box in the bottom-right corner does not. If I click on two atoms and undo them both, the text box still shows two atoms to be selected.
2. I would like different text to appear on the buttons. Should we have something like "Undo last manual pick", "Undo last click", "Undo one click", "Undo one atom"? (Also should the main undo/redo buttons have text like "Undo row"/"Redo row"?)

ChiCheng45:
> It's almost there. Two things:
> 
> 1. When I undo the manual selection step by step, the 3D view selection gets updated, but the text box in the bottom-right corner does not. If I click on two atoms and undo them both, the text box still shows two atoms to be selected.
> 2. I would like different text to appear on the buttons. Should we have something like "Undo last manual pick", "Undo last click", "Undo one click", "Undo one atom"? (Also should the main undo/redo buttons have text like "Undo row"/"Redo row"?)

Thanks, both these should be fixed now. I went with Undo previous action and for gui and for the undo/redo button to undo selection/redo selection.

MBartkowiakSTFC (approved):
I think this is all fine now. Looks good.

PR #1058, opened 2025-11-07
Refactor plotters

oerc0122:
**Description of work**
- Refactor all the plotters to use generator constructs, gridspecs and attempt to reduce code-duplication.
- Adds new "grouped" plot type with one plot per dataset.

**Fixes**
Prevent number of plots corresponding to number of datasets.

**To test**
Try to break plotting with weird sequences of button presses.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
@@ -33,11 +33,11 @@ from MDANSE_GUI.Tabs.Models.PlottingContext import PlotArgs
 from MDANSE_GUI.Tabs.Plotters.Plotter import Plotter
 
 if TYPE_CHECKING:
-    import numpy.typing as npt
     from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as Toolbar
     from matplotlib.figure import Figure
     from matplotlib.image import AxesImage
 
+    from MDANSE.util_types import FloatArray
     from MDANSE_GUI.Tabs.Models.PlottingContext import PlottingContext
 
 
@@ -48,17 +48,19 @@ class Heatmap(Plotter):
     @dataclass
     class BackupInfo:
         ind: int
-        image: AxesImage | None = None
-        array: np.ndarray = field(default_factory=lambda: np.empty((0,), dtype=float))
+        image: AxesImage = None
+        array: FloatArray = field(default_factory=lambda: np.empty((0,), dtype=float))
         minmax: tuple[float, float] = (-np.inf, np.inf)
         limits: tuple[float, float, float, float] = (-np.inf, np.inf, -np.inf, np.inf)
         interp: interp1d = None
+        title: str = ""
 
     def __init__(self) -> None:
         """Initialise all plotting parameters to defaults."""
         super().__init__()
         self._figure = None
-        self._backup: dict[int, Heatmap.BackupInfo] = {}
+        self._axes_titles: list[str] = []
+        self._backup: dict[str, Heatmap.BackupInfo] = {}
         self._current_x_axes = []
         self._initial_values = [0.0, 100.0]
         self._slider_values = [0.0, 100.0]
@@ -116,9 +118,8 @@ class Heatmap(Plotter):
     def handle_slider(self, new_value: list[float]):
         """Adjust colormap values based on slider values."""
         super().handle_slider(new_value)
-        target = self._figure
 
-        if target is None or new_value[1] <= new_value[0]:
+        if self._figure is None or new_value[1] <= new_value[0]:
             return
 
         self._slider_values = [new_value[0], new_value[1]]
@@ -153,7 +154,19 @@ class Heatmap(Plotter):
                 else:
                     self._figure.canvas.draw_idle()
                     backup.minmax = (newmin, newmax)
-        target.canvas.draw()
+        self._figure.canvas.draw()
+
+    def toggle_legend(self, enabled: bool) -> None:
+        if self._figure is None:
+            return
+
+        for axes, title in zip(self._axes, self._axes_titles, strict=True):
+            axes.set_title(
+                title if enabled else "",
+                fontsize=self.title_fontsize(title),
+            )
+
+        self._figure.canvas.draw()
 
     def check_curve_lengths(self):
         """Find the maximum number of elements in the x axes of the plot data."""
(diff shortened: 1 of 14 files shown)


MBartkowiakSTFC:
I have been testing the plotter. What I found so far is:
1. `Grid` plots of 1D datasets have an empty legend.
2. If I plot into `Grid` or `Single` with legend switched off, and switch it on afterwards, I get an exception: <class 'AttributeError'> 'NoneType' object has no attribute 'set_visible'
3. Heatmap plot is labelling the plot axes wrong. In my test, time was shown as Q and Q was shown as time.
Screenshot 2025-11-10 at 11 51 59

4. Heatmap plots start off without any legend. If I switch legend off and on, the legend appears.
5. If I try to plot a vector summary from the plot creator like this

Screenshot 2025-11-10 at 12 07 03

the plot type ends up being `Single` and not `Vectors`:

Screenshot 2025-11-10 at 12 06 42

While we do not want to allow `Vectors` plot type for normal data sets, it should still be possible to use `Vectors` for vectors.

6. For `Single` and `Heatmap` plots, switching the legend off and on makes all the subplots resize.
Before:
Screenshot 2025-11-10 at 12 18 02

After:
Screenshot 2025-11-10 at 12 18 12

MBartkowiakSTFC:
I spent some time plotting data using this branch. A lot of it works already.
I think that these points would need to be looked at:
1. I managed to make the Heatmap plot fail by changing the main axis of one of the 3 subplots:
Screenshot 2026-06-09 at 11 19 39
Similarly, unchecking the "Use it" checkbox for one of the datasets makes the entire `Heatmap` (or `Grouped`) plot fail.

2. When I got the Heatmap plot of DISF to appear using the Q vector as the main axis, the plot appears reversed. The signal should become broader in energy at large Q values, but here it is the other way around:
Screenshot 2026-06-09 at 11 23 25

3. In `Single` plots, the legend for each curve used to include information on the position of each curve on the other axis. That is, this DISF plot would normally have a legend telling me that these curves correspond to different Q values:
Screenshot 2026-06-09 at 11 27 01
`Grid` and `Grouped` plots in this branch do it correctly and show the q position of each cut.

4. After switching the main axis there and back (to Q and to energy again) and going through `Single`, `Grid`, `Grouped` plotters, when I go back to `Heatmap` I get a plot where both axes are the energy axis:
Screenshot 2026-06-09 at 11 31 15

MBartkowiakSTFC (approved):
I think it would make sense to merge this PR once the linter is happy with it.

Normalisation of heatmap plots will require more work and will be best dealt with in a separate PR.
The only other detail is that the first plot (available vs. selected vectors per shell) in vector statistics is now a histogram (and it used to show a point per shell). However, the values are still correct. We can always decide at a later time which plotting style works the best.

ChiCheng45:
Some issues with the plotter, 3D plotter, and resizing.

Animation43

MBartkowiakSTFC:
> Some issues with the plotter, 3D plotter, and resizing.
> 

OK, you're right. It can happen on my computer too, if I shrink the GUI window enough.

In case it helps, my first guess is that it has something to do with the line `self._figure.set_layout_engine("tight")` which has been commented out in this PR.

oerc0122:
This issue is actually due to the new length of the legend. I am going to also make long legend labels collapse with a `...` which seems reasonable.

MBartkowiakSTFC:
> This issue is actually due to the new length of the legend. I am going to also make long legend labels collapse with a `...` which seems reasonable.

This may be the case. Still, you can create a legend larger than the plot itself also in the protos branch, and there the plot does not get resized.

On my machine I found that the minimal change to this branch that eliminates this specific problem is to remove the layout keyword argument from PlotWidget line 384
`mpl.figure(layout="constrained")`
and just create a default figure instead
`mpl.figure()`
But this leads to other problems later when switching to Heatmap and back, since Heatmap changes the matplotlib layout engine at the moment. There could be other consequences in other plot types too...

ChiCheng45:
On the single plot, the number of plots is limited to a small number of line plots per dataset. Can you also do this from the grouped plots? There are some issues if you have a lot of line plots and you use the Y offset.

Here I have 441 line plots. When Y offset is used, it becomes quite messy.

Animation44

Also, on the grouped plot, the home button doesn't work as expected compared to single. On single plot, the home button resets the plot so it fills the space; this doesn't always happen on grouped.


Animation45

ChiCheng45:
Looks like this fixes the rotation issue in #1267.

image

image

image

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
```diff
@@ -419,3 +336,42 @@ def _get_datasets(axis: Axes) -> Iterator[AxesImage]:
...
+        colour: str,
+    ) -> AxesImage:
+        dataset = databundle.dataset
+
+        x_label, y_label = axis_labels
+
+        x_axis = dataset.x_axis(x_label)
+        y_axis = dataset.x_axis(y_label)
+
+        limits = (x_axis[0], x_axis[-1], y_axis[0], y_axis[-1])
```
ChiCheng45: Can we change these limits to

```python
        dx = (x_axis[-1] - x_axis[0]) / (len(x_axis) - 1)
        dy = (y_axis[-1] - y_axis[0]) / (len(y_axis) - 1)

        limits = (
            x_axis[0] - dx / 2, x_axis[-1] + dx / 2, 
            y_axis[0] - dy / 2, y_axis[-1] + dy / 2
        )
```

That will pad it out by half the step size on the edges, ensure the pixel is centred correctly. This will fix the second issue of #1267.

Image

Image

ChiCheng45 (commented):
Just a small styling suggestion. Can you add a a borders around the bar chart bars in the available vectors chart.

Image

ChiCheng45 (commented):
Bar chart look great thanks. One more small bug I found, Its a bit complicated to get occur.

1. Load multiple datasets and open in single plot.
2. On one of the dataset set use it? to something invalid e.g. 99999.
3. Switch to the grouped plot. Now the plot is stuck and no amount of changes to the use it? option fixes it. You have to switch away from Grouped and fix use it and then switch back.

Image

ChiCheng45:
Thanks for the fixes. Another thing.

The normalising data option causes the offset to be reset back to zero.

Animation51

ChiCheng45:
Can you also update the plotter so it gives the error after the plot type change instead of showing the previous plot?

So, in the below, right after I changed it to heatmap, can you make it show the error plot?

Animation53

ChiCheng45:
Also, for the vector |q| statistics, the plot is empty when there is only one or two shells.

image

image

MBartkowiakSTFC:
Now the vector summary plot shows values in "Available vectors", but are the labels swapped?

In the vector summary, I get this:
automatic_vector_plot

When I plot the "Available vectors" datasets manually, I get:
manual_plot_vectors

BTW you are welcome to suggest some better labels for the two datasets.

MBartkowiakSTFC:
I think I found one more problem:

When using the `Grouped` plotter, I can deselect one of the datasets, and another plot ends up being empty.

Before:
Screenshot 2026-08-13 at 14 35 51

After:
Screenshot 2026-08-13 at 14 35 59

oerc0122:
> I think I found one more problem:
> 
> When using the `Grouped` plotter, I can deselect one of the datasets, and another plot ends up being empty.
> 
> Before: img
> After: img

Odd, what dataset was this with?

MBartkowiakSTFC:
> > I think I found one more problem:
> > When using the `Grouped` plotter, I can deselect one of the datasets, and another plot ends up being empty.
> > Before: img
> > After: img
> 
> Odd, what dataset was this with?

I think what really happens is that the scale of the subplot is fixed to that from another dataset, and it cannot be reset. So, you should be able to reproduce it with any datasets that appear on a different scale. Try plotting mean square displacement and temperature at the same time, then switch one of them off.

MBartkowiakSTFC:
Plots now look fine, thanks!

Hopefully the last thing, this time in Text Viewer: if I send a 2D data array (like the `s(q,f)` dataset from DISF) to the text viewer instead of a plot viewer, the preview looks fine. If I uncheck the box for "Preview only (truncate)", it fails to show the data and triggers an error:
`EXCEPTION: <class 'ValueError'> zip() argument 2 is longer than argument 1 <traceback object at 0x171ecedc0>`

MBartkowiakSTFC (approved):
I think this is fine to be merged.

PR #1059, opened 2025-11-07
Add save plot data button

### LINKED ISSUE #765: [ENHANCEMENT] Save data from the plot to a text file.
Add the ability to save plot data from the plotter to a text file. 

This will be useful as the plotter allows for a few different options to modify the plot results from the results in the MDA file. After modifying the results, the user may want to save them to a text file to process them in another program, for example.

oerc0122:
**Description of work**
Adds button to save plot data to file.

**Fixes**
Fixes #765 

**To test**
Use save button. There are options to put `%axis%` and `%line%` to save each graph and each line to separate files.

### DIFF
--- a/Doc/pages/H_Plotting.rst
+++ b/Doc/pages/H_Plotting.rst
@@ -170,6 +170,37 @@ Changing the main axis will result in an updated plot
 
    The same 2D array is now plotted against the energy axis.
 
+Saving the Data
+~~~~~~~~~~~~~~~
+
+Basic saving
+------------
+
+It is possible to save the data as they are presented (including
+modifications, shifts and normalisations) to file for further
+processing. These data are saved in an annotated ``.csv`` format for
+transferability and ease of use.
+
+Each data block (axis, line) starts with a commented header-line
+(using ``#`` as a comment marker) detailing the contents of the block.
+
+.. note::
+
+   Because of potentially mismatched axes (e.g. due to numerical
+   rounding) it is not possible to stack multiple lines side-by-side in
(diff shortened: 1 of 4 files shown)


MBartkowiakSTFC (commented):
I did not manage to find the button which saves the plot data to a file.

MBartkowiakSTFC:
It seems to me that the heatmap plot does not get saved correctly. I took a dynamic coherent structure factor calculation, s(q,omega) results, with energy axis from -2040 meV to 2040 meV, and Q from 1 to 2 inverse Angstroms. The resulting file had the energy values from 1 to -2040 and Q from 2 to 2040. So, the limits are probably swapped.

Minor detail 1: I was able to save an empty file from a plotting tab with no plot, from a plot with no datasets, and from a heatmap plot of a 1D dataset. Would it be better to disable the save button when we can expect the resulting file to be empty? That is, if that's easy to detect.

Minor detail 2: Could we extend the text of the save button, and maybe add a tooltip? Just to make it clear to the user that the "Save" button will not save an image, but a text file.

MBartkowiakSTFC:
Now my S(q,omega) heatmap has the correct limits for Q and omega, but it gets written out with 1000 steps in Q and 33 steps in energy. It should have been 33 steps in Q and 1000 in energy.

MBartkowiakSTFC (approved):
I think this works fine now.

MBartkowiakSTFC:
For the record, I did some more testing of Single and Heatmap plots with and without normalisation. Everything worked as expected.

PR #1060, opened 2025-11-10
Update the molecule finder code

MBartkowiakSTFC:
**Description of work**
In `TrajectoryEditor`, the molecule finding code has been becoming increasingly outdated. This PR is meant to make it more compatible with the other analysis jobs.

**Fixes**
1. Updated `Connectivity` docstrings.
2. The same code is now used for 3D viewer bond finding and TrajectoryEditor bond finding (except the TrajectoryEditor uses periodic boundary conditions).
3. `Trajectory` object is accessed directly in the job and not through the `configurator`.

**To test**
The results should be the same as before or better. (Molecules that were previously fragmented are more likely to be recognised as whole molecules). This method of finding bonds should be faster than the previous one.


ChiCheng45 (approved):
Looks good to me.

PR #1062, opened 2025-11-12
Fixed issues with GUI with system dark or light mode changes

### LINKED ISSUE #1038: [BUG] Changing dark and light mode while MDANSE GUI is running
**Description of the error**
When the user switches from dark to light or light to dark mode then the icon are left in their original dark or light mode states. This also can happen when a user has auto dark or light mode which can switch between the two modes depending on the time of day.

Icons should change to dark or light mode when changed.

**Additional details**
Image

ChiCheng45:
**Description of work**
Fixed so that MDANSE GUI so that it reacts correctly to dark or light mode changes by the system. 

mode fix example

**Fixes**
Fixes #1038
Correlation frames widget tooltip description.

**To test**
Check that the GUI react correctly to dark or light mode system changes. Check a job which uses frames or correlation frames works as usual.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/CorrelationFramesWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/CorrelationFramesWidget.py
@@ -24,13 +24,15 @@ from .FramesWidget import FramesWidget
 
 
 class CorrelationFramesWidget(FramesWidget):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+
+    def build_fields(self):
+        super().build_fields()
 
         if self._last_frame <= 0:
             c_frames = "1"
         else:
             c_frames = str(math.ceil(self._last_frame * 0.5))
+
         field = QLineEdit(c_frames, self._base)
         validator = QIntValidator(field)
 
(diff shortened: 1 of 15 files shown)


MBartkowiakSTFC:
I am still testing it and it works great.

One thing which is a bit unexpected is that now in dark mode the combo box text changes colour when the window loses focus. That is, when MDANSE window is the active one, everything is consistent:
Screenshot 2025-11-12 at 17 30 38

But if I change focus to another window, while the MDANSE window is still visible, the text in the combo boxes goes dark:
Screenshot 2025-11-12 at 17 30 43

This is probably not important, since the text colour is correct when it matters, but I was curious if you can reproduce this.

ChiCheng45:
> I am still testing it and it works great.
> 
> One thing which is a bit unexpected is that now in dark mode the combo box text changes colour when the window loses focus. That is, when MDANSE window is the active one, everything is consistent: Screenshot 2025-11-12 at 17 30 38
> 
> But if I change focus to another window, while the MDANSE window is still visible, the text in the combo boxes goes dark: Screenshot 2025-11-12 at 17 30 43
> 
> This is probably not important, since the text colour is correct when it matters, but I was curious if you can reproduce this.

I think this is something to do with the fusion style rather than MDANSE. This will happen if you build a Qt app with a single combobox.

combo

MBartkowiakSTFC:
Screenshot 2025-11-13 at 10 16 13
Interestingly enough, in the plot tab there is one section where the combo boxes are not affected.

MBartkowiakSTFC (approved):
The appearance of the GUI is definitely better now. I did not find any problems.

ChiCheng45:
> Screenshot 2025-11-13 at 10 16 13 Interestingly enough, in the plot tab there is one section where the combo boxes are not affected.

Yea its stange. Another is that it doesn't happen on Windows.

PR #1070, opened 2025-11-24
Fix issue with GUI tests

oerc0122:
**Description of work**
GUI tests previously weren't running due to switch to disable loading widgets if no trajectory present.

**Fixes**
Forces GUI to load small trajectory before selecting GUI button.

**To test**
GUI tests should detect when invalid widgets in Job definitions.

Running `pytest -rP` should display logs of actually generating GUI widgets.


MBartkowiakSTFC (approved):
I tried deliberately breaking some widgets, and it made some of the tests fail. Looks good.

PR #1071, opened 2025-11-24
Allow for full user defined filename

### LINKED ISSUE #986: [BUG] Wrong handling of output file names with commas

RobBuchananCompPhys:
**Description of work**
This permits the use of output filenames including '.' separators preceeding the .mda extension (i.e. 'PositionPowerSpectrum.result1.mda').

**Fixes**
Closes #986

**To test**

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
+++ b/MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
@@ -181,7 +181,8 @@ class HDFFormat(IFormat):
             outputFile = h5py.File.in_memory()
 
         else:
-            filename = Path(filename).with_suffix(extension)
+            filepath = Path(filename)
+            filename = filepath.with_name(filepath.name + extension)
 
             # The HDF output file is opened for writing.
             PLATFORM.create_directory(filename.parent)
--- a/MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
+++ b/MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
@@ -182,7 +182,10 @@ class HDFFormat(IFormat):
 
         else:
             filepath = Path(filename)
-            filename = filepath.with_name(filepath.name + extension)
+            if "." in filepath:
+                filename = filepath.with_name(filepath.name + extension)
+            else:
+                filename = filepath.with_suffix(extension)
 
             # The HDF output file is opened for writing.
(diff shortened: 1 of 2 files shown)


MBartkowiakSTFC:
I think that the new code works around the problem of `.` characters in the file name, but fails to recognise the cases where the extension is already at the end of the file name, and adds a second one. So, in some tests the output filename is given as `output_filename.mda`, but the code ends up writing `output_filename.mda.mda`.

MBartkowiakSTFC:
When I run jobs from the GUI, the converter jobs are fine and the output trajectory names are handled correctly. The analysis jobs still have a conflict where `.mda` gets appended again to a name that already ends with `.mda`. Then the GUI protects the name `output.mda` and tries to load `output.mda` which fails, because the real output file name was `output.mda.mda`.

MDANSE/Src/MDANSE/Framework/Formats/HDFFormat.py
```diff
@@ -181,7 +181,13 @@ def write(
             outputFile = h5py.File.in_memory()
 
         else:
-            filename = Path(filename).with_suffix(extension)
+            user_given_name = Path(filename).name
+            filepath = Path(filename).with_suffix(".mda")
```
MBartkowiakSTFC:
```suggestion
            filepath = Path(filename).with_suffix(extension)
```

Since this is `HDFFormat.py` and not `MDAFormat.py`, _normally_ it would make sense not to fix the suffix here to ".mda". I am pretty sure that your PR will work just as well if you use the line above, since it is the lines you added below that fix the problem.

Ultimately, the reason why the current version works correctly is because `MDTFormat.write` is never used (!). It seems that all the output filename handling is done by `OutputTrajectoryConfigurator` instead. I will create an issue to deal with it later.

MBartkowiakSTFC (approved):
I tried writing out different files with and without the extension, using names with and without the `.` character. I did not run into any problems. Looks good.

PR #1072, opened 2025-11-26
Replace SubclassFactory with a RegisterFactory

oerc0122:
**Description of work**
- Replace `SubclassFactory` with a classic `RegisterFactory` to avoid metaclass magic, allow effective ABCs and explicit inclusion (i.e. not including `Converter` as an `IJob` and thus making a non-instantiable object a key) and naming (including multiple names for the same thing). 
- Add `UCDict` class for case-insensitive dictionary keys.
- Add `Dummy` `IJob` child to account for `ABC` working properly.

PR is broken down since it is a large number of files touched (though small changes):
- First commit implements the factory and UCDict
- Middle commits affect each historic `SubclassFactory` 
- Last commit makes changes necessary for code to function

**NB.** Converters are registered in both IJob and Converter as was historical.

**Fixes**
Part of the motivation for this is to support a factory-like mechanism for GUI Widgets from Descriptor types.

**To test**
Standard tests should run.


MBartkowiakSTFC:
The changes did not affect anything major, as far as I can tell. However, the way the class names have been converted to uppercase characters has been interfering with combo boxes.

Screenshot 2025-11-28 at 13 40 50

So far, I have found the following problems:
1. The instrument definitions used to set the q vector generator type to "SphericalLatticeQVectors". As this has been replaced with "SPHERICALLATTICEQVECTORS", the combo box remains set to "CIRCULARLATTICEQVECTORS" when an instrument profile has been selected.
2. When a new plot is created, the plotter type used for the plot is "Single", but the combo box shows "GRID". Also, it is possible to switch the plotter type to "TEXT" or "VECTORS", which should normally be disabled.

MBartkowiakSTFC:
One problem: it seems that the instrument definition GUI cannot find any q-vector generators in this branch.
Screenshot 2026-06-02 at 09 37 23
The q-vector type combo box does not have any entries.

MBartkowiakSTFC:
One more thing: in `MDANSE_GUI/Tabs/Plotters/Vectors3D.py`, there is no `@Plotter.register("Vectors3D")`. This causes problems with vector distribution plots in the q vectors helper dialog.

MBartkowiakSTFC (approved):
I have not found any problems. Looks good.

ChiCheng45 (approved):
Looks good to me.

PR #1074, opened 2025-11-28
Fix separator handling in Text.py and DataWidget.py

### LINKED ISSUE #1073: [BUG] Saving to CSV files fails for 2D datasets
**Description of the error**
I have been unable to export my 2D array of S(q,w) to a text file using the text view in the plotter.

A csv file should be written, containing all the selected datasets.

An empty file is created. Log output:
`ERROR - process[41913] - DataWidget 199 - Could not write data file (new_results.csv): DatasetFormatter.process_data() got an unexpected keyword argument 'separator'`

**Suggested fix**
N/A

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Minor changes that allow us to write out .csv files.

closes #1073 

**Fixes**
1. Removed the `separator` keyword argument from methods which don't use it.
2. Passed the separator as `delimiter` to the CSV file writer.
3. Changed standard delimiter to `,`.
4. Limited the length of the separator input field to 1 character.

**To test**
Load a data file and try to save a dataset from the plotter text view. Check if the separator changes affect the output files.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Text.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Text.py
@@ -52,7 +52,7 @@ class DatasetFormatter:
         self._preview_columns = 10
         self._rounding_prec = 5
         self._comment = "#"
-        self._separator = " "
+        self._separator = ","
 
     def take_new_input(self, pc: PlottingContext):
         """Assign the input PlottingContext to the plotter.
@@ -97,7 +97,7 @@ class DatasetFormatter:
             return ["No data selected"]
 
         for databundle in self._plotting_context.datasets().values():
-            yield self.process_data(databundle.dataset, separator=self._separator)
+            yield self.process_data(databundle.dataset, main_axis=databundle.main_axis)
 
     def make_dataset_header(self, dataset: SingleDataset, comment_character="#"):
         """Return the dataset informartion as text.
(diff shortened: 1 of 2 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Text.py
```diff
@@ -180,8 +180,6 @@ def process_1D_data(
         ----------
         dataset : SingleDataset
             A SingleDataset read from an .MDA file (HDF5).
-        separator : str, optional
```
oerc0122, line 183: Why remove the doc here? Is it not still being passed?

oerc0122 (changes requested):

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Text.py
MBartkowiakSTFC: As far as I can tell, it is not used here, it is not handled by the method, and trying to pass the separator as a keyword argument to `process_data` was the thing causing the error that made saving CSV files impossible.

oerc0122: Oh, I misread the collapsed block.

oerc0122 (approved):

PR #1076, opened 2025-12-03
Improve output file widget

### LINKED ISSUE #1075: [BUG] Output file widgets fail on specific directory and file name combinations
**Description of the error**
The output file name generators in the GUI suggest output file names of this type:
`/path/to/input/file/JobName_resultN`
where N is a positive number. If a file with a given name and number already exists, the GUI will keep incrementing the number to find a file name that is not in the filesystem.

The way the GUI code splits strings, it is possible to use an input directory containing a string "result" (or "trajectory" for converters) which will make the output file widget fail to initialise if the default output file name is already taken.

This should work for any combination of input file name, job name and path on the disc.

Trying to run "MeanSquareDisplacement" on an input file `results/trajectory.mdt` will work for the first time. Once this run has finished, trying to pick "MeanSquareDisplacement" again will crash the output file widget.

**Suggested fix**
Implement a new filename generating function for both analysis results and trajectories.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
The code that generates standard output file names had been moved out of the GUI widgets. It now generates file names without string splitting operations.

closes #1075 

**Fixes**
1. A new function in `MDANSE.IO.IOUtils` generates file names for the GUI widgets, without splitting strings.

**To test**
Create testing directory named "trajectory_result" and copy a short trajectory there. Try converting it multiple times from the GUI and check that a new name is suggested correctly each time.
Try running an analysis on that trajectory. Again, check that a new output file name is suggested by the GUI if the output files already exist.

### DIFF
--- a/MDANSE/Src/MDANSE/IO/IOUtils.py
+++ b/MDANSE/Src/MDANSE/IO/IOUtils.py
@@ -25,9 +25,11 @@ from pathlib import Path
 from typing import Any
 
 import numpy as np
-from more_itertools import value_chain
+from more_itertools import first_true, value_chain
 
 from MDANSE.MLogging import LOG
+
+MAX_FILE_COUNT = 2048
 
 
 class UCEnum(Enum):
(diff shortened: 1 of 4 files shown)


IfyyAgu (approved):
Looks good. One question, why did you switch from Purepath to path? Is there a major difference?

MBartkowiakSTFC:
> Looks good. One question, why did you switch from Purepath to path? Is there a major difference?

`Path` can access the filesystem and check if a file `exists()`, whereas `PurePath` couldn't do this.

Issue #500, opened 2024-07-15
[ENHANCEMENT] Add packing/supercell option to the 3D viewer.


ChiCheng45:
**Is your feature request related to a problem? Please describe.**
Add an option to generate a supercell in the 3D viewer.

**Additional context**
Here is an example of how something similar is done in Mercury.
![image](https://github.com/user-attachments/assets/21bceca5-1ce9-42cf-b0e4-49993283a1b6)

PR #1078, opened 2025-12-13
MolecularViewer refactor and optimisations

ChiCheng45:
**Description of work**
Refactored the molecule viewer code.

Improved bond and atom trace calculation performance.

Changed the bonds setting so that it can be calculated from the first, last, or every frame, or it can be taken from the file; the default is every frame. 

Screenshot 2025-12-22 at 11 23 18

 Improved performance, larger systems should run faster. I think after this change the performances are probably good enough to be able implement #500 and not have it slow down too much. Here is the before and after on my old Mac with a 61K atom trajectory. 
 
Before
before

 After
after

 
**To test**
All aspects of the 3D viewer including atom selection, and atom tracing plotting. Test the different bond settings. 3D viewer should be much faster, test with a large trajectory.


MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
             add_arrow(np.eye(3)[i], matrix[i])
             add_text(label, matrix[i])
 
-    def _new_trajectory_object(self, fname: str, trajectory: Trajectory):
-        """Creates and sets a new trajectory reader for the input trajectory.
+    def change_axes(self, axes_option: str):
```
oerc0122: Possibly wants to be an `Enum`?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
-            ("center_sphere.stl", (0.6, 0.6, 0.6), 0.4, (0, 0, 0)),  # grey center
-            ("yellow_sphere.stl", (0.2, 0.85, 0.25), 0.4, (0, 0, 0)),
-            ("blue_sphere.stl", (0.2, 0.35, 0.85), 0.4, (0, 0, 0)),
-            ("red_sphere.stl", (0.85, 0.24, 0.2), 0.4, (0, 0, 0)),
-        ]
-
-        for filename, colour, position, scale in spheres:
-            actor = self.create_actor_for_placeholder_3d_model(
-                filename, colour, position, scale
+        if self.atom_label_type == "index":
```
oerc0122: Might want to be a `match` and/or `Enum`

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+            labels = list(range(self._reader._n_atoms))
+        elif self.atom_label_type == "label":
+            label_dict = self._reader._trajectory.chemical_system._labels
+            if not label_dict:
+                return
```
```suggestion
        elif self.atom_label_type == "label":
            if not label_dict := self._reader._trajectory.chemical_system._labels:
```

Could also be:

```suggestion
        elif self.atom_label_type == "label" and label_dict := self._reader._trajectory.chemical_system._labels:
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+            keys = more_itertools.run_length.decode(
+                ((k, len(v)) for k, v in label_dict.items())
+            )
+            labels = sorted(keys, key=label_dict.__getitem__)
+        elif self.atom_label_type == "atom":
+            labels = self._atoms
+        elif self.atom_label_type == "molecule":
+            label_dict = self._reader._trajectory.chemical_system._clusters
+            if not label_dict:
+                return
```
oerc0122: As above

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
 
-        self._renderer.AddActor(self._actors)
+        if self._current_coords is None or len(labels) != self._reader._n_atoms:
+            return
 
-        # side view of 3d model
-        self._camera.SetPosition(20, 0, 0)
-        self._camera.SetFocalPoint(0, 0, 0)
-        self._camera.SetViewUp(0, 0, 1)
+        atom_label_actors = []
```
oerc0122: Why bother with the temp?

```suggestion
        self.atom_label_actors = []
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+            self.atom_label_actors,
+            self._current_coords,
+            strict=True,
+        ):
+            follower.SetPosition(*coord)
 
-        The components of the 3D placeholder are rotated independently.
-        A second rotation by a smaller angle is applied to the rings to
-        create the impression of the spheres moving along their rings.
+    def change_atom_labels(self, label_option: str) -> None:
```
oerc0122: As above

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        if self._cell_visible and uc is not None:
+            # update the unit cell
+            a = uc.a_vector
+            b = uc.b_vector
+            c = uc.c_vector
+            uc_points = vtk.vtkPoints()
+            uc_points.SetNumberOfPoints(8)
+            for i, v in enumerate([[0, 0, 0], a, b, c, a + b, a + c, b + c, a + b + c]):
+                x, y, z = v
+                uc_points.SetPoint(i, x, y, z)
```
```suggestion
            for i, (x, y, z) in enumerate([[0, 0, 0], a, b, c, a + b, a + c, b + c, a + b + c]):
                uc_points.SetPoint(i, x, y, z)
```

Or


```suggestion
            for i, v in enumerate([[0, 0, 0], a, b, c, a + b, a + c, b + c, a + b + c]):
                uc_points.SetPoint(i, *v)
```

oerc0122: Also these might want to be a property of `UnitCell`?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+                (1, 4),
+                (1, 5),
+                (4, 7),
+                (2, 4),
+                (2, 6),
+                (5, 7),
+                (3, 5),
+                (3, 6),
+                (6, 7),
+            ]:
```
oerc0122, line 452: These might want to be a constant?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
-        self.create_atom_label_actors()
-        self._iren.GetRenderWindow().Render()
-        self._iren.Render()
+        self._renderer.RemoveActor(self.atom_actor)
+        self.atom_actor = None
 
-    def trace_from_dialog(self, params: dict[str, Any]):
-        """Passes the input parameter dictionary to the method
-        which draws an isosurface in the 3D view.
+    def create_atoms(self, ball_opacity: float = 1.0):
```
oerc0122: Do we just want these globally called `opacity`? Perhaps as kwarg-only?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+            VTK object storing the atom properties used in 3D view (colour, radius)
+        ball_opacity : float, optional
+            opacity (alpha) of atom spheres, by default 1.0
+
+        Returns
+        -------
+            Two vtk.vtkLODActor instances, for atoms
+
+        """
+        temp_radius = float(1.0 * self._scale_factor)
```
```suggestion
        temp_radius = float(self._scale_factor)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        sphere.SetCenter(0, 0, 0)
+        sphere.SetRadius(temp_radius)
+        sphere.SetThetaResolution(self._resolution)
+        sphere.SetPhiResolution(self._resolution)
+
+        glyph_mapper = vtk.vtkGlyph3DMapper()
+        glyph_mapper.SetInputData(polydata)
+        glyph_mapper.SetSourceConnection(sphere.GetOutputPort())
+
+        temp_scale = float(1.0 * self._scale_factor)
```
```suggestion
        temp_scale = float(self._scale_factor)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        line_mapper.SetInputData(polydata)
+        line_mapper.SetLookupTable(self._colour_manager._lut)
+        line_mapper.ScalarVisibilityOn()
+        line_mapper.ColorByArrayComponent("scalars", 1)
+        line_actor = vtk.vtkLODActor()
+        line_actor.GetProperty().SetLineWidth(3 * self._scale_factor)
+        line_actor.SetMapper(line_mapper)
+        line_actor.GetProperty().SetAmbient(0.2)
+        line_actor.GetProperty().SetDiffuse(0.5)
+        line_actor.GetProperty().SetSpecular(0.3)
```
oerc0122, line 605: Perhaps also args, but kwarg only?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        glyph_mapper.SetLookupTable(self._colour_manager._lut)
+        glyph_mapper.SetScalarRange(
+            polydata.GetPointData().GetArray("colours").GetRange()
+        )
+
+        ball_actor = vtk.vtkLODActor()
+        ball_actor.SetMapper(glyph_mapper)
+        ball_actor.GetProperty().SetAmbient(0.2)
+        ball_actor.GetProperty().SetDiffuse(0.5)
+        ball_actor.GetProperty().SetSpecular(0.3)
```
oerc0122, line 571: Perhaps kwarg only args with defaults?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+            if bonds_exist:
+                self._atm_polydata.SetLines(bonds)
+                return
+
+    def create_bond_cell_array(
+        self,
+        rs: np.ndarray,
+        covs: np.typing.NDArray[float],
+        not_du: np.typing.NDArray[bool],
+        tolerance: float = 0.04,
```
oerc0122, line 656: Kwarg only?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        vtk.vtkCellArray, bool
+            a VTK array of pairs of atom indices, and a flag True if some bonds were found
+        """
+        # determine and set bonds without PBC applied
+        bonds = vtk.vtkCellArray()
+
+        dist, js, ks, _ = distance_calculation(rs, 2 * np.max(covs) + tolerance)
+        sum_radii = (covs[js] + covs[ks] + tolerance) ** 2
+        js = js[(dist > 0) & (dist < sum_radii)]
+        ks = ks[(dist > 0) & (dist < sum_radii)]
```
oerc0122: I hope `dist > 0`?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        ls = not_du[js]
+        ms = not_du[ks]
+
+        n_points = len(ls)
+        idxs = np.zeros((len(ls), 3), dtype=np.int64)
```
```suggestion
        idxs = np.zeros((n_points, 3), dtype=np.int64)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        idxs[:, 0] = 2
+        idxs[:, 1] = ls
+        idxs[:, 2] = ms
+        bonds.SetCells(n_points, numpy_support.numpy_to_vtkIdTypeArray(idxs.flatten()))
+
+        return bonds, len(ls) > 0
```
```suggestion
        return bonds, n_points > 0
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        ----------
+        scale_factor : float
+            Sphere radii in 3D view will be multiplied by this factor
+        """
+        self._scale_factor = scale_factor
+        if self._reader is not None:
+            self.create_atoms()
+            self.update_renderer()
+
+    def _new_visibility(self, flags: list[bool]):
```
oerc0122: Could do a Flag Enum if desired

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        self.create_bonds()
+        self.update_renderer()
+
+    def trace_from_dialog(self, params: dict[str, Any]):
+        """Passes the input parameter dictionary to the method
+        which draws an isosurface in the 3D view.
+
+        Parameters
+        ----------
+        params : Dict[str, Any]
```
```suggestion
        params : dict[str, Any]
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+            surface = self._surfaces[trace_number]
+        except IndexError:
+            return
+        else:
+            surface.VisibilityOff()
+            surface.ReleaseGraphicsResources(self._iren.GetRenderWindow())
+            self._renderer.RemoveActor(surface)
+            self._surfaces.pop(trace_number)
+            self.update_renderer()
+            self.changed_trace.emit()
```
oerc0122: Don't need `else` after `return`
```suggestion
        surface.VisibilityOff()
        surface.ReleaseGraphicsResources(self._iren.GetRenderWindow())
        self._renderer.RemoveActor(surface)
        self._surfaces.pop(trace_number)
        self.update_renderer()
        self.changed_trace.emit()
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,270 +252,485 @@ def add_text(text, coord):
...
+        LOG.info(f"Computing isosurface of atom {index}")
+        if params is None:
+            params = copy.copy(TRACE_PARAMETERS)
+
+        traj_sampling = params.get("traj_samples", 1000)
+        smearing_factor = params.get("smearing_factor", 1)
+        grid_step = params.get("grid_sampling", 0.02)
+        rgb = params.get("surface_colour", (0, 0.5, 0.75))
+        opacity = params.get("surface_opacity", 0.5)
+        trace_isovalue = params.get("trace_isovalue", 0.5)
```
oerc0122: Does `TRACE_PARAMETERS` contain these defaults? Could use a `collections.ChainMap`

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -615,247 +805,37 @@ def _draw_isosurface(self, index: int, params: dict[str, Any] | None = None):
...
-            strict=True,
-        ):
-            follower.SetPosition(*coord)
+        self._renderer.AddActor(new_surface)
 
-    def clear_atom_labels(self):
-        """Clears the atoms labels."""
-        if not self.atom_label_actors:
-            return
+        new_surface.SetPosition(lower_limit[0], lower_limit[1], lower_limit[2])
```
```suggestion
        new_surface.SetPosition(*lower_limit)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -615,247 +805,37 @@ def _draw_isosurface(self, index: int, params: dict[str, Any] | None = None):
...
+        self._surfaces.append(new_surface)
+        self._isocontours.append(new_isocontour)
 
-        for actor in self.atom_label_actors:
-            self._label_renderer.RemoveActor(actor)
+        self.update_renderer()
 
-        self.atom_label_actors = []
+        LOG.info(f"Finished calculating the trace of atom {index}")
```
```suggestion
        LOG.info(f"Finished calculating the trace of atom %d", index)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerExtended.py
```diff
@@ -0,0 +1,168 @@
...
+
+        picked_actor = picker.GetActor()
+        if picked_actor is None:
+            return
+
+        picked_pos = np.array(picker.GetPickPosition())
+        _, picked_index = KDTree(self._current_coords).query(picked_pos)
+
+        if picked_index < 0 or picked_index >= self._n_atoms:
+            return
```
oerc0122:
```suggestion
        if not 0 <= picked_index < self._n_atoms:
            return
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self.clicked_atom_index.emit(idx)
+        self.pick_atom(idx)
+        self.picked_atoms_changed.emit(self.picked_atoms)
+
+    def pick_atom(self, picked_atom):
+        if picked_atom in self.picked_atoms:
+            self.picked_atoms.remove(picked_atom)
+        else:
+            self.picked_atoms.add(picked_atom)
```
oerc0122: Would this be equivalent to `self.picked_atoms.symmetric_difference(other)`?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: This should be to ignore cases where dist == 0, so it does not bond itself.

ChiCheng45: Actually, it looks like this shouldn't be possible anymore

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: Do you mean to put this

```python
CUBE_EDGE_PONTS = [
                (0, 1),
                (0, 2),
                (0, 3),
                (1, 4),
                (1, 5),
                (4, 7),
                (2, 4),
                (2, 6),
                (5, 7),
                (3, 5),
                (3, 6),
                (6, 7),
            ]
```

at the top-level and then have

```python
            for i, j in CUBE_EDGE_PONTS:
```

Not sure if this is better since `CUBE_EDGE_PONTS` is currently only used in `update_uc_polydata`.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: Could do, maybe we can do this when we need it? Currently, there isn't anything in MDANSE that needs to change these.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: I replaced it with enums. I've left it to use if/else, as it didn't seem drastically cleaner to use match.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: I've changed the visibility flags, so there are only two flags now. Do you think its still worth doing?

ChiCheng45:
@MBartkowiakSTFC any preference on the default choice for the bond calculation setting? I think it should be either first frame or every frame.

MDANSE/Src/MDANSE/MolecularDynamics/Connectivity.py
```diff
@@ -55,19 +55,10 @@ def distance_calculation(
...
-        n_idxs = len(idxs)
-        if n_idxs == 0:
-            continue
-        js[start : start + n_idxs] = i
-        ks[start : start + n_idxs] = idxs
-        start += n_idxs
-
+    n_contacts = np.fromiter((len(i) for i in contacts), dtype=int)
+    mask = n_contacts > 0
+    js = np.repeat(np.nonzero(mask)[0], n_contacts[mask])
```
oerc0122, line 64: Is this essentially a `more_itertools.run_length.decode`? Not quite following what's being constructed here.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -53,16 +53,34 @@ def array_to_3d_imagedata(data: np.ndarray, spacing: tuple[float, float, float])
...
-        image.SetNumberOfScalarComponents(1)
-    else:
-        image.AllocateScalars(vtk.VTK_DOUBLE, 1)
-
+    image.AllocateScalars(vtk.VTK_DOUBLE, 1)
     image.GetPointData().SetScalars(numpy_support.numpy_to_vtk(data.ravel("F")))
     return image
 
 
+class AtomLabelType(Enum):
```
oerc0122, line 60: `StrEnum` is 3.11+. That may be useful when 3.11 is standard.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -205,17 +205,28 @@ def clear_axes(self):
...
+        """Clears and creates a new set of axes actors and adds
+        them to the renderer.
+        """
 
-            vtk_matrix = vtk.vtkMatrix4x4()
-            for j in range(3):
-                for k in range(3):
-                    vtk_matrix.SetElement(j, k, rot[j, k])
-            vtk_matrix.SetElement(3, 3, 1.0)
+        def add_arrow(color: list[float | int], direction: list[float | int]):
```
oerc0122: `int` is a strict subtype of `float` so can be omitted. Also the size is relevant, so we probably want:
```suggestion
ThreeVector = tuple[float, float, float]

        def add_arrow(color: ThreeVector, direction: ThreeVector):
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -244,10 +264,10 @@ def add_text(text, coord):
 
         self.clear_axes()
 
-        if self.current_axes_type == "none":
+        if self.axes_type == AxesType.NONE:
```
oerc0122: Can use `is` for `Enum`s. Alternatively we might want a `match`?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -262,10 +282,10 @@ def add_text(text, coord):
         if uc is None:
             return
```
oerc0122, line 282: Do we want to silently return or should these error?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,245 +294,430 @@ def add_text(text, coord):
...
+    def create_atom_labels(self):
+        """Clear and create atom label actor, setting the text to the
+        one chosen by atom_label_type, and add it to the label renderer.
+        """
+        self.clear_atom_labels()
+
+        if self._reader is None:
+            return
+
+        if self.atom_label_type == AtomLabelType.INDEX:
```
oerc0122: Might want to become a `match`? (Single checked var and simple comparisons, the `label_dict`s can be implemented as guards)

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -274,245 +294,430 @@ def add_text(text, coord):
...
+            return
+
+        if self._atoms_visible or self.bond_calc != BondCalc.NONE:
+            atoms = vtk.vtkPoints()
+            atoms.SetData(numpy_support.numpy_to_vtk(self._current_coords))
+            self._atm_polydata.SetPoints(atoms)
+
+    def change_atm_polydata_lines(self):
+        """Calculate and/or updates the atom polydata bonds."""
+        if self.bond_calc == BondCalc.EVERY:
```
oerc0122: Could become a `match`

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -563,18 +771,10 @@ def _draw_isosurface(self, index: int, params: dict[str, Any] | None = None):
...
-            n_idxs = len(idxs)
-            if n_idxs == 0:
-                continue
-            js[start : start + n_idxs] = i
-            ks[start : start + n_idxs] = idxs
-            start += n_idxs
+        n_contacts = np.fromiter((len(i) for i in contacts), dtype=int)
+        mask = n_contacts > 0
+        js = np.repeat(np.nonzero(mask)[0], n_contacts[mask])
+        ks = np.concatenate([contacts[i] for i in np.nonzero(mask)[0]])
```
oerc0122, line 798: Given this is repeated, should this be a function?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -1211,258 +1030,88 @@ def set_reader(self, reader):
...
-        if self._atoms_visible:
-            self._picking_domain = ball_actor
-            actors.append(ball_actor)
-            actors.append(picked_ball_actor)
-        else:
-            self._picking_domain = None
-        return actors
+        # set everything to some empty/zero value
+        self._n_atoms = 0
+        self._n_frames = 0
```
oerc0122, line 1057: Should the `clear`s not do this? Risk of desync?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerExtended.py
```diff
@@ -0,0 +1,183 @@
...
+        ) / np.linalg.norm(x_21)
+        selected = dist_to_line < scaled_radii
+        dist_x_21 = np.dot(coords - x_1, x_21)
+
+        if not np.any(selected):
+            return
+
+        picked_index = idxs[selected][np.argsort(dist_x_21[selected])][0]
+        self.clicked_atom_index.emit(picked_index)
+        LOG.debug(f"Click event picked up atom index {picked_index}")
```
oerc0122:
```suggestion
        LOG.debug(f"Click event picked up atom index %d", picked_index)
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
```diff
@@ -0,0 +1,315 @@
...
+
+        actor = self.create_bond_actor(self._picked_polydata, opacity=opacity)
+        self._renderer.AddActor(actor)
+        self.picked_bond_actor = actor
+
+    def update_picked_polydata(self):
+        """Updates the picked polydata coordinates."""
+        atoms = vtk.vtkPoints()
+
+        if len(self.picked_atoms) == 0:
```
oerc0122: I've been having a bit of a think and wonder if here (and in several of the previous classes) we want an `@property` (e.g. `n_picked_atoms`) to get the len of these where relevant?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
```diff
@@ -0,0 +1,315 @@
...
+    def change_picked_polydata_lines(self):
+        """Calculate and/or updates the picked polydata bonds."""
+        if len(self.picked_atoms) <= 1:
+            self._picked_polydata.SetLines(vtk.vtkCellArray())
+            return
+
+        picked = np.array(sorted(self.picked_atoms))
+        not_du = np.arange(len(self.picked_atoms))[self.du_log[picked]]
+
+        if self.bond_calc == BondCalc.EVERY:
```
oerc0122: Match?

oerc0122 (commented):
Looks good. Couple of stylistic and defensive things.

MBartkowiakSTFC:
> @MBartkowiakSTFC any preference on the default choice for the bond calculation setting? I think it should be either first frame or every frame.

I think that "from file" also has a fairly good claim to be the default option. But if you want to make sure that the users see the bonds immediately, then it should be "every frame", because it avoids glitches with bonds stretching across the entire box.

Side note: should we extend the labels to say something like "calculate every frame", so the users realise that this will be a slower option, as opposed to "read from file"? If it doesn't fit, you could always change the grid layout to be:
```
empty | atom checkbox | cell checkbox
axes | combo box spans two columns
labels | combo box spans two columns
bonds | combo box spans two columns 
```

MBartkowiakSTFC:
In this branch, manual selection is more tricky on my machine, since the atoms are pretty much invisible once I have deselected them:
Atoms selected:
Screenshot 2026-01-06 at 10 16 12
Some atoms deselected:
Screenshot 2026-01-06 at 10 16 23

In this case I could still manage to find the atoms by switching the bonds on. Do you think this needs improving?

MBartkowiakSTFC:
Overall, the performance is greatly improved, as demonstrated in my tests yesterday with 4,000,000 atoms in a trajectory.

However, on my computer the performance becomes bad even for small trajectories if I turn on the labels, e.g. atom indices.

ChiCheng45:
@MBartkowiakSTFC 

I've change the bond calc description, increased the unpicked opacity from 0.20 to 0.30, and reverted the text label to use followers again. Not sure why it was so slow on the Mac, the performance of the newer implementation of the atom labels seemed similar to the follower version on windows. Let me know if the opacity change is better or if I needs to be increased further.

MBartkowiakSTFC:
Thanks for the changes. The speed is back to normal when using labels. The opacity is better. Even though I find some atoms still hard to see, it can be easily fixed by changing the colour of the background.

I noticed one more problem that may need fixing.
1. I load a trajectory that does not contain **atom labels** or **molecule definitions**,
2. I start playing the animation,
3. It is possible to change different settings while playing the animation: cell/bonds/atoms visibility, axis type, etc.
4. If I set `labels:` to `label` or `molecule`, the atom positions stop updating in the 3D view, and every frame produces an error:
```
<class 'ValueError'>
zip() argument 2 is longer than argument 1
<traceback object at 0x314e3ee80>
  File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 1094, in set_coordinates
    self.update_atom_labels()
  File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 374, in update_atom_labels
    for follower, coord in zip(
ValueError: zip() argument 2 is longer than argument 1
```
At the same time, the animation progress bar is still moving and changing the frame numbers.

oerc0122:
> Thanks for the changes. The speed is back to normal when using labels. The opacity is better. Even though I find some atoms still hard to see, it can be easily fixed by changing the colour of the background.
> 
> I noticed one more problem that may need fixing.
> 
> 1. I load a trajectory that does not contain **atom labels** or **molecule definitions**,
> 2. I start playing the animation,
> 3. It is possible to change different settings while playing the animation: cell/bonds/atoms visibility, axis type, etc.
> 4. If I set `labels:` to `label` or `molecule`, the atom positions stop updating in the 3D view, and every frame produces an error:
> 
> ```
> <class 'ValueError'>
> zip() argument 2 is longer than argument 1
> <traceback object at 0x314e3ee80>
> Traceback (most recent call last):
>   File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 1094, in set_coordinates
>     self.update_atom_labels()
>   File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 374, in update_atom_labels
>     for follower, coord in zip(
> ValueError: zip() argument 2 is longer than argument 1
> ```
> 
> At the same time, the animation progress bar is still moving and changing the frame numbers.

Probably want `itertools.zip_longest(..., fillvalue=None)` or similar

ChiCheng45:
> Thanks for the changes. The speed is back to normal when using labels. The opacity is better. Even though I find some atoms still hard to see, it can be easily fixed by changing the colour of the background.
> 
> I noticed one more problem that may need fixing.
> 
> 1. I load a trajectory that does not contain **atom labels** or **molecule definitions**,
> 2. I start playing the animation,
> 3. It is possible to change different settings while playing the animation: cell/bonds/atoms visibility, axis type, etc.
> 4. If I set `labels:` to `label` or `molecule`, the atom positions stop updating in the 3D view, and every frame produces an error:
> 
> ```
> <class 'ValueError'>
> zip() argument 2 is longer than argument 1
> <traceback object at 0x314e3ee80>
> Traceback (most recent call last):
>   File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 1094, in set_coordinates
>     self.update_atom_labels()
>   File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 374, in update_atom_labels
>     for follower, coord in zip(
> ValueError: zip() argument 2 is longer than argument 1
> ```
> 
> At the same time, the animation progress bar is still moving and changing the frame numbers.

Thanks, this should be fixed now.

For the opacity, 0.2 was the value on protos for unpicked. It was set that way because so that it was easier to see picked atoms, if unpicked opacity becomes too large then it can be more difficult to determine what is and isn't picked. Probably need to add some opacity setting, but this will have to be done in another PR I think.

MDANSE/Src/MDANSE/MolecularDynamics/Connectivity.py
ChiCheng45: Yes it's similar. I want an array of indexes where 0, 1, 2, ... is repeated depending on the number of close contacts.

So if point with index 0 has 3 close contacts then I get,

[0, 0, 0, ...]

Then later I have a ks array which at the index of the point that is in close contact,

[1, 2, 4, ...]

So combined js and ks are array indexes of the points in close contact.

[0, 0, 0, ...]
[1, 2, 4, ...]

So point with index 0 is close to point with indexes 1, 2, 4 and etc.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: Return is fine I think. It's for the case when a trajectory is not loaded, or a unit cell doesn't exist, it returns early and doesn't draw an axes when either the direct or reciprocal options are selected.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: It could, but I don't think it's worth it right now. There are only two copies.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
ChiCheng45: Maybe the naming of the clear functions should be changed? 

self.clear_atoms() means clear the atom actors. The actual trajectory is not cleared in self.clear_atoms() so self._n_atoms should still have a value.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
ChiCheng45: I have applied this suggestion. Were there any other things in particular that you saw that needed this?

ChiCheng45:
@MBartkowiakSTFC 

In the molecular viewer, there are a few unused functions that I have chosen to leave in the code. Did you still want these here? Do you have some plans to update the molecule viewer to use them at some point?

```python
    def change_atom_trace_opacity(self, surface_index: int, opacity: float):
        """This method should allow changing the opacity of an already existing
        isosurface. Updates the renderer.

        Currently not connected to any widgets.

        Parameters
        ----------
        surface_index : int
            index of the isosurface in self._surfaces
        opacity : float
            new opacity value for the isosurface
        """

        if surface_index >= len(self.surface_actors):
            return

        self.surface_actors[surface_index].GetProperty().SetOpacity(opacity)
        self.update_renderer()

    def change_atom_trace_isocontour_level(self, surface_index: int, level: float):
        """This method should allow changing the isocontour level for an already existing
        isosurface. Updates the renderer.

        Currently not connected to any widgets.

        Parameters
        ----------
        surface_index : int
            index of the isosurface in self._surfaces
        level : float
            new value of isocontour level
        """

        if surface_index >= len(self.surface_actors):
            return

        self.isocontours[surface_index].SetValue(0, level)
        self.isocontours[surface_index].Update()
        self.update_renderer()

    def change_atom_trace_rendering_type(self, surface_index: int, rendering_type: str):
        """Method for changing the rendering style of an existing isosurface.
        Updates the renderer.

        Currently not connected to any widgets.

        Parameters
        ----------
        surface_index : int
            index of the isosurface in self._surfaces
        rendering_type : str
            one of the following: wireframe, surface, points
        """

        if surface_index >= len(self.surface_actors):
            return

        surface = self.surface_actors[surface_index]

        if rendering_type == "wireframe":
            surface.GetProperty().SetRepresentationToWireframe()
        elif rendering_type == "surface":
            surface.GetProperty().SetRepresentationToSurface()
        elif rendering_type == "points":
            surface.GetProperty().SetRepresentationToPoints()
            surface.GetProperty().SetPointSize(3)
        else:
            return

        self.update_renderer()
```

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
oerc0122: Just places where we check the lengths of properties directly, it might be useful.

MBartkowiakSTFC:
> For the opacity, 0.2 was the value on protos for unpicked. It was set that way because so that it was easier to see picked atoms, if unpicked opacity becomes too large then it can be more difficult to determine what is and isn't picked. Probably need to add some opacity setting, but this will have to be done in another PR I think.

We don't have to deal with this now, but definitely this branch produces different level of transparency than protos. I saved a screenshot of the same atoms using the main branch and then using this branch. Then I checked the RGB values of the brightest spot on the carbon atom. This branch is still darker.
transparency

I agree that this can be addressed in a separate PR.

MBartkowiakSTFC:
> 
> In the molecular viewer, there are a few unused functions that I have chosen to leave in the code. Did you still want these here? Do you have some plans to update the molecule viewer to use them at some point?
> 

I think that it is becoming increasingly unlikely that we will use these methods. If they are not used at the moment, you are welcome to remove them.

ChiCheng45:
> > For the opacity, 0.2 was the value on protos for unpicked. It was set that way because so that it was easier to see picked atoms, if unpicked opacity becomes too large then it can be more difficult to determine what is and isn't picked. Probably need to add some opacity setting, but this will have to be done in another PR I think.
> 
> We don't have to deal with this now, but definitely this branch produces different level of transparency than protos. I saved a screenshot of the same atoms using the main branch and then using this branch. Then I checked the RGB values of the brightest spot on the carbon atom. This branch is still darker. transparency
> 
> I agree that this can be addressed in a separate PR.

I see, I guess the difference is because I used a different actor type for the atoms. I've increased atoms and bond opacity to 0.4, so unpicked bonds will be slightly more opaque.

Protos
image


This branch
image


I've also removed the unused code for the surfaces, we can always get this back from the git history if we need it again.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
ChiCheng45: OK, I  think the other one was `len(self.surface_actors)` but it was in unused methods which I have now removed.

MDANSE/Src/MDANSE/MolecularDynamics/Connectivity.py
oerc0122: Maybe a comment at the top or make it a function for clarity?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py
```diff
@@ -132,7 +142,7 @@ def create_picked_bonds(self, *, opacity: float = 1.0):
         """
         self.clear_picked_bonds()
 
-        if self.bond_calc == BondCalc.NONE or len(self.picked_atoms) == 0:
+        if self.bond_calc == BondCalc.NONE or self.n_picked_atoms == 0:
```
oerc0122:
```suggestion
        if self.bond_calc is BondCalc.NONE or self.n_picked_atoms == 0:
```

oerc0122 (approved):
Looks good, just one instance of `==` where there should be `is` then I'm happy.

PR #1082, opened 2026-01-14
Add Panel base class and fix some typing

oerc0122:
**Description of work**
Reduce some code duplication and fix some typing by introducing abstract `Panel` base (not strictly abstract due to metaclass conflict).

**Fixes**
- Add `Panel` ABC to panels
- Fix some typing

**To test**
Should be no visible changes.


MBartkowiakSTFC:
It might be that `connect_logging` method ended up not being used for anything, as it was added before our plan for handling the logging was ready. Now we have the `LOG` object and Qt side of things catches everything that goes through it. You should be able to remove the logging-related things from `Panel`.

MBartkowiakSTFC:
For some reason these changes seem to have affected the order of the widgets in my GUI. I have a couple of screenshots as an example.

gui_widget_order

oerc0122:
Yes, it also affected the scaling ratio of the QSplitter, still trying to work out why.

ChiCheng45:
A few things have stopped working in the GUI e.g. the 3D viewer.

Screenshot 2026-05-14 at 10 00 23

PR #1087, opened 2026-01-15
Trajectory filter chunked calculation

### LINKED ISSUE #893: [BUG] Trajectory filter for long trajectories.
**Description of the error**
Currently the trajectory filter loads the entire trajectory before applying the filtering. This means that for some machine with lower amounts of RAM or very long trajectories this job might cause some out of memory issues.

I think there might be way around this since what we want to do is

```math
\mathrm{FT}^{-1}[r(\omega) g(\omega)] = r(x) * g(x) = r'(x)
```

where $r'(x)$ is the filtered positions, $r(\omega)$ is the FT of the positions,  $g(\omega)$ is the frequency filter, and $g(x)$ is the inverse FT of the frequency filter. 

If we use the convolution of $r(x)$ and $g(x)$ to determine $r'(x)$ then I think we can use one of the following methods

https://en.wikipedia.org/wiki/Overlap%E2%80%93add_method
https://en.wikipedia.org/wiki/Overlap%E2%80%93save_method

so we should be able to calculate the convolution of $r(x)$ and $g(x)$ and therefore $r'(x)$ in parts maybe something like 200 frames at a time or something.

ChiCheng45:
**Description of work**

**Fixes**
Closes #893

**To test**

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import json
 import math
+from traceback import print_tb
 
 import h5py
 import numpy as np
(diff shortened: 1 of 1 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
```diff
@@ -128,19 +162,59 @@ def run_step(self, index):
 
         """
         LOG.debug(f"Running step: {index}")
-        trajectory = self.trajectory
 
-        # get atom index
-        atom_index = self.trajectory.atom_indices[index]
+        self.atomic_trajectory_array.fill(0)
+        for idx in range(len(self.trajectory.atom_indices)):
+            atom_index = self.trajectory.atom_indices[idx]
```
oerc0122, line 168:
```suggestion
        for atom_index in self.trajectory.atom_indices:
```

ChiCheng45:
Actually, I think this is not the best idea. I think is possible to use replace filtfilt with two overlap-add or overlap-save passes. However, this would probably mean we will not have some sort of temporary file anyway. It might be better to simply do filtfilt per atom and save the results into a temporary hdf file.

PR #1090, opened 2026-01-16
Fix the line actor scale factor in the 3D viewer

ChiCheng45:
**Description of work**
The scale factor can change the width of the line actors. The bonds and uc box. This change fixes the scaling factor of the line actors.

**Fixes**
Line widths change when atoms size scaling is changed.

Animation15

**To test**
Try adjusting the atom size scaling in the GUI, change the trajectories, and etc. The bond and uc box line width should be fixed.


ChiCheng45:
@MBartkowiakSTFC Could also add another option to change the bond scaling? Figured it's better to fix this and leave that option for v2.1.

MBartkowiakSTFC (approved):
Looks good. The line width is fixed now. I could not find a way to change the line width after these changes.

We don't want to add new inputs at this stage. I agree that we can save this idea for 2.1.

PR #1093, opened 2026-01-16
Fix instruments tab

### LINKED ISSUE #1091: [BUG] Issues with custom instruments and instrument names
**Description of the error**
Some issues when creating new instruments in the instruments tab.

- Creating a new instrument can't be deleted.

Image

- If a custom instrument has the same name, then only one of them is saved.

Image

MBartkowiakSTFC:
**Description of work**
Minimal changes meant to ensure that every instrument in the instrument database has a unique name.

closes #1091 

**Fixes**
1. Instrument database can't be saved if the instrument names are not unique.
2. A validator prevents the user from typing an instrument name that is already in use.
3. New instruments get numbers appended to the name to avoid name conflicts.

**To test**
Try adding new instrument, copying existing instruments and deleting the new instruments. Save the instrument database and check if it loads correctly on startup. It should not be possible to create a new instrument with a duplicate name, or to save a database with duplicate names.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/InstrumentList.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/InstrumentList.py
@@ -46,10 +46,12 @@ class InstrumentList(QListView):
         self._backup_instruments = {}
 
     def get_blocked_names(self) -> set[str]:
+        """Return a set of all names that are already present in the model."""
         model = self.model()
         return {model.index(row, 0).data() for row in range(model.rowCount())}
 
-    def all_names_are_unique(self):
+    def all_names_are_unique(self) -> bool:
+        """Return True is every name only appears once in the model, False otherwise."""
         return len(self.get_blocked_names()) == self.model().rowCount()
 
     def contextMenuEvent(self, event: QContextMenuEvent) -> None:
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/InstrumentList.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/InstrumentList.py
@@ -19,7 +19,7 @@ import os
 
 import tomlkit
 from qtpy.QtCore import QModelIndex, Signal, Slot
-from qtpy.QtGui import QContextMenuEvent, QStandardItem
+from qtpy.QtGui import QContextMenuEvent, QStandardItem, QValidator
 from qtpy.QtWidgets import QAbstractItemView, QListView, QMenu
 from tomlkit.parser import ParseError
 from tomlkit.toml_file import TOMLFile
@@ -45,14 +45,14 @@ class InstrumentList(QListView):
         self._all_instruments = set()
         self._backup_instruments = {}
 
-    def get_blocked_names(self) -> set[str]:
+    def get_existing_names(self) -> set[str]:
         """Return a set of all names that are already present in the model."""
         model = self.model()
         return {model.index(row, 0).data() for row in range(model.rowCount())}
 
     def all_names_are_unique(self) -> bool:
         """Return True is every name only appears once in the model, False otherwise."""
-        return len(self.get_blocked_names()) == self.model().rowCount()
+        return len(self.get_existing_names()) == self.model().rowCount()
 
     def contextMenuEvent(self, event: QContextMenuEvent) -> None:
         index = self.indexAt(event.pos())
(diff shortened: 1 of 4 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
```diff
@@ -30,6 +30,47 @@
...
 
 
+class FreeNameValidator(QValidator):
+    """Validator which compares the input text to the stored set of strings.
+
+    The intention is not to allow the user to type in a name that is already
+    present in the input set.
+    """
+
+    def __init__(self, *args, instrument_list=None, **kwargs):
```
oerc0122:
```suggestion
    def __init__(self, *args, instrument_list: InstrumentList | None = None, **kwargs):
```
?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
```diff
@@ -30,6 +30,47 @@
...
+            Cursor position.
+
+        """
+        state = QValidator.State.Intermediate
+        if self.instrument_list is None:
+            return QValidator.State.Acceptable, input_string, position
+        if not input_string or input_string in self.instrument_list.get_blocked_names():
+            state = QValidator.State.Invalid
+        else:
+            state = QValidator.State.Acceptable
```
oerc0122: There are no branches where `Intermediate` might be used. 

```suggestion
        state = QValidator.State.Acceptable if (
            input_string 
            and self.instrument_list 
            and input_string not in self.instrument_list.get_blocked_names()
        ) else QValidator.State.Invalid
```
Or similar would do

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentInfo.py
```diff
@@ -26,6 +27,31 @@
...
+        Starting name, by default "Generic neutron instrument".
+
+    Returns
+    -------
+    str
+        First name root plus number that is not already in the input set.
+    """
+    if blocked_names is None:
+        return name_root
+    name_generator = (" ".join([name_root, str(number)]) for number in range(1, 2048))
```
oerc0122: Why `2048`?

```suggestion
    name_generator = (f"{name_root} {number" for number in range(1, 2048))
```
If we're limiting it to 2048, we probably want to raise a friendly error if not found (we *should* never run into this, but it's good to do)

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Views/InstrumentList.py
```diff
@@ -45,6 +45,15 @@ def __init__(self, *args, **kwargs):
         self._all_instruments = set()
         self._backup_instruments = {}
 
+    def get_blocked_names(self) -> set[str]:
```
oerc0122: Not sure I like the name "blocked_names", "used_names" more appropriate?

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentInfo.py
MBartkowiakSTFC: The choice of 2048 was completely arbitrary, leaving the users enough space to create new entries, and yet hoping that they stop creating new entries or copies after the first two thousands.

Would you prefer some solution without an upper limit (something from `itertools`)?

oerc0122: I'm fine with limited, but we need a nice way to handle the error if they exceed the limits.

MBartkowiakSTFC:
In the end I decided to replace `blocked_names` with `existing_names`, and use name generation without an upper limit by replacing `range` with `itertools.count`. I also noticed that it was possible to create multiple copies of the same instrument and end up with colliding names, so now the `(Copy)` also has a number next to it.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentInfo.py
```diff
@@ -26,6 +29,35 @@
 from MDANSE_GUI.Widgets.ResolutionWidget import ResolutionCalculator, widget_text_map
 
 
+def generate_name(
```
oerc0122, line 32: Out of curiosity would this be best placed in utils and used by the filename generator for MDANSE results too? Maybe a separate PR

oerc0122 (approved):
Looks good for now, may be able to extract some more utility from that name generator though

PR #1099, opened 2026-01-20
Add fixup call to FreeNameValidator

### LINKED ISSUE #1091: [BUG] Issues with custom instruments and instrument names

MBartkowiakSTFC:
**Description of work**
It is still possible to give instruments conflicting names. This PR extends the validator for the name field to catch invalid inputs after the user has finished editing.

closes #1091

**Fixes**
`FreeNameValidator.fixup` will add a `"_duplicate{number}"` tag to the instrument name if the user stops editing on a duplicate name.

**To test**
Try to input a duplicate name using all possible methods. The name should get corrected once the QLineEdit loses focus.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
@@ -75,15 +75,34 @@ class FreeNameValidator(QValidator):
             if (
                 input_string
                 and self.instrument_list
-                and input_string not in self.instrument_list.get_existing_names()
+                and input_string not in self.instrument_list.get_other_names()
             )
             else QValidator.State.Invalid
         )
         return state, input_string, position
 
-    def fixup(self, a0):
+    def fixup(self, invalid_string: str) -> str:
+        """Modify the current input string so it can pass validation.
+
+        This method will be called on the input string if the user stops editing
+        leaving the text in invalid state. Since this validator prevents duplicate
+        names from being entered, it will append a tag to the current name with
+        a number that is incremented as needed to find a name that does not
+        currently appear on the list.
+
+        Parameters
+        ----------
+        invalid_string : str
+            Contents of the text input field, currently not passing validation.
+
+        Returns
+        -------
+        str
+            Modified string which will be passed to validate method again.
+        """
         new_name = generate_name(
-            self.instrument_list.get_existing_names(), prefix=f"{a0}_duplicate"
+            self.instrument_list.get_other_names(),
+            prefix=f"{invalid_string}_duplicate",
         )
         return super().fixup(new_name)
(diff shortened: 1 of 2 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
```diff
@@ -77,6 +81,12 @@ def validate(self, input_string: str, position: int) -> tuple[int, str]:
         )
         return state, input_string, position
 
+    def fixup(self, a0):
+        new_name = generate_name(
+            self.instrument_list.get_existing_names(), prefix=f"{a0}_duplicate"
```
oerc0122: Is there a better name than `a0`? I know it's fairly obvious from context, just seems out of place.

oerc0122 (approved):
Looks ok if it works.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
MBartkowiakSTFC: It seems that the Qt API considers this to be a good name for an input string:
Screenshot 2026-01-20 at 11 17 49
Should I rename it to make things clearer, or keep it as `a0` to help us remember that this is a specific method defined by Qt that will be called implicitly?

oerc0122: Strictly we should follow that convention (hence the approval) but if we're only ever calling it positionally we can use whatever name we like. 

It's fine, it just goes against our conventions (mind you, so does much of the QT GUI stuff due to method overrides).

MBartkowiakSTFC:
I will make the function signature better.

After extra checks I still see some unexpected behaviour in this PR after all, because now it appends things also to names that should have been correct, if I delete a chunk of the name and change to a different entry on the list. I'll need a bit more time to fix it.

ChiCheng45:
Looks good, seems to work, so I guess all custom names will have _duplicate{number} postfix? I think its good but I'd imagine some users might complain about not being able to remove the _duplicate{number} postfix.

MBartkowiakSTFC:
> Looks good, seems to work, so I guess all custom names will have _duplicate{number} postfix? I think its good but I'd imagine some users might complain about not being able to remove the _duplicate{number} postfix.

This was the unexpected behaviour I wanted to get rid of. I think I got it right this time, though. The latest version from a few minutes ago should make more sense.

ChiCheng45 (approved):
Looks good, I am unable to find any issues.

Issue #1098, opened 2026-01-19
[BUG] DL_POLY converter molecule definitions don't work with CentreOfMassesTrajectory


MBartkowiakSTFC:
**Description of the error**
DL_POLY converter recognises and stores molecule definitions. The GUI recognises them as valid, but they do not work the same as molecule definitions from `TrajectoryEditor`.

**Describe the expected result**
`CentreOfMassesTrajectory` should replace each molecule with a single particle.
`MeanSquareDisplacement` of the COM trajectory should be lower than that of the trajectory with individual atoms.

The same DL_POLY trajectory processed using `TrajectoryEditor` with molecule detection **behaves as expected.**

**Describe the actual result**
After running `CentreOfMassesTrajectory` on the DL_POLY water, we get a trajectory without dummy atoms, but otherwise essentially unchanged:

<img width="712" height="560" alt="Image" src="https://github.com/user-attachments/assets/a88c6018-bcdc-40cb-8a72-44928da8ad73" />

The output trajectory still contains H and O atoms instead of artificial atoms replacing each water molecule.

Additionally, even though the trajectory appears unchanged, the results of `MeanSquareDisplacement` are different, with the trajectory produced by `COMT` showing larger MSD:

<img width="1221" height="635" alt="Image" src="https://github.com/user-attachments/assets/24d4b4c3-ed16-485f-8d16-cc3140d8c403" />

**Suggested fix**
Converters should write out the same molecule/bond information as `TrajectoryEditor`.

**Additional details**
DL_POLY molecules don't show correctly in the `rdkit` molecule view.

<img width="616" height="722" alt="Image" src="https://github.com/user-attachments/assets/da6e727e-740a-4c33-ba59-e616c21fc054" />

MBartkowiakSTFC:
Two observations here:
1. The reason why COMT does not work as expected is most likely the atom selection de-selecting all the dummy atoms. COMT only processes molecules that are **fully** selected, so it will do nothing in this case. We need to change either the default selection of COMT or the criteria applied to molecules.
2. The main difference between the DL_POLY trajectory and the TrajectoryEditor trajectory is that the latter saves the bond indices. This does not matter here; however, maybe we want to add some kind of bond search routine to the converters? Or even an optional molecule detection on conversion for all the converters?

oerc0122:
@ChiCheng45 @MBartkowiakSTFC Has this been solved by #1123? Seem to work now?

[...]

PR #1105, opened 2026-01-20
Include molecules with deselected dummy atoms in COMT

MBartkowiakSTFC:
**Description of work**
To partially address issue #1098 code of CenterOfMassesTrajectory is adjusted to accept and convert incompletely selected molecules, if the only atoms missing from the selection are dummy atoms. This is based on the assumption that dummy atoms should not affect the centre of mass of a molecule, and so it is safe to ignore them.

The dummy atoms are still included in the name of the output particle, not to implicitly rename molecules.

**Fixes**
1. Use only non-dummy atom indices in the COMT analysis.

**To test**
Load DL_POLY water trajectory with dummy atoms and run CentreOfMassesTrajectory. All the molecules should be replaced by artificial atoms.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
@@ -72,9 +72,15 @@ class CenterOfMassesTrajectory(IJob):
 
         self.cluster_composition = {}
         original_atom_list = chemical_system.atom_list
-        nondummy_indices = {ind for ind, dummy in 
-        zip(chemical_system._atom_indices, chemical_system.atom_property("dummy"), strict=True)
-        if not dummy}
+        nondummy_indices = {
+            ind
+            for ind, dummy in zip(
+                chemical_system._atom_indices,
+                chemical_system.atom_property("dummy"),
+                strict=True,
+            )
+            if not dummy
+        }
         selected_indices = (
             set(self.trajectory._selection)
             if self.trajectory._selection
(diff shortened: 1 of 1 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
```diff
@@ -72,6 +72,11 @@ def initialize(self):
 
         self.cluster_composition = {}
         original_atom_list = chemical_system.atom_list
+        nondummy_indices = set(
+            np.array(chemical_system._atom_indices)[
+                np.logical_not(chemical_system.atom_property("dummy"))
+            ]
+        )
```
oerc0122: Are some of these already numpy arrays? How big are they? Is avoiding the cast/copy more effcient?

```suggestion
        nondummy_indices = {ind for ind, dummy in 
        zip(chemical_system._atom_indices, chemical_system.atom_property("dummy"), strict=True)
        if not dummy}
```

MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
```diff
@@ -80,15 +85,24 @@ def initialize(self):
...
+                    nondummy_cluster
+                    and selected_indices
+                    and nondummy_cluster <= selected_indices
+                ):
                     if cluster_name not in self.cluster_composition:
                         self.cluster_composition[cluster_name] = [
                             original_atom_list[ind] for ind in cluster
                         ]
+                    if cluster_name not in self.temp_clusters:
+                        self.temp_clusters[cluster_name] = []
```
oerc0122:
```suggestion
                    self.temp_clusters.setdefault(cluster_name, [])
```

MDANSE/Src/MDANSE/Framework/Jobs/CenterOfMassesTrajectory.py
MBartkowiakSTFC: I am quite sure that these are Python lists, not arrays. The number of elements is the number of atoms in the trajectory. On the other hand, we only execute these lines once per run. In any case, I will accept your suggestion.

MBartkowiakSTFC:
This PR has been superseded by #1118

PR #1107, opened 2026-01-21
Always create plot legend, even if not visible

### LINKED ISSUE #1092: [BUG] Plot legend can disappear, checking/unchecking the legend box doesn't work anymore.
**Description of the error**
Unchecking the plot legend and then changing the plot to single or grid can cause the legend to disappear until you do something to make it reappear. 

**Additional details**

Image

MBartkowiakSTFC:
**Description of work**
Makes sure that legend is always created, so that toggling the legend checkbox always works.

Also, for `Grid` plots this PR replaces legend with subplot title, which should be valid since `Grid` always shows a single curve per plot. You are welcome to comment if this is meaningful or not.

Screenshot 2026-01-21 at 16 44 33

closes #1092

**Fixes**
1. Every plotter creates a legend, which can then be set to visible or not.
2. `Grid` plotter puts the single legend entry into the subplot title instead.

**To test**
Try toggling the legend on and off and change plotters in different order. Check if any errors are triggered.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
@@ -20,6 +20,8 @@ import math
 from itertools import islice
 from typing import TYPE_CHECKING, Any
 
+from matplotlib import rcParams
+
 from MDANSE.MLogging import LOG
 from MDANSE_GUI.Tabs.Plotters.Plotter import Plotter
 
@@ -39,6 +41,7 @@ class Grid(Plotter):
         self._active_curves = []
         self._backup_curves = []
         self._plot_limit = 8
+        self._title_length_limit = 30
 
     def slider_labels(self) -> list[str]:
         """Return labels to show that sliders are not used."""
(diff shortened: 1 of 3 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
```diff
@@ -84,6 +87,25 @@ def change_normalisation(self, new_value: dict[str, Any]):
             self._toolbar.update()
             self._toolbar.push_current()
 
+    def title_fontsize(self, title_text: str) -> int:
+        normal_size = rcParams["font.size"]
+        new_size = (
+            normal_size
+            if len(title_text) < self._title_length_limit
+            else normal_size - round(len(title_text) / self._title_length_limit)
```
oerc0122, line 95: Curious about the maths here, how was that form reached or is this standard?

ChiCheng45 (approved):
Looks good to me. Issue seems fixed and I think the legend as the title in grid view is a good idea.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
MBartkowiakSTFC: There isn't really much maths here. I just noticed that for long file names I end up getting subplot titles which overlap and are illegible. While most likely the user will want to change them anyway to something shorter, I decided to lower the font size temporarily for long labels.
So, if the label is shorter than the arbitrary limit I set, we use the normal font size. If it is longer, we reduce the font size by 1, or 2, etc.
It would be better to have some reliable check to tell us if the subplot titles really are overlapping.

PR #1108, opened 2026-01-21
Pass legend labels from PlotCreator to PlotHolder

### LINKED ISSUE #1101: [BUG] Plot Creator legend label is editable.
**Description of the error**
The legend label in the plot creator can be changed, but this change is not seen in the plot holder plots.

The legend label should not be editable, or the change should be seen in the resulting plot.

**Additional details**

Image

MBartkowiakSTFC:
**Description of work**
The easier of the two solutions to #1101 is to use the modified labels from `PlotCreator` to the plots in the next tab. It would be difficult to make the legend label non-editable in one tab and editable in the other, as this property is set on the side of the model and not the view.

closes #1101

There is still one questionable situation that could be handled better:
1. Put datasets into the table in PlotCreator.
2. Change their legend labels.
3. Send them to a plot. They will appear with the new labels.
4. Go back to PlotCreator, change the labels again and send them to the same plot.
5. The plot still contains the datasets with labels from point 2. (It recognised that the datasets are already present in the plot, so it does not accept them and does not modify the existing ones.)
Not sure what the correct behaviour should be in this case, and how often this scenario will occur.

**Fixes**
1. `PlotArgs` are passed to the new `PlottingContext` together with datasets, making it possible to access the new legend label.

**To test**
Try plotting different datasets in different plots while changing (or not) their legend labels on the PlotCreator side.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
@@ -788,9 +788,13 @@ class PlottingContext(QStandardItemModel):
         into this slot.
 
         """
-        for dataset in other._datasets.values():
-            self.add_dataset(dataset)
-            self.add_dataset(dataset.spawn_imaginary_dataset())
+        for dataset, plot_args in zip(
+            other._datasets.values(), other.datasets().values(), strict=True
+        ):
+            self.add_dataset(dataset, optional_values=plot_args)
+            self.add_dataset(
+                dataset.spawn_imaginary_dataset(), optional_values=plot_args
+            )
         self.set_axes()
 
     @Slot(dict)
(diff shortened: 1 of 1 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -881,7 +889,9 @@ def add_dataset(self, new_dataset: SingleDataset):
             QStandardItem(str(x))
             for x in [
                 new_dataset._name,
-                new_dataset._labels["medium"],
+                new_dataset._labels["medium"]
+                if optional_values is None
+                else optional_values.legend_label,
```
oerc0122: Could do

```suggestion
                getattr(optional_values, "legend_label", new_dataset._labels["medium"]),
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -788,9 +788,13 @@ def accept_external_data(self, other: PlottingContext):
         into this slot.
 
         """
-        for dataset in other._datasets.values():
-            self.add_dataset(dataset)
-            self.add_dataset(dataset.spawn_imaginary_dataset())
+        for dataset, plot_args in zip(
+            other._datasets.values(), other.datasets().values(), strict=True
```
oerc0122, line 792: Slightly concerned about `_datasets` and `datasets` being different here, but I guess that's baked in.

MBartkowiakSTFC: True, this is not perfect. While it would be good to clean it up, I would prefer not to risk messing up all the plotters just before the release.

ChiCheng45 (approved):
Looks good. For the question, I think it is fine as it is, since labels can be changed in the plot holder anyway.

PR #1111, opened 2026-01-23
Use Frames settings to set time steps for TrajectoryFilter

### LINKED ISSUE #1086: [BUG] Trajectory filter fails when "in steps of" setting is used.
**Description of the error**
I ran the trajectory filter with the DFTB water trajectory and changed the "in steps of" to 2. The job fails with the following error.

```
    (x[index], y[index], z[index]) for (x, y, z) in filtered_coordinates
     ~^^^^^^^
IndexError: index 50 is out of bounds for axis 0 with size 50
```

Should work. The output trajectory should be the same, but with half the number of frames and twice the time step size.

MBartkowiakSTFC:
**Description of work**
At the moment, the `TrajectoryFilter` uses the total number of trajectory frames as input. This causes conflicts with the actual inputs if the user has changed the parameters in `Frames`.

closes #1086

**Fixes**
1. `TrajectoryFilter` takes time step information directly from `Frames` and not from the trajectory.
2. The spinbox widgets update their step settings when frame parameters are changed.
3. `n_steps` and `time_step_ps` are not included as filter parameters.

**To test**
1. Confirm that the output of `TrajectoryFilter` is still the same for the default input parameters.
2. Try running the filter with "in steps of" larger than 1. Compare atom positions at specific timesteps to those of the trajectory with "in steps of"=1.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
@@ -40,6 +40,7 @@ class TrajectoryFilterConfigurator(IConfigurator):
     _default_filter = DEFAULT_FILTER
 
     _settings = filter_default_attributes()
+    time_axis = {"n_steps": 10, "time_step_ps": 0.001}
 
     @classmethod
     def get_default(cls) -> str:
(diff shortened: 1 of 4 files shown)


MBartkowiakSTFC:
This work has been superseded by #1176.

PR #1112, opened 2026-01-23
Add optional formatting of trajectory info

### LINKED ISSUE #1110: [BUG] Space in filename in the trajectory info textbox.
**Description of the error**
The number of spaces in the trajectory file textbox is not correct when multiple spaces are used.

**Additional details**
Files have a different number of spaces after the "a" but the path in the text box has squashed the spaces so both trajectories have the same path. This is also seen in the Conversion history information for the inputs/output_files info.

Image

Image

MBartkowiakSTFC:
**Description of work**
HTML `<pre></pre>` tags have been added to the trajectory name in the GUI.
In the CLI output, `html.escape` is skipped, as it made it difficult to read lines containing single and double quotes. (These were shown e.g. as `&quot;type=h1o_ff&quot` in the terminal output.)

closes #1110

**Fixes**
1. Extra formatting added to `trajectory_summary`

**To test**
Load a trajectory with spaces in the name into the GUI. The spaces should be preserved and shown correctly.
Use `mdanse traj SOME_NAME` in the CLI. Check if the output contains only legible characters.

### DIFF
--- a/MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
+++ b/MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
@@ -61,7 +61,8 @@ class GroupingLevels(UCEnum):
         return self.name.title()
 
 
-def trajectory_summary(traj: Trajectory):
+def trajectory_summary(traj: Trajectory, use_html: bool = False):
+    head = []
     val = []
     try:
         time_axis = traj.time()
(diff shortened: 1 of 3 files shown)


MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -73,8 +92,12 @@ def trajectory_summary(traj: Trajectory):
...
             timeline = f"[{summarise_array(time_axis, maxlen=5)}]\n"
 
-    val.append("Path:")
-    val.append(f"{traj.filename}\n")
+    filename = html.escape(f"{traj.filename}")
+    if use_html:
+        head.append(f"Path: <pre>{filename}</pre>")
+    else:
+        head.append("Path:")
+        head.append(f"{filename}\n")
```
oerc0122: Given these are fixed statements and we already have `val` is there any reason to create `head` and not just append these to `val`?

If it's due to the `HTML.escape`, we could still just create this as a string rather than an array to join:


```suggestion
    if use_html:
        head = f"Path: <pre>{filename}</pre>"
    else:
        head = "Path:\n{filename}\n"
```

oerc0122 (approved):

PR #1113, opened 2026-01-23
Check for warnings after changing instrument profiles

### LINKED ISSUE #1104: [BUG] Q-vector warning not shown when switching instruments
**Description of the error**
The q-vector input widget shows a warning message and is highlighted in yellow if the input parameters result in some (but not all) empty shells. However, this does not seem to happen when the q-vector parameters are changed using instrument profiles.

When changing the instrument definition, we expect to see the vector parameters change. At that moment the vector widget should also immediately show a warning if some of the shells are empty.

I changed the vector parameters by switching to a different instrument. The parameters changed and no error was shown. When I then manually changed the vector parameters, the warning came up and did not disappear after I changed the parameters back.

**Suggested fix**
Switching the instrument should trigger widget update.

**Additional details**
Screenshots of the same inputs showing a warning or not. The one without a warning followed a change of the instrument profile.

Image

Image

MBartkowiakSTFC:
**Description of work**
An additional check is performed after changing to a different instrument profile. The QVectorsWidget will show a warning immediately after the change if one of the generated q-vector shells is empty.

closes #1104

**Fixes**
- call `Action.allow_execution` again after changing the instrument profile.

**To test**
Load a trajectory. Pick an analysis that uses q vector generators (DISF, DCSF). Start changing instrument profiles. If no warning is shown after instrument change, manually change the random seed and check if the warning appears. (We expect the warning either to be there all the time, or not to be there at all. The current main branch fails to show the warning at first, and then keeps showing it after the first change of the random seed, which is not the correct behaviour.)

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/JobTab.py
@@ -184,6 +184,7 @@ class JobTab(GeneralTab):
             # we only update the widget if a job is selected from the
             # actions tree
             self.action.apply_instrument()
+            self.action.allow_execution()
 
     @Slot(int)
     def update_action_after_instrument_change(self, index: int):


oerc0122 (approved):
Huge PR, if it does what it's meant to, I'm happy with it.

Issue #1114, opened 2026-01-26
Default atom selection different in CLI and GUI


MBartkowiakSTFC:
All analysis types which enable atom selection start off with a selection composed of two operations:
1. Select all atoms,
2. Remove all dummy atoms.

However, this selection is created on the GUI side, in the `SelectionModel` class in `AtomSelectionWidget.py`. A script constructed using the CLI will contain just an empty selection, which is equivalent to selecting all atoms, **including** dummy atoms.

**Suggested solution**
1. The default selection should be created by code in `MDANSE` and not in `MDANSE_GUI`.
2. Each job should have a way to initialise the atom selection with its own defaults.
3. The GUI should introduce the items into `SelectionModel` based on the initial value of the specific job that has been selected.

oerc0122:
This was fixed by #1118

PR #1118, opened 2026-01-26
Unify GUI and CLI interfaces for `Selection`

oerc0122:
**Description of work**
- Make `default` on `AtomSelectionConfigurator` a used parameter
- Make GUI fall back on `_configurator.default`
- Pass `default` into `SelectionModel` from parent
- Add `reset` method to `SelectionModel` which restores this default.
- Override defaults on Jobs which should include dummies.

**Fixes**
#1114

**To test**
- Ensure GUI and CLI have same defaults.
- Ensure that `SelectionModel`, `SelectionHelper` and all descendants (`TransmutationWidget`, `PartialChargeWidget`, `SelectionWidget`) load and reset properly.


MBartkowiakSTFC (approved):
Looks good. I did not find any problems.

PR #1120, opened 2026-01-26
Show each dataset's scaling factor in plot details table

### LINKED ISSUE #1115: [ENHANCEMENT] Show the scaling factor in plot details
Currently the plot details table contains a checkbox in the "Apply weights" field which determines if the scaling factor is applied to the dataset or not. This could be improved by showing the actual value of the scaling factor in the text field in the same column.

If the scaling factor is a single scalar, it should be shown in the text field of "Apply weights".
If it is an array of numbers (e.g. for X-ray Static Structure Factor), `...` can be shown instead.
It should _not_ be possible to edit the scaling factor value.

We could also show the information about what property was used for calculating weights, but this can be accessed in the plot creator.

**Additional context**
N/A

MBartkowiakSTFC:
**Description of work**
Adds a fixed text to the "Apply weights?" field, showing the scaling factor of the dataset. For an array of scaling factors, it shows a string of `f"{first_value}, ..., {last_value}"`.

closes #1115

**Fixes**
1. Replaced the empty string in `PlottingContext.add_dataset()` with values from `SingleDataset._scaling_factor`.
2. Applied rounding to the values with `SCALE_FACTOR_ROUNDING=3` precision.

**To test**
Create plots of different analysis results. Check if numbers are shown next to the checkbox in the "Apply weights?" field. If the number is `1.0`, checking/unchecking the box should not affect the plot.
Run `XrayStaticStructureFactor` on a trajectory and plot the results. Check if text in "Apply weights?" is showing the first and last elements of a larger array.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
 
 NUMBERS_FOR_SLICE = 3
 NUMBERS_FOR_RANGE = 2
+SCALE_FACTOR_ROUNDING = 3
 
 
 class PlotArgs(NamedTuple):
@@ -889,7 +890,9 @@ class PlottingContext(QStandardItemModel):
                 self.next_colour(),
                 new_dataset._linestyle,
                 new_dataset._marker if new_dataset._marker else "None",
-                "",
+                round(new_dataset._scaling_factor, SCALE_FACTOR_ROUNDING)
+                if isinstance(new_dataset._scaling_factor, float)
+                else f"{round(new_dataset._scaling_factor[0], SCALE_FACTOR_ROUNDING)}, ..., {round(new_dataset._scaling_factor[-1], SCALE_FACTOR_ROUNDING)}",
                 new_dataset._filename,
             ]
         ]
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
@@ -36,6 +36,7 @@ from more_itertools import nth_product
 from qtpy.QtCore import QModelIndex, Qt, Signal, Slot
 from qtpy.QtGui import QColor, QStandardItem, QStandardItemModel
 
+from MDANSE.IO.IOUtils import summarise_array
 from MDANSE.MLogging import LOG
 
 if TYPE_CHECKING:
@@ -43,7 +44,7 @@ if TYPE_CHECKING:
 
 NUMBERS_FOR_SLICE = 3
 NUMBERS_FOR_RANGE = 2
-SCALE_FACTOR_ROUNDING = 3
+SCALE_FACTOR_FORMAT = "3.3f"
 
 
 class PlotArgs(NamedTuple):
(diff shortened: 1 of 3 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -889,7 +890,9 @@ def add_dataset(self, new_dataset: SingleDataset):
                 self.next_colour(),
                 new_dataset._linestyle,
                 new_dataset._marker if new_dataset._marker else "None",
-                "",
+                round(new_dataset._scaling_factor, SCALE_FACTOR_ROUNDING)
+                if isinstance(new_dataset._scaling_factor, float)
+                else f"{round(new_dataset._scaling_factor[0], SCALE_FACTOR_ROUNDING)}, ..., {round(new_dataset._scaling_factor[-1], SCALE_FACTOR_ROUNDING)}",
```
oerc0122: We have `MDANSE.IO.IOUtils.summarise_array` for this.

oerc0122 (changes requested):

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
MBartkowiakSTFC: Sure, but I will have to modify `summarise_array` to add rounding to it.

What would be an elegant way of introducing a `round(..., ROUNDING_PRECISION)` into `", ".join(map(str, array))`?

oerc0122:
```patch
diff --git a/MDANSE/Src/MDANSE/IO/IOUtils.py b/MDANSE/Src/MDANSE/IO/IOUtils.py
index 7b8fa36f..9ab89750 100644
--- a/MDANSE/Src/MDANSE/IO/IOUtils.py
+++ b/MDANSE/Src/MDANSE/IO/IOUtils.py
@@ -17,18 +17,20 @@ from __future__ import annotations
 
 import json
 import re
-from collections.abc import Iterable, Iterator, Sequence
 from enum import Enum
 from functools import singledispatch
 from itertools import filterfalse
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import numpy as np
-from more_itertools import first_true, value_chain
+from more_itertools import first_true, last, take, value_chain
 
 from MDANSE.MLogging import LOG
 
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator, Sequence
+
 MAX_FILE_COUNT = 2048
 
 
@@ -251,7 +253,7 @@ def strip_comments(
     return strip_function(data, comment_char=comment_char)
 
 
-def summarise_array(array: Sequence, *, maxlen: int = 6, show: int = 3) -> str:
+def summarise_array(array: Sequence, *, maxlen: int = 6, show: int = 3, arr_fmt: str | None = None) -> str:
     """
     Return a summarised string of the array.
 
@@ -266,6 +268,8 @@ def summarise_array(array: Sequence, *, maxlen: int = 6, show: int = 3) -> str:
         Maximum length before elision (min 4).
     show : int
         Number of elements to show.
+    arr_fmt : str, optional
+        Format values in array before printing.
 
     Returns
     -------
@@ -282,11 +286,16 @@ def summarise_array(array: Sequence, *, maxlen: int = 6, show: int = 3) -> str:
     '0, 1, 2, 3, 4, 5, 6, 7, 8, 9'
     >>> summarise_array(range(10), show=6)
     '0, 1, 2, 3, 4, 5, ..., 9'
+    >>> summarise_array([x / 3 for x in range(10)], arr_fmt="3.1f")
+    '0.0, 0.3, 0.7, ..., 3.0'
     """
+    fmt = str if arr_fmt is None else lambda x: format(x, arr_fmt)
+    arr = map(fmt, array)
+
     if len(array) <= maxlen or len(array) < show + 1:
-        return ", ".join(map(str, array))
+        return ", ".join(arr)
 
-    return ", ".join(map(str, value_chain(array[:show], "...", array[-1])))
+    return ", ".join(value_chain(take(show, arr), "...", last(arr)))
 
 
 def unused_standard_output_filename(
```

MBartkowiakSTFC:
Following the merge of #1127 this PR now uses `summarise_array`, and other `summarise_array` calls also use a format string.

oerc0122 (approved):

PR #1123, opened 2026-01-28
Remove weight from MSD calculation

### LINKED ISSUE #1109: Consistent handling of dummy atoms
Currently, the dummy atoms are not included in the atom selection by default. This has lead to problems with algorithms such as `CenterOfMassesTrajectory`, where this would make all the molecules with dummy atoms incompletely selected and dropped from the analysis (issue #1098). At the same time, the dummy atoms are included in the total atom count, which leads to the weights in different analysis types scaling differently.

This is an example of `MeanSquareDisplacement` results for water. In the DFTB trajectory there were no dummy atoms, and the results are all consistent. In the DL_POLY trajectory each water contains one dummy atom, and the MSD result appears unusually low:

Image
However, this is just a matter of the scaling factor, and the results are still correct if the scaling by the weight factor is not applied in the plotter:

Image

**Summary of the problem**

1. Dummy atoms can distort results at the moment if they were to be counted the same way as real atoms.
2. At the same time, dummy atoms may carry some physical property that needs to be conserved for the system (such as charge, torque, etc.), so they cannot simply be dropped from the trajectory.
3. The current atom selection which excludes dummy atoms is a sensible starting point for most analysis types, but there will be exceptions. E.g. infrared spectra may need to consider the charge on dummy atoms.

**Possible solutions**

1. Short-term: ignore the dummy atoms when calculating weights.
2. Long-term: create rule sets for different job types, specifying which behaviour should be applied: jobs dealing with electrical charged may either include the dummy atoms in the selection from the start or produce a warning if charge-carrying dummy atoms are not in the selection.
3. Other considerations: some converters may try to replace unknown/unidentified element types with dummy atoms. We need to make sure that the users notice this and prompt them to confirm that this is the behaviour they wanted - or to manually specify the chemical element of the atom that could not be recognised.

ChiCheng45:
**Description of work**
To avoid confusion with dummies and etc, I think it's best to simply remove weighting from the MSD calculation. This is will be consistent with the RMSD calculation. 

I have also renamed Apply weight? in the GUI to Apply scaling? which is more correct.

**Fixes**
Closes #1109

**To test**
Run the MSD calculation and check that results are correct for a trajectory with and without dummy atoms. Also check with atom or molecule grouping.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/MeanSquareDisplacement.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/MeanSquareDisplacement.py
@@ -90,16 +90,6 @@ class MeanSquareDisplacement(IJob):
             }
         },
     )
-    settings["weights"] = (
-        "WeightsConfigurator",
-        {
-            "dependencies": {
-                "trajectory": "trajectory",
-                "atom_selection": "atom_selection",
-                "atom_transmutation": "atom_transmutation",
-            }
-        },
-    )
     settings["output_files"] = ("OutputFilesConfigurator", {})
     settings["running_mode"] = ("RunningModeConfigurator", {})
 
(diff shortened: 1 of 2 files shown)


ChiCheng45:
@MBartkowiakSTFC The alternative to set apply weights unchecked by default is possible and I have a a branch for that but this would probably be the only job that did this and might also be confusing. I think it best to simply remove the weight scheme from MSD.

MBartkowiakSTFC (approved):
I think this solution is fine. Some other jobs will need changes too, but that should be done in another PR.

oerc0122 (approved):
Looks good.

PR #1129, opened 2026-01-29
Improve handling of dummy atoms

### LINKED ISSUE #1124: [BUG] Root Mean Square Fluctuation fails for dummy atoms
**Description of the error**
At the moment, the RMSF calculation uses the standard selection, where dummy atoms are discarded.

The calculation should run and produce results for the selected atoms only.

The calculation fails soon after start with the error:
```
File "MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py", line 141, in combine
idxs = self.ele_idxs[name]
~~~~~~~~~~~~~^^^^^^
KeyError: 'Du'
```

**Suggested fix**
RMSF should not try to access the `Du` key if no `Du` atoms were selected.

**Additional details**
N/A

### LINKED ISSUE #1125: [BUG] Dummy atoms affect averaging
**Description of the error**
The following analysis types are affected by the presence of dummy atoms AND have no atom selection input:
1. Temperature,
2. (less important) Density (atomic part only).

I ran the analysis twice: on the original DL_POLY water, and on the same trajectory with the `Du` atoms removed using `TrajectoryEditor`.

Temperature:
Image

Density (atomic):
Image

Ideally, the results should be the same for both trajectories.

The temperature is visibly different for the trajectory containing dummy atoms.

**Suggested fix**
Should we add atom selection to the temperature calculation?

**Additional details**
N/A

### LINKED ISSUE #1126: [BUG] Invert Selection does not work on DL_POLY water
**Description of the error**
The "Invert Selection" operation in atom selection helper dialog produces an error for DL_POLY water trajectory. The same trajectory works fine after the dummy atoms have been removed.

The selection should become inverted, i.e. only the dummy atoms should be selected.

```
2026-01-28 13:53:07,056 - ERROR - process[89930] - main 36 - EXCEPTION:
<class 'ValueError'>
zero-size array to reduction operation maximum which has no identity
<traceback object at 0x35a9aa740>
  File "MDANSE_GUI/InputWidgets/AtomSelectionWidget.py", line 570, in recalculate_selection
    self.view_3d._viewer.change_picked(self.selected)
  File "MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py", line 281, in change_picked
    self.change_picked_polydata_lines()
  File "MDANSE_GUI/MolecularViewer/MolecularViewerWithPicking.py", line 220, in change_picked_polydata_lines
    bonds = self.create_bond_cell_array(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MDANSE_GUI/MolecularViewer/MolecularViewer.py", line 685, in create_bond_cell_array
    dist, js, ks, _ = distance_calculation(rs, 2 * np.max(covs) + tolerance)
                                                   ^^^^^^^^^^^^
  File "python3.11/site-packages/numpy/core/fromnumeric.py", line 2810, in max
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "python3.11/site-packages/numpy/core/fromnumeric.py", line 88, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: zero-size array to reduction operation maximum which has no identity
```

**Suggested fix**
This may have something to do with dummy atoms having 0 covalent radius.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
This PR should remove several inconsistencies caused the the presence of dummy atoms in the input trajectory.

closes #1124 
closes #1125 
closes #1126

**Fixes**
1. RootMeanSquareFluctuation now works also with an incomplete atom selection.
2. Temperature and Density do not count dummy atoms when counting all the atoms in the system.
3. MolecularViewer does not calculate distances if it is given an empty list of covalent radii.

**To test**
Load the DL_POLY water into the GUI. Try to invert the selection in atom selection dialog.
Remove dummy atoms from DL_POLY water. Run Temperature and Density on the trajectory with and without dummy atoms. The results should be the same.
Run RootMeanSquareFluctuation on DL_POLY water with both atom and molecule grouping. Both should run successfully.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/Density.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/Density.py
@@ -62,14 +62,10 @@ class Density(IJob):
 
         self._n_frames = self.numberOfSteps
 
-        real_atoms = {
-            at_symbol
-            for at_symbol in self.trajectory.unique_elements
-            if not self.trajectory.get_atom_property(at_symbol, "dummy")
-        }
-
         self._n_atoms = ilen(
-            symbol for symbol in self.trajectory.atom_types if symbol in real_atoms
+            symbol
+            for symbol in self.trajectory.atom_types
+            if symbol in self.trajectory.non_dummy_elements
         )
 
         self._symbols = self.configuration["trajectory"][
--- a/MDANSE/Src/MDANSE/Framework/Jobs/Density.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/Density.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import numpy as np
+from more_itertools import ilen
 
 from MDANSE.Framework.Jobs.IJob import IJob
 from MDANSE.Framework.Units import measure
@@ -61,9 +62,15 @@ class Density(IJob):
 
         self._n_frames = self.numberOfSteps
 
-        self._n_atoms = self.configuration["trajectory"][
-            "instance"
-        ].chemical_system.number_of_atoms
+        real_atoms = {
+            at_symbol
+            for at_symbol in self.trajectory.unique_elements
+            if not self.trajectory.get_atom_property(at_symbol, "dummy")
+        }
+
+        self._n_atoms = ilen(
+            symbol for symbol in self.trajectory.atom_types if symbol in real_atoms
+        )
 
         self._symbols = self.configuration["trajectory"][
             "instance"
--- a/MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
@@ -20,6 +20,7 @@ from more_itertools import always_iterable
 
 from MDANSE.Framework.Jobs.IJob import IJob
 from MDANSE.Mathematics.Geometry import center_of_mass, moment_of_inertia
+from MDANSE.MLogging import LOG
 
 
 class Eccentricity(IJob):
@@ -103,6 +104,11 @@ class Eccentricity(IJob):
         conf = conf.contiguous_configuration()
         series = conf["coordinates"][self._indices, :]
 
+        if np.sum(np.abs(self._selectionMasses)) == 0.0:
+            LOG.warning(
+                "All atoms have 0 mass. Replacing mass with 1 to allow eccentricity calculation."
+            )
+            self._selectionMasses = np.ones(len(series))
         com = center_of_mass(series, masses=self._selectionMasses)
 
         # calculate the inertia moments
--- a/MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
@@ -104,7 +104,7 @@ class Eccentricity(IJob):
         conf = conf.contiguous_configuration()
         series = conf["coordinates"][self._indices, :]
 
-        if np.sum(np.abs(self._selectionMasses)) == 0.0:
+        if np.allclose(self._selectionMasses, 0.0):
             LOG.warning(
                 "All atoms have 0 mass. Replacing mass with 1 to allow eccentricity calculation."
             )
--- a/MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
@@ -87,14 +87,17 @@ class RootMeanSquareFluctuation(IJob):
             main_result=True,
         )
 
-        self.group_indices = {}
         if self.group_molecules:
-            for grp in self.trajectory.group_lookup:
-                self.group_indices[grp] = {
+            self.group_indices = {
+                grp: {
                     job_i: at_i
                     for job_i, at_i in enumerate(self.trajectory.atom_indices)
                     if f"<{grp}>" in self._names[job_i]
                 }
+                for grp in self.trajectory.group_lookup
+             }
+         else:
+             self.group_indices = {}
 
         for name in self.trajectory.unique_names:
             idxs = [
--- a/MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
@@ -70,7 +70,9 @@ class RootMeanSquareFluctuation(IJob):
         self.group_molecules = self.configuration["grouping_level"]["value"] != "atom"
         self.ele_idxs = {}
 
-        self._names = self.trajectory.atom_names
+        self._names = [
+            self.trajectory.atom_names[index] for index in self.trajectory.atom_indices
+        ]
 
         self._outputData.add(
             "rmsf/axes/indices/all",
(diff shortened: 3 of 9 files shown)


MDANSE/Src/MDANSE/Framework/Jobs/Density.py
```diff
@@ -61,9 +62,15 @@ def initialize(self):
...
-        ].chemical_system.number_of_atoms
+        real_atoms = {
+            at_symbol
+            for at_symbol in self.trajectory.unique_elements
+            if not self.trajectory.get_atom_property(at_symbol, "dummy")
+        }
+
+        self._n_atoms = ilen(
+            symbol for symbol in self.trajectory.atom_types if symbol in real_atoms
+        )
```
oerc0122: Possibly less clear but shorter:

```suggestion
        self._n_atoms = ilen(filter(real_atoms.__contains__, self.trajectory.atom_types))
```

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
```diff
@@ -86,19 +87,46 @@ def initialize(self):
...
-            idxs = [i for i in self.trajectory.atom_indices if names == self._names[i]]
-            self.ele_idxs[names] = idxs
+        self.group_indices = {}
+        if self.group_molecules:
+            for grp in self.trajectory.group_lookup:
+                self.group_indices[grp] = {
+                    job_i: at_i
+                    for job_i, at_i in enumerate(self.trajectory.atom_indices)
+                    if f"<{grp}>" in self._names[job_i]
+                }
```
oerc0122: Not sure if clearer?
```suggestion
        if self.group_molecules:
            self.group_indices = {
                grp: {
                    job_i: at_i
                    for job_i, at_i in enumerate(self.trajectory.atom_indices)
                    if f"<{grp}>" in self._names[job_i]
                }
                for grp in self.trajectory.group_lookup
             }
         else:
             self.group_indices = {}
```

MDANSE/Src/MDANSE/Framework/Jobs/Density.py
```diff
@@ -61,9 +62,15 @@ def initialize(self):
 
         self._n_frames = self.numberOfSteps
 
-        self._n_atoms = self.configuration["trajectory"][
-            "instance"
-        ].chemical_system.number_of_atoms
+        real_atoms = {
```
oerc0122: Would this be better becoming a property/method on `Trajectory`?

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
```diff
@@ -86,19 +87,46 @@ def initialize(self):
...
+
+        for name in self.trajectory.unique_names:
+            idxs = [
+                at_i
+                for job_i, at_i in enumerate(self.trajectory.atom_indices)
+                if self._names[job_i] == name
+            ]
```
oerc0122, line 109: Is this not a groupby operation?

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
```diff
@@ -681,6 +681,8 @@ def create_bond_cell_array(
         """
         # determine and set bonds without PBC applied
         bonds = vtk.vtkCellArray()
+        if not len(covs):
```
oerc0122, line 687: Is this part of this PR?

MBartkowiakSTFC: I included it here intentionally. This change allows us to invert selection on trajectories with dummy atoms.

MDANSE/Src/MDANSE/Framework/Jobs/Density.py
MBartkowiakSTFC: Sounds good. I made it a property: `Trajectory.non_dummy_elements`.

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
MBartkowiakSTFC: There probably is a way of re-writing this code to make it use `itertools.groupby`, but I would prefer not to do it, as I would most likely introduce new mistakes in indexing the atoms.

ChiCheng45 (commented):
It's possible to have a trajectory of just dummy atoms. Do we want to support this? Because there appear to be some issues when you try to do anything with just dummy atoms. E.g. atom selection helper doesn't open up with a trajectory of just dummy atoms.

Otherwise, everything else looks good.

MBartkowiakSTFC:
> It's possible to have a trajectory of just dummy atoms. Do we want to support this? Because there appear to be some issues when you try to do anything with just dummy atoms. E.g. atom selection helper doesn't open up with a trajectory of just dummy atoms.
> 

OK, I will have a look at how difficult it would be to make things work for trajectories with dummy atoms only. In principle we should aim for a situation where things work, or fail explicitly with a reasonable explanation for it. So, atom selection not opening at all is not what we want.

MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
```diff
@@ -103,6 +104,11 @@ def run_step(self, index: int):
         conf = conf.contiguous_configuration()
         series = conf["coordinates"][self._indices, :]
 
+        if np.sum(np.abs(self._selectionMasses)) == 0.0:
+            LOG.warning(
+                "All atoms have 0 mass. Replacing mass with 1 to allow eccentricity calculation."
+            )
+            self._selectionMasses = np.ones(len(series))
```
oerc0122: Is that true? We have a couple of issues here.

1. Do we have negative mass ions? Why the abs?
2. If yes, why would the sum of these equalling zero say they have zero mass?

I think we need to ask the question we want answered.


```suggestion
        if np.allclose(self._selectionMasses, 0.0):
            LOG.warning(
                "All atoms have 0 mass. Replacing mass with 1 to allow eccentricity calculation."
            )
            self._selectionMasses = np.ones(len(series))
```

MDANSE/Src/MDANSE/Framework/Jobs/RootMeanSquareFluctuation.py
oerc0122: Yes, it's sadly not terribly clean
```python
from more_itertools import bucket, first, last

    groups = bucket(
        zip(self._names, self.trajectory.atom_indices), first
    )
    for name in groups:
        idxs = list(map(last, groups[name]))
```

oerc0122 (commented):
Looks good, just one minor question

ChiCheng45:
Just noticed that the default atom selection for moleculartrace is to select all, including dummies. This was done in #1118, was this intentional? If not, maybe we can fix it here.

Otherwise, everything else looks good. A trajectory of just dummy atoms can be run for most jobs, I think, only the ROG, temperature, and Xraystaticstructurefactor fail. Mainly because properties are zero or don't exist for the dummy atom type. 

For the xray job it completes but results contain nan.

MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
MBartkowiakSTFC: Short answers:
1. Nothing seems to prevent us from creating a negative-mass atom in MDANSE. I just created a trajectory of Ar and antiAr atoms, where the total mass is 0 while none of the individual masses are 0. This gives rise to many new questions, since for such a trajectory some jobs will fail soon after start because of zero division when calculating weights.
2. I expect that, for real numbers, something like this should hold:
Screenshot 2026-02-04 at 16 35 28

However, a proof would be necessary.

In any case, thanks for the suggestion. I will accept it now.

MBartkowiakSTFC:
> Just noticed that the default atom selection for moleculartrace is to select all, including dummies. This was done in #1118, was this intentional? If not, maybe we can fix it here.
> 

I see that `MolecularTrace` just calculates a histogram for all the selected atoms, so it makes sense not to include the dummy atoms. I changed the default selection of `MolecularTrace` to the standard one, without the dummy atoms.

ChiCheng45 (approved):
Looks good to me.

MDANSE/Src/MDANSE/Framework/Jobs/Eccentricity.py
oerc0122: I agree with your analysis. I fear for the consequences of negative matter in standard potentials for the MD though. Mind you, no reason you couldn't invert the potential too. It's all just a frame of reference after all.

MBartkowiakSTFC: Currently it is possible to introduce negative mass atoms into the analysis, and to get negative temperatures as the result. But this is not going to happen accidentally, and only as a result of deliberate editing by the user. I am happy not to do anything about it.

oerc0122: Likewise.

PR #1145, opened 2026-02-02
Add Path as viable option to ChemicalSystem

### LINKED ISSUE #1144: [BUG] `Path` passed to Chemical system fails
**Description of the error**
The file being passed as `Path` to the `ChemicalSystem` constructor fails to be loaded and is passed through.

Failure for other reasons.

Failure trying to index path.

**Suggested fix**
Add `Path` to the `isinstance` check.

**Additional details**
N/A

oerc0122:
**Description of work**
Add `Path` to `isinstance` check.

**Fixes**
Fixes #1144

**To test**
Check failure reason for loading various files.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -510,16 +510,16 @@ class ChemicalSystem:
         if close_on_end:
             source.close()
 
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC:
I guess that if we expect `Path` objects, then we could say so also in the type hint or docstring?

MBartkowiakSTFC (approved):

Issue #1136, opened 2026-02-02
[ENHANCEMENT] Include file format version in output files


MBartkowiakSTFC:
**Is your feature request related to a problem? Please describe.**
The file format documentation is being created according to #1081 
Following this, we should include the information about the file format version in each output file.

Subtasks:
- [ ] #1141 
- [ ] #1142 

**Describe the solution you'd like**
The file format version of .mda and .mdt files should be included in every file.

**Describe alternatives you've considered**
N/A

**Additional context**
All the files created with pre-release versions will not be officially supported. We should produce a warning every time a user loads a file which does NOT have a version number at all.

oerc0122:
Currently the MDANSE version is written into the main file as part of the metadata. It would be sensible to use this to determine compatibility. 

MDANSE already uses `importlib.metadata.version` to write the current version, besides adding a dirty-flag for git versions, it would also be useful to add a warning on load if the versions don't match.

PR #1146, opened 2026-02-02
Add warning for differing MDANSE versions when loading file

oerc0122:
**Description of work**
Add `info` when loading files from different MDANSE versions and a `warning` if no version is present at all.

**Fixes**
Addresses part of #1136 

**To test**
Check when loading some of the test files that the appropriate warning is raised (note that many of the outdated test files fail to load due to earlier checks, you may have to dummy the version)

### DIFF
--- a/MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
+++ b/MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
@@ -25,6 +25,7 @@ import numpy as np
 import numpy.typing as npt
 from more_itertools import first
 
+import MDANSE
 from MDANSE.Chemistry import ATOMS_DATABASE
 from MDANSE.Chemistry.ChemicalSystem import ChemicalSystem
 from MDANSE.Chemistry.Databases import str_to_num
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC:
It works. There is one thing I would probably like to see: if a really old trajectory is loaded, the warning from line 131 is issued:
`WARNING - process[6080] - MdanseTrajectory 131 - Could not find /composition from badtime_ase_cuau.mdt. MDANSE will try to read it as H5MD next.`
and the loading method exits early, without issuing the other, more meaningful warning recommending to regenerate the file with the current version of MDANSE.

At the same time, we don't want to tell the users to regenerate the file if they are genuinely loading an H5MD file and not an MDT file.

oerc0122:
Do we want to do the version checks first, do you think?

MBartkowiakSTFC:
> Do we want to do the version checks first, do you think?

I tried swapping things around. I think doing the version check before the ChemicalSystem check should work.

MBartkowiakSTFC (approved):
Looks good. I think it produces meaningful warning messages.

PR #1148, opened 2026-02-02
Clear the property viewer on trajectory delete

### LINKED ISSUE #1134: [BUG] Property viewer does not clear after trajectory is removed.
**Description of the error**
The property viewer retains values from the previous trajectory even though it has been removed.

**Suggested fix**
The property viewer should clear when the trajectory is deleted.

**Additional details**
Image

ChiCheng45:
**Description of work**
The property viewer now clears when the trajectory is deleted in the trajectory tab.

**Fixes**
Fixes #1134
Error message no longer seen when atoms or trajectory is selected in the first drop-down menu in the property viewer when no trajectory is loaded.

**To test**
Load a trajectory and check that the property viewer work as usual. Delete the trajectory from the trajectory tab and check that the property viewer is cleared.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/PropertyWidget.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/PropertyWidget.py
@@ -423,8 +423,8 @@ class PropertyWidget(QWidget):
         """Clears the property viewer."""
         self._prop_selection.clear()
         self._prop_model.clear()
-        self._atom_props = {}
-        self._raw_props = {}
-        self._frame_props = {}
-        self._trajectory_props = {}
-        self.curr_selection = {}
+        self._atom_props.clear()
+        self._raw_props.clear()
+        self._frame_props.clear()
+        self._trajectory_props.clear()
+        self.curr_selection.clear()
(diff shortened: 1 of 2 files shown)


MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/PropertyWidget.py
```diff
@@ -416,3 +418,13 @@ def update_table(self) -> None:
...
+
+    def clear_viewer(self):
+        """Clears the property viewer."""
+        self._prop_selection.clear()
+        self._prop_model.clear()
+        self._atom_props = {}
+        self._raw_props = {}
+        self._frame_props = {}
+        self._trajectory_props = {}
+        self.curr_selection = {}
```
oerc0122: `dict`s all have a `.clear` as well.

```suggestion
        self._atom_props.clear()
        self._raw_props.clear()
        self._frame_props.clear()
        self._trajectory_props.clear()
        self.curr_selection.clear()
```

oerc0122 (approved):
Apart from non-functional change for symmetry all looks good if it works.

oerc0122 (approved):

PR #1150, opened 2026-02-02
Filter designer refactor

### LINKED ISSUE #1088: [BUG] Trajectory filter error when comb filter is used
**Description of the error**
When I load the DL poly water trajectory from the test files, open the filter designer, and select the comb filter, it does not appear to load up. The following error is seen in the console.

```
    raise ValueError('fs must be divisible by w0.')
ValueError: fs must be divisible by w0.
```

Using the setting anyway causes the job to fail with the same error.

### LINKED ISSUE #1138: [ENHANCEMENT] External PPS loaded from .mda file for filter designer UI
See #1137 
Split out the PPS calculation from inside the filter designer input widget, and replace it with a file source allowing for the user to load a PPS from an existing .mda file, eliminating the need to undertake the costly operation of calculating a fresh PPS.

### LINKED ISSUE #1139: [ENHANCEMENT] Remove irrelevant input fields from TrajectoryFilter
See #1137 

Remove irrelevant input fields from the TrajectoryFilter job page.

### LINKED ISSUE #1173: [BUG] Trajectory filter helper does not open with the default settings.
**Description of the error**
The trajectory filter opens with some default settings. When I open the trajectory filter for the first time, the filter function plot I see is not the default one. When I click apply settings, the settings change even though I didn't change anything in the trajectory filter helper.

**Additional details**
Image

RobBuchananCompPhys:
**Description of work**
This PR changes the behaviour of TrajectoryFilter, simplifying the filter preview mechanism and reducing the number of input parameters.

Closes #1138
Closes #1139 
Closes #1173
Closes #1088

**Fixes**
1. Removed input fields related to `PositionPowerSpectrum` from `TrajectoryFilter`.
2. Removed the on-demand calculation of position power spectrum from the filter designer widget.
3. Added single- and double-peak functions as dummy spectra to visualise attenuation.
4. Added the option of loading a pre-calculated position power spectrum into the attenuation plot.

**To test**
Load a trajectory. Run the `PositionPowerSpectrum` analysis, once with default parameters and again with non-standard frames settings. Plot the results.
Switch to `TrajectoryFilter`. Check the attenuation preview in the filter designer widget. Load each of the PPS files from the previous step and check if the spectrum in the attenuation plot matches the plots from the previous steps.


ChiCheng45 (commented):
Works great I don't see any issues. I have one suggestion, might be useful to warn or block the user from using PPS if the trajectory filename is different or trajectory lengths is different etc.

oerc0122:
> Works great I don't see any issues. I have one suggestion, might be useful to warn or block the user from using PPS if the trajectory filename is different or trajectory lengths is different etc.

I thought part of the objective was you can use any trajectory just to see what the attenuation function will do, so as to be able to run a short PPS you believe would be representative for your use cases to use as your template.

ChiCheng45:
> I thought part of the objective was you can use any trajectory just to see what the attenuation function will do, so as to be able to run a short PPS you believe would be representative for your use cases to use as your template.

OK I guess it makes sense to do that.

oerc0122:
> > I thought part of the objective was you can use any trajectory just to see what the attenuation function will do, so as to be able to run a short PPS you believe would be representative for your use cases to use as your template.
> 
> OK I guess it makes sense to do that.

Although I wonder if your suggestion of a warning might be good. Might just be good to highlight "This may not be what your final PPS looks like"

ChiCheng45 (commented):
Here I run a PPS calculation on a trajectory (vasp) with 29 correlation frames. Then I load up the DFTB water trajectory which has 100 frames and load the PPS trajectory from the calculation with VASP.

Image

Looks like there are some issues with the attenuation plot, too few points?

MBartkowiakSTFC:
I think that we should try to get #1170 and #1176 merged first. Then I will get back to this one and see how bad the plot problem is after the changes.

MBartkowiakSTFC:
> 
> Although I wonder if your suggestion of a warning might be good. Might just be good to highlight "This may not be what your final PPS looks like"

In general, I have nothing against a warning, but I am worried that this one will just be triggered all the time. Is there any situation in which we would NOT issue this warning?

oerc0122:
> > Although I wonder if your suggestion of a warning might be good. Might just be good to highlight "This may not be what your final PPS looks like"
> 
> In general, I have nothing against a warning, but I am worried that this one will just be triggered all the time. Is there any situation in which we would NOT issue this warning?

I don't know. In general most people might still take the longer route of getting the full PPS just because that's what they know. In that case the final PPS will be close, but maybe just a disclaimer message warning that the final PPS won't necessarily be this? Nothing is also fine.

MBartkowiakSTFC:
> Looks like there are some issues with the attenuation plot, too few points?

I think I got the plot to some sensible scaling. There were two issues, one with the x-axis scaling and the other with the plot limits.

MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
ChiCheng45: I think running_mode is unused, maybe this can be removed?

MDANSE/Src/MDANSE/Framework/Jobs/TrajectoryFilter.py
```diff
@@ -61,46 +61,32 @@ class TrajectoryFilter(IJob):
...
-            },
+                "frames": "frames",
+                "pps_input_file": "pps_input_file",
+            }
         },
     )
+    settings["atom_selection"] = (
+        "AtomSelectionConfigurator",
+        {"dependencies": {"trajectory": "trajectory"}},
+    )
```
ChiCheng45, line 89: Is it worth adding atom transmutation aswell?

ChiCheng45:
> > Looks like there are some issues with the attenuation plot, too few points?
> 
> I think I got the plot to some sensible scaling. There were two issues, one with the x-axis scaling and the other with the plot limits.

I still see the same issue. When I load up a vasp PPS with DFTB loaded in the trajectory filter.

image

The trajectory response and attenuation have a different number of plot points.

MBartkowiakSTFC:
> The trajectory response and attenuation have a different number of plot points.

I think that this will always be the case. There are two scenarios:
1. We are using the correct PPS for the same trajectory. Then the filter response is calculated for the number of points covered by the trajectory frames. The PPS will have the number of points corresponding to the number of correlation frames, which can be half of trajectory frames, but may also be some other number.
2. We are using the wrong PPS from a different trajectory. Then my intention was to show a clear mismatch of the expected attenuation compared to the energy axis of the PPS, and it is done by keeping the axes separate and only showing the change to the spectrum in the range that is covered by the input trajectory.

Do you think that this is an acceptable way of dealing with the problem of potential mismatch.

The important thing for me is to make sure that the preview is correct and usable if we use the matching PPS.

ChiCheng45:
> > The trajectory response and attenuation have a different number of plot points.
> 
> I think that this will always be the case. There are two scenarios:
> 
> 1. We are using the correct PPS for the same trajectory. Then the filter response is calculated for the number of points covered by the trajectory frames. The PPS will have the number of points corresponding to the number of correlation frames, which can be half of trajectory frames, but may also be some other number.
> 2. We are using the wrong PPS from a different trajectory. Then my intention was to show a clear mismatch of the expected attenuation compared to the energy axis of the PPS, and it is done by keeping the axes separate and only showing the change to the spectrum in the range that is covered by the input trajectory.
> 
> Do you think that this is an acceptable way of dealing with the problem of potential mismatch.
> 
> The important thing for me is to make sure that the preview is correct and usable if we use the matching PPS.

Actually, yes the current way is fine actually now I've thought about it.

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
```diff
@@ -1116,56 +1211,55 @@ def resample_and_normalise(self, values, to_len):
...
-            new_params,
-            ["pps/axes/romega", "pps/isotropic/total"],
+        freqs = tr_filter.freq_response.frequencies
+        attenuation_fn = FilterAttenuationGroup.functional_forms[source]
+        if source == "Gaussian":
+            values = attenuation_fn(freqs, freqs.max() / 2, freqs.max() / 4)
+        elif source == "Double gaussian":
+            values = attenuation_fn(freqs, freqs.max() / 2, freqs.max() / 8)
+        elif source in {"Gaussian noise", "Uniform noise"}:
+            values = attenuation_fn(freqs)
```
oerc0122: These don't exist in `FilterAttenuationGroup.functional_forms`. Why are they here?

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
```diff
@@ -1189,20 +1285,25 @@ def set_trajectory_power_spectrum(
         """
         response = tr_filter.freq_response
 
-        # Trajectory power spectrum data
-        freqs, values = copy.deepcopy(self._trajectory_power_spectrum)
+        file_freqs, file_values = read_pps_from_file(pps_filename)
+        freqs, _ = response
+        file_freqs /= 2 * np.pi
```
oerc0122: `file_freqs` might be `None` here.

oerc0122 (approved):
If it's working as expected, I'm happy.

ChiCheng45 (approved):
Looks good.

PR #1155, opened 2026-02-03
Fix unit cell draw on trajectory switch

### LINKED ISSUE #1152: [BUG] Unit cell is shown in 3D view for trajectory without unit cell definition
**Description of the error**
When switching from a trajectory with unit cell to a trajectory without one, we get a 3D view showing the old unit cell superimposed onto the new trajectory.

The 3D view should show the contents of the current trajectory file and nothing else.

Image

Also, changing to a different trajectory frame updates the atom positions and does NOT remove the unit cell from the view.

**Suggested fix**
Remove the unit cell when changing to a new trajectory?

**Additional details**
N/A

ChiCheng45:
**Description of work**
Fixes the issue where the unit cell is drawn for a system without a unit cell.

**Fixes**
Fixes #1152

**To test**
Load a trajectory with unit cell data and without unit cell data. Switch between the trajectories in the 3D viewer. The unit cell should only be shown for the trajectory with UC data.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/MolecularViewer.py
@@ -428,15 +428,14 @@ class MolecularViewer(QtWidgets.QWidget):
             return
 
         uc = self._reader.read_pbc(self._current_frame)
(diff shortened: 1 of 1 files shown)


MBartkowiakSTFC (approved):
Looks good. The unit cell is shown only when it really exists.

PR #1156, opened 2026-02-03
Input parameter GUI labels

### LINKED ISSUE #1147: [ENHANCEMENT] Use consistent naming for input parameters
Some of the parameter input names are not clear at the moment, especially "project coordinates", which uses the word "project" as a verb in imperative form, where we would normally expect a noun based on the other variable names. Also, the same name uses a space to separate words, while other names use an underscore.

Ideally, we should use the same characters as separators in all names. What is better, space or underscore?

Trajectory converters use spaces more frequently and also have a variable "Fold coordinates into box" with an imperative verb. This is less confusing, but we can also aim to replace it for consistency.

MBartkowiakSTFC:
**Description of work**
This PR attempts to make the input parameter labels more uniform in the GUI. Current rules:
1. Labels consist of multiple words separated by spaces,
2. The first word starts with an uppercase letter,
3. Imperative form allowed for boolean checkboxes.

The labels are applied with the following priority:
1. The specific label set in the job file using the dictionary of the configurator,
2. The general label defined as the configurator class attribute,
3. If all the labels are missing, use the dictionary key from the job settings (but this should not happen).

The new labels are **verbose**. Comments are welcome on whether this is a good or a bad idea.

Since the scripts, examples and unit tests use the settings dictionary and its keys for every job, this PR should change the appearance of the GUI without breaking anything in the tests or MDANSE-Examples.

closes #1147

**Fixes**
1. Introduced label attribute in several configurators,
2. Added code to Action.py to check for configurator class attribute "label",
3. Removed multiple "label" entries from configurator dictionaries in individual jobs.

**To test**
Run the GUI, check if the input field labels make sense.
Check if the labels are too long.
Save a script and see if the labels are saved as comments.


MDANSE/Src/MDANSE/Framework/Configurators/OutputFilesConfigurator.py
```diff
@@ -49,7 +49,8 @@ class OutputFilesConfigurator(IConfigurator):
         ["MDAFormat", "TextFormat", "FileInMemory"],
         "no logs",
     )
-    _label = "Output filename and formats (filename, [format, ...])"
+    label = "Analysis result files (name and format)"
+    tooltip = "It is possible to use MDAFormant and TextFormat simultaneously."
```
oerc0122:
```suggestion
    tooltip = "It is possible to use MDAFormat and TextFormat simultaneously."
```

ChiCheng45 (approved):
Looks good to me. Just the typo as mentioned above.

PR #1159, opened 2026-02-04
Introduce nuclear_spin into atoms.json as float

### LINKED ISSUE #1131: [BUG] nuclear spin is not correct for some isotopes.
**Description of the error**
Noticed that the nuclear spin of oxygen-17 is zero in the MDANSE database, I believe it should be 5/2. 

Image

There are other isotopes that are not correct. Carbon-13 is another one.

MBartkowiakSTFC:
**Description of work**
This PR is meant to bring back the values of nuclear spin in the atoms database which had been lost due to the data format  being specified as `int` for that field in the file.

closes #1131

**Fixes**
1. Changed data type of "nuclear_spin" to float in "atoms.json,
2. Copied the values of "nuclear_spin" from the "elements_database.csv" (legacy branch), but ignoring the data type "int" used in that file. The values have been written into "atoms.json".

**To test**
Check if the values have been replaced correctly.


oerc0122 (approved):
Thanks, this must have been a lot of tedious work. If everything's passing, I'm happy.

PR #1166, opened 2026-02-06
Copy over charges in TrajectoryFilter

### LINKED ISSUE #1151: [BUG] TrajectoryFilter removes charge information
**Description of the error**
If `TrajectoryFilter` is used on an input trajectory with charge information, the charges will not be written into the output trajectory.

It would be good to preserve the charge information where possible, in case the user was planning to compare IR spectra before and after the filter has been applied.

The charges are shown as 0 in the output trajectory.

**Suggested fix**
It should be sufficient to copy the charges from each input trajectory to the output.

**Additional details**
N/A

### LINKED ISSUE #1168: [BUG] Molecule information lost after trajectory filtering
**Description of the error**
After running the trajectory filter, the molecule information is lost.

Similarly to the trajectory editor, the molecule information should be copied into the new trajectory file.

MBartkowiakSTFC:
**Description of work**
Adds code to `TrajectoryFilter` to make sure that the charge information included in the input trajectory is written to the output trajectory.

ADDED: 
1. The same functions handles molecule definitions in `TrajectoryEditor` and `TrajectoryFilter`. `TrajectoryFilter` will now copy the existing molecule definitions into the output trajectory.
2. `TrajectoryFilter` should run without errors when not all atoms/frames are selected.

closes #1151
closes #1168

**Fixes**
Add a loop over input frames that passes the charges of selected atoms to TrajectoryWriter.

**To test**
Run TrajectoryFilter on a trajectory containing charge information (DFTB water?). The output trajectory should contain the same charges.
Use atom selection in TrajectoryFilter to select an arbitrary subset of atoms and check if the charges were assigned correctly in the output trajectory.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -69,7 +69,7 @@ def assign_molecules_after_atom_selection(
         for val in vals:
             new_cluster = set(val) & selected_idxs
             if new_cluster:
-                new_cluster = sorted([indx_map[i] for i in new_cluster])
+                new_cluster = sorted(indx_map[i] for i in new_cluster)
                 selected_clusters[key].append(new_cluster)
 
     new_cs._clusters = selected_clusters
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import copy
 import itertools as it
+from collections import defaultdict
 from collections.abc import Iterable
 from functools import reduce
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
Actually, it seems that atom selection causes the job to fail in this branch and in the main branch at the moment.

UPDATE: I was wrong, it seems it was actually a problem with my frame selection.

MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -35,6 +36,45 @@
...
+        if i in selected_idxs and j in selected_idxs
+    ]
+    new_cs.add_bonds(selected_bonds)
+
+    selected_clusters = defaultdict(list)
+    for key, vals in original_cs._clusters.items():
+        for val in vals:
+            new_cluster = set(val) & selected_idxs
+            if new_cluster:
+                new_cluster = sorted([indx_map[i] for i in new_cluster])
```
oerc0122: `sorted` works happily on `Iterable`s:
```suggestion
                new_cluster = sorted(indx_map[i] for i in new_cluster)
```

oerc0122 (approved):
All looks reasonable if it works

PR #1169, opened 2026-02-06
Reactivate output prediction

### LINKED ISSUE #1167: [BUG] Output prediction stopped updating in the GUI
**Description of the error**
For each analysis type, the MDANSE GUI shows the prediction of the output axes of the calculated results. This is meant to allow the users to check the covered range of time, energy, etc. resulting from the input parameters _before_ they run the analysis.

At the moment the output prediction does not change when the input parameters are changed.

The output prediction should change every time a corresponding input parameter has been changed by the user.

The output prediction is calculated once and never changes. The actual range covered by the output file does not match the prediction if the input parameters were changed.

**Suggested fix**
Most likely a problem around `Action.py`.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Reactivates the output prediction updates.

This PR has been tested on a trajectory with 40000 frames and 34000 atoms to check that the GUI is responsive for large trajectory files.

closes #1167

**Fixes**
1. Connected `WidgetBase.value_changed` to `Action.show_output_prediction` for all widgets.
2. Removed `value_updated` signal altogether.

**To test**
Load a trajectory and pick different Analysis jobs. Check that the output prediction updates when:
1. the input parameters are changed manually,
2. the user picks a different instrument in the combo box.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/DistHistCutoffConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/DistHistCutoffConfigurator.py
@@ -23,6 +23,7 @@ class DistHistCutoffConfigurator(RangeConfigurator):
         self.prediction = PredictionSettings(
             key="value",
             label="Interatomic distance",
+            unit="nm",
         )
 
     def configure(self, value):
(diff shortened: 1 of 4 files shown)


oerc0122 (approved):
If this does what it sets out to, looks good.

ChiCheng45 (commented):
Looks great, works fine.

Can we also fix the units for interatomic distances while we are at it?

Image

Here is the output prediction for PDF. Should be angstrom.

MBartkowiakSTFC:
> 
> Can we also fix the units for interatomic distances while we are at it?
> 

It should work now.

ChiCheng45 (approved):
Looks good.

PR #1176, opened 2026-02-16
Make filter fail early for invalid input parameters

MBartkowiakSTFC:
**Description of work**
At the moment it is possible to run the TrajectoryFilter with wrong input parameters and have it run all the way to 100% progress before it first attempts to create the filter (and fails). To save the user's time, we should prefer the filter to fail early.

This PR connects the frames widget with filter widget, but does not solve all the problems such as plot scaling. (Should it?)

**Fixes**
1. Create the filter in the `TrajectoryFilter.initialize` method to avoid processing the trajectory if the filter can't be used anyway.
2. Test the filter parameters in `TrajectoryFilterConfigurator` to catch invalid filter parameters in the GUI and prevent the run.
3. Take the time step information from frames configurator to prevent the GUI tests from failing.

**To test**
Try introducing deliberately wrong filter parameters in the input. Ideally, the configurator will pick it up and block the execution in the GUI. If it does not, the job should fail immediately and not only after processing the trajectory.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
@@ -18,6 +18,8 @@ from __future__ import annotations
 import json
 from traceback import format_exc
 
+import numpy as np
+
 from MDANSE.Framework.Configurators.IConfigurator import IConfigurator
 from MDANSE.Mathematics.Signal import (
     DEFAULT_FILTER,
@@ -94,9 +96,17 @@ class TrajectoryFilterConfigurator(IConfigurator):
         )
 
         try:
-            FILTER_MAP[dict_value["filter"]](**dict_value["attributes"])
+            filter_instance = FILTER_MAP[dict_value["filter"]](
+                **dict_value["attributes"]
+            )
         except Exception as e:
             self.error_status = f"Could not create the filter. {e}: {format_exc()}"
+            return
+
+        try:
+            filter_instance.apply(np.empty((frames_configurator["number"],)))
+        except Exception as e:
+            self.error_status = f"Could not apply the filter. {e}: {format_exc()}"
             return
 
         self.error_status = "OK"
--- a/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
@@ -77,8 +77,10 @@ class TrajectoryFilterConfigurator(IConfigurator):
         try:
             dict_value = json.loads(value)
 
-            if not self._expected_keys <= dict_value.keys():
-                self.error_status = f"The input dictionary does not contain the expected keys {self._expected_keys}."
+            if missing := self._expected_keys - dict_value.keys():
+                self.error_status = (
+                    f"The input dictionary does is missing the expected keys {missing}."
+                )
                 return
 
         except (TypeError, ValueError):
@@ -90,7 +92,9 @@ class TrajectoryFilterConfigurator(IConfigurator):
         frames_configurator = self.configurable[self.dependencies["frames"]]
 
         dict_value["attributes"].setdefault("n_steps", frames_configurator["number"])
-        dict_value["attributes"].setdefault("time_step_ps", frames_configurator["time_step"])
+        dict_value["attributes"].setdefault(
+            "time_step_ps", frames_configurator["time_step"]
+        )
 
         try:
             FILTER_MAP[dict_value["filter"]](**dict_value["attributes"])
--- a/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
@@ -73,6 +73,7 @@ class TrajectoryFilterConfigurator(IConfigurator):
         """
         if not self.update_needed(value):
             return
+        self.warning_status = ""
 
         self._settings = value
 
@@ -108,6 +109,15 @@ class TrajectoryFilterConfigurator(IConfigurator):
         except Exception as e:
             self.error_status = f"Could not apply the filter. {e}: {format_exc()}"
             return
+
+        expected_attributes = set(filter_instance.default_settings) | {
+            "n_steps",
+            "time_step_ps",
+        }
+        if unknown_attributes := dict_value["attributes"].keys() - expected_attributes:
+            self.warning_status = (
+                f"Unexpected filter attributes: {','.join(unknown_attributes)}"
+            )
 
         self.error_status = "OK"
         self["value"] = self._settings
--- a/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
@@ -110,13 +110,23 @@ class TrajectoryFilterConfigurator(IConfigurator):
             self.error_status = f"Could not apply the filter. {e}: {format_exc()}"
             return
 
-        expected_attributes = set(filter_instance.default_settings) | {
+        expected_attributes = filter_instance.default_settings.keys() | {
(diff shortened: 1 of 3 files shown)


MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
```diff
@@ -76,10 +78,27 @@ def configure(self, value: str):
             dict_value = json.loads(value)
 
             if not self._expected_keys <= dict_value.keys():
-                self.error_status = f"The dictionary \n{dict_value}\n does not contain the expected keys {self._expected_keys}."
+                self.error_status = f"The input dictionary does not contain the expected keys {self._expected_keys}."
+                return
```
oerc0122: Could do:
```suggestion
            if missing := self._expected_keys - dict_value.keys():
                self.error_status = f"The input dictionary does not contain the expected keys, missing {', '.join(missing)}."
                return
```

MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
```diff
@@ -76,10 +78,27 @@ def configure(self, value: str):
...
+                f"Input {value} is not of correct format (expected JSON string)."
+            )
+            return
+
+        frames_configurator = self.configurable[self.dependencies["frames"]]
+
+        if "n_steps" not in dict_value["attributes"]:
+            dict_value["attributes"]["n_steps"] = frames_configurator["number"]
+        if "time_step_ps" not in dict_value["attributes"]:
+            dict_value["attributes"]["time_step_ps"] = frames_configurator["time_step"]
```
oerc0122:
```suggestion
        dict_value["attributes"].setdefault("n_steps", frames_configurator["number"])
        dict_value["attributes"].setdefault("time_step_ps", frames_configurator["time_step"])
```

MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
```diff
@@ -76,10 +78,27 @@ def configure(self, value: str):
...
+
+        try:
+            _ = FILTER_MAP[dict_value["filter"]](**dict_value["attributes"])
```
oerc0122:
```suggestion
            FILTER_MAP[dict_value["filter"]](**dict_value["attributes"])
```

MDANSE/Src/MDANSE/Framework/Configurators/TrajectoryFilterConfigurator.py
```diff
@@ -77,8 +77,10 @@ def configure(self, value: str):
         try:
             dict_value = json.loads(value)
 
-            if not self._expected_keys <= dict_value.keys():
-                self.error_status = f"The input dictionary does not contain the expected keys {self._expected_keys}."
+            if missing := self._expected_keys - dict_value.keys():
+                self.error_status = (
+                    f"The input dictionary does is missing the expected keys {missing}."
```
oerc0122: Minor typo, plus I think it might be more descriptive as:
```suggestion
                    f"The input dictionary is missing the expected keys: {', '.join(missing)}."
```
or
```suggestion
                    f"The input dictionary is missing the expected keys (missing: {', '.join(missing)})."
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
oerc0122: `match` could be more expressive here, in my PR I'm pulling out:
```python
def get_val(widget: QWidget) -> Any:
    """Get standard value from a widget.

    Parameters
    ----------
    widget : QWidget
        Widget to get value from.

    Returns
    -------
    Any
        Widget's value as appropriate type.
    """
    match widget:
        case QComboBox():
            return widget.currentText()
        case QCheckBox():
            return widget.isChecked()
        case QSpinBox():
            return widget.value()
        case QDoubleSpinBox():
            return round(widget.value(), FLOAT_SPINBOX_DECIMALS)
        case QLineEdit() if widget.text().isdigit():  # Int line
            return int(widget.text())
        case QLineEdit() if all(x.isdigit() for x in widget.text().split(".", 1)):  # Float line
            return float(widget.text())
        case QLineEdit():
            return widget.text()
```

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
```diff
@@ -645,6 +643,30 @@ def as_grid(self) -> QGridLayout:
...
+        n_steps : int
+            Number of trajectory steps currently selected by frames widget.
+        time_step : float
+            Time step between consecutive trajectory frames.
+        """
+        bin_width = Filter.frequency_resolution(n_steps, time_step, units=self.units)
+        vmax = Filter.nyquist(time_step, units=self.units) - bin_width
+        with block_signals(self):
+            for key in ("cutoff_freq", "fundamental_freq"):
+                if key in self.widgets:
```
oerc0122:
```suggestion
            for key in {"cutoff_freq", "fundamental_freq"} & self.widgets.keys():
```
If widgets is a `dict`

MDANSE_GUI/Src/MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py
MBartkowiakSTFC: I just wanted to make sure: this code will _not_ create an instance of a widget every time it is executed, right? Because it looks like it would, but I expect that `match` has some special rules.

ChiCheng45 (commented):
Looks good, a few points from me

```
{"filter": "Butterworth", "attributes": {"n_steps": 20, "time_step_ps": 0.01, "order": 1, "attenuationn_type": "aa", "cutoff_freq": 45.0}}
```

If the key is made up or a typo like "attenuationn_type" the setting is still valid. Maybe we can make the configurator to reject these as well?


If I set a very large order, for example

```
{"filter": "Butterworth", "attributes": {"n_steps": 100, "time_step_ps": 0.01, "order": 10000, "attenuation_type": "lowpass", "cutoff_freq": 1.0}}
```

Then the job gets to 100%, but still fails with the error.

```
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 557, in run
    self.finalize()
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\TrajectoryFilter.py", line 194, in finalize
    filtered_coords = apply(
                      ^^^^^^
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\TrajectoryFilter.py", line 286, in apply
    output = filter.apply(component)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Mathematics\Signal.py", line 374, in apply
    return signal.sosfiltfilt(self.sos, input)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\...\AppData\Local\miniconda3\envs\MDANSE\Lib\site-packages\scipy\signal\_signaltools.py", line 5128, in sosfiltfilt
    edge, ext = _validate_pad(padtype, padlen, x, axis,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\...\AppData\Local\miniconda3\envs\MDANSE\Lib\site-packages\scipy\signal\_signaltools.py", line 4874, in _validate_pad
    raise ValueError(
ValueError: The length of the input vector x must be greater than padlen, which is 30003.
```

MBartkowiakSTFC:
@ChiCheng45 Thanks for the suggestions. The configurator will now try to apply the filter to an empty array of the same length as the trajectory. This way we can pick up potential errors caused by the filter parameters increasing the required minimum number of frames.

As for the attribute names, I decided to react with a _warning_ if an attribute is given which was not expected, or an expected one is missing.

ChiCheng45 (approved):
Looks good, works great.

PR #1178, opened 2026-02-17
Check for nan when saving plot limits

### LINKED ISSUE #1177: [BUG] Initial plot scaling is wrong for DCSF data with NaN values
**Description of the error**
Plots of the s(q,f) data shown here are not scaled correctly. The dataset contains `nan` values, which may be relevant.

All curves in `Single` mode should fit in the plot.
`Heatmap` plot colorbar should cover the range from minimum to maximum.

Curves in `Single` were visibly cropped. In `Heatmap` the maximum of the color bar was close to 0 intensity.

**Extra error**: in Python 3.14 switching to `Grid` caused a crash.

Image

Image

**Suggested fix**
Check if the problem is caused by `nan` values affecting the results of `min` and `max` operations.

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
This PR adds more checks for `nan` values, making improving the initial scaling of plots.

The crash experienced with python 3.14 could not be reproduced.

closes #1177

**Fixes**
1. Replaced `min` and `max` with `np.nanmin` and `np.nanmax` when saving the plot limits in `Single` and `Heatmap` plotters.

**To test**
Run a DCSF or DISF calculation. Make sure that the q vector widget warns you about some of the q shells being empty.
Try to plot the s(q,f) total result. The initial scaling of the `Single` plot and initial colorbar in `Heatmap` should both be correct (as opposed to their values in the main branch).

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Heatmap.py
@@ -50,7 +50,6 @@ class Heatmap(Plotter):
         self._backup_limits = {}
         self._initial_values = [0.0, 100.0]
         self._slider_values = [0.0, 100.0]
-        self._last_minmax = [-1, -1]
         self._slice_axis = 2
         self._plot_limit = 1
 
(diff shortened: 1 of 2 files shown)


oerc0122 (approved):
Looks good!

PR #1179, opened 2026-02-17
Add copyright checking to CI

oerc0122:
**Description of work**
Add automatic checking for presence of copyright notice in files.

Since this is under the `--preview` block, it is in a separate workflow to avoid enabling features which may be problematic.

To run manually, run `ruff check --preview --select "CPY"`

**Fixes**
N/A

**To test**
N/A

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,46 +10,27 @@ concurrency:
   group: lint-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  ruff_version: 0.13.0
+
 jobs:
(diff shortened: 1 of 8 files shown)


MBartkowiakSTFC:
Not sure why the lint_copyright test failed, but running the command manually on my machine worked fine.

I think the copyright test should run also on `MDANSE\Src` and not just `MDANSE_GUI\Src`. It seems to me that at the moment only the GUI code is covered.

Also, good to see that a few files were missing the copyright notice after all.

MBartkowiakSTFC (approved):

PR #1182, opened 2026-02-18
Prefer default atom properties over user-defined ones

MBartkowiakSTFC:
**Description of work**
Since we do not allow users to edit the atom properties in the GUI, we should also make sure that we don't use atom properties from already existing atom property files that may have been edited using earlier MDANSE versions.

**Fixes**
1. Added a `collections.ChainMap` to the atom database.
2. Added a warning to inform users that the default values will be used for standard atoms.

**To test**
Create a custom atom and change its properties.
Restart the GUI and check that the properties have been saved and loaded correctly.
Manually edit the atoms database in `~/.mdanse` and change properties of standard atoms (e.g. `H`, `O`, etc.)
Restart the GUI. The changes to standard atoms should not be visible.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/Databases.py
+++ b/MDANSE/Src/MDANSE/Chemistry/Databases.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 import copy
 import json
-from collections import defaultdict
+from collections import ChainMap, defaultdict
 from collections.abc import ItemsView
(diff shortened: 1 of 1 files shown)


oerc0122 (approved):
Looks good if it works as intended!

ChiCheng45 (approved):
Looks good, does the job.

PR #1186, opened 2026-02-20
Fix a minor issue in the lammps config file docs/doctests

oerc0122:
**Description of work**
Fix reference to old name which wasn't updated.

**Fixes**
N/A

**To test**
Doctest should pass

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Parsers/LAMMPSConfig.py
+++ b/MDANSE/Src/MDANSE/Framework/Parsers/LAMMPSConfig.py
@@ -452,13 +452,13 @@ class LAMMPSConfigFile(Parser, dict):
 
         Examples
         --------
-        >>> ConfigFileConfigurator._guess_type("0")
+        >>> LAMMPSConfigFile._guess_type("0")
         <class 'bool'>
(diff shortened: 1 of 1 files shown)


MDANSE/Src/MDANSE/Framework/Parsers/LAMMPSConfig.py
```diff
@@ -535,7 +535,7 @@ def atoms_parser(
 
         See Also
         --------
-        ConfigFileConfigurator._guess_atom_type : Mechanism for guessing atom type.
+        ~LAMMPSConfigFile._guess_atom_type : Mechanism for guessing atom type.
```
MBartkowiakSTFC, line 538: BTW what is the convention for using the `~` in the docstrings? I don't know if we should have it here or not.

oerc0122: `~` says "Don't print the whole "path" to the result. Since this is on the same class, I decided that we don't want the docs to display "LAMMPSConfigFile._guess_atom_type", just "_guess_atom_type"

MBartkowiakSTFC (approved):
Looks good.

PR #1190, opened 2026-02-24
Remove sys.exit() from ChemicalSystem

### LINKED ISSUE #1189: [BUG] `sys.exit` in ChemicalSystem
**Description of the error**
In ChemicalSystem there are two calls to `sys.exit` (ll. 476, 484) which violates how MDANSE usually handles errors.

These should probably raise sensibly and let other handlers handle this.

See above

**Suggested fix**
Remove sys.exit

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Removes unnecessary tests and exit calls from `ChemicalSystem`.

closes #1189

**Fixes**
1. Removes the `try`/`except` structure from `ChemicalSystem` dataset creation and lets it raise its own exceptions.

**To test**
All tests should pass.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -467,23 +467,10 @@ class ChemicalSystem:
         grp = h5_file.create_group("/composition")
         grp.attrs["name"] = self.name
 
-        try:
-            grp.create_dataset("atom_types", data=self._atom_types, dtype=string_dt)
-        except TypeError:
-            LOG.error(f"Bad array: {self._atom_types}")
-            import sys
-
-            sys.exit(1)
+        grp.create_dataset("atom_types", data=self._atom_types, dtype=string_dt)
         if self._atom_names is not None:
-            try:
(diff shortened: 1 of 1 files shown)


MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -467,23 +467,10 @@ def serialize(self, h5_file: h5py.File) -> None:
...
         grp.attrs["name"] = self.name
 
-        try:
-            grp.create_dataset("atom_types", data=self._atom_types, dtype=string_dt)
-        except TypeError:
-            LOG.error(f"Bad array: {self._atom_types}")
-            import sys
-
-            sys.exit(1)
+        grp.create_dataset("atom_types", data=self._atom_types, dtype=string_dt)
```
oerc0122, line 470: Do we still want to log the error or are we happy with the default message from h5py?

MBartkowiakSTFC: I think that the standard error message should be fine, since this is not an operation we expect to fail. So, we are not likely to say anything more meaningful about the problem than what h5py will tell us.

oerc0122 (approved):

PR #1191, opened 2026-02-26
Rework trajectory filter widget

oerc0122:
**Description of work**
Rework TrajectoryFilterWidget to support setting parameters similar to InstrumentResolution.

**Fixes**
- Add functions for turning strings into dict keys and back (although dict keys can be arbitrary)
- Fix typing in Signal
- Add `get_dep` function to find a dependency on a parent widget.
- Add constrained spinbox types using Qt mechanisms.

**To test**
GUI should work, be dynamic and tests should pass.

### DIFF
--- a/MDANSE/Src/MDANSE/IO/IOUtils.py
+++ b/MDANSE/Src/MDANSE/IO/IOUtils.py
@@ -29,7 +29,7 @@ from more_itertools import first_true, last, take, value_chain
 from MDANSE.MLogging import LOG
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Iterator, Sequence
+    from collections.abc import Callable, Iterable, Iterator, Sequence
 
 MAX_FILE_COUNT = 2048
 
@@ -92,6 +92,55 @@ def _(value: Path) -> dict[Any, Any]:
         raise FileNotFoundError(f"Unable to open ({value}) as json file .") from _
     except Exception as err:
         raise ValueError("Unable to load JSON string.") from err
+
+
+def standardise_name(name: str, transform: Callable[[str], str]=str.capitalize) -> str:
+    """Standardise dictionary keys to display name.
+
+    Parameters
+    ----------
+    name : str
+        String to standardise.
+    transform : Callable
+        String transformation function.
+
+    Returns
+    -------
+    str
+        Standardised name.
+
+    Examples
+    --------
+    >>> standardise_name("a_key")
+    'A key'
+    >>> standardise_name("another_key", transform=str.title)
+    'Another Key'
+    """
+    return transform(name.replace("_", " "))
+
+def destandardise_name(name: str) -> str:
+    """Returns key-like name (all lower, underscore-separated)
+
+    Parameters
+    ----------
+    name : str
+        Name to make key-like.
+
+    Returns
+    -------
+    str
+        Key-like name.
+
+    Examples
+    --------
+    >>> destandardise_name("My name")
+    'my_name'
+    >>> destandardise_name("a_string")
+    'a_string'
+    >>> destandardise_name("AaaAa")
+    'aaaaa'
+    """
+    return "_".join(map(str.lower, name.split()))
 
 
 def _strip_inline_comments(
--- a/MDANSE/Src/MDANSE/IO/IOUtils.py
+++ b/MDANSE/Src/MDANSE/IO/IOUtils.py
@@ -94,7 +94,9 @@ def _(value: Path) -> dict[Any, Any]:
         raise ValueError("Unable to load JSON string.") from err
 
 
-def standardise_name(name: str, transform: Callable[[str], str]=str.capitalize) -> str:
+def standardise_name(
+    name: str, transform: Callable[[str], str] = str.capitalize
+) -> str:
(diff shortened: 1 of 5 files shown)


MBartkowiakSTFC:
Thanks for the new widget design!

Not surprisingly, considering the extent of changes introduced into the code, there are still a few functional differences between the main branch and this branch. After the first test, my main observations are:
1. The spin box for the cutoff frequencies does not snap to the nearest value when I try to edit the text field. Instead it goes back to the value that was there before.
2. It is not possible to show the pre-calculated spectrum in the filter designer dialog. I get this error:
```
2026-02-26 16:15:18,315 - ERROR - process[99290] - main 36 - EXCEPTION:
<class 'KeyError'>
'External file (.mda)'
<traceback object at 0x312d36800>
  File "MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py", line 687, in update_filter
    self.render_canvas_assets()
  File "MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py", line 1029, in render_canvas_assets
    attenuation = self.get_attenuation_function(filter_preview, source)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MDANSE_GUI/InputWidgets/TrajectoryFilterWidget.py", line 759, in get_attenuation_function
    attenuation_fn = attenuation_examples[source]
                     ~~~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'External file (.mda)'
```
3. Minor thing: changing the filter type in the main GUI does not affect the filter type in the filter designer. I would expect the decisions made in the main GUI to transfer to the filter designer dialog.

MBartkowiakSTFC:
Thanks for the changes! All the problems from my previous comment have been solved.

In principle, the current version is usable and works correctly. The "Use setting" and "close" buttons do the same thing, so it should be enough to keep one of them.

I am happy to wait for opinions from other reviewers. I am not sure if we need to synchronise the other main GUI widgets with the filter designer dialog widgets or not. In any case, this is a minor detail.

oerc0122:
> Thanks for the changes! All the problems from my previous comment have been solved.
> 
> In principle, the current version is usable and works correctly. The "Use setting" and "close" buttons do the same thing, so it should be enough to keep one of them.
> 
> I am happy to wait for opinions from other reviewers. I am not sure if we need to synchronise the other main GUI widgets with the filter designer dialog widgets or not. In any case, this is a minor detail.

Do we want the close button changing to "do nothing" or one removing?

MBartkowiakSTFC:
> 
> Do we want the close button changing to "do nothing" or one removing?

Were you thinking of two buttons with functions "accept changes and close" and "reject changes and close"? I'm fine with this. One note: overall I want the default behaviour to be to use the NEW values, while rejecting the new values should be an exception. So, if we just close the dialog without pressing any buttons, it should still use the new values. This is the behaviour in the main branch at the moment.

(It's my personal preference as a GUI user that the GUI should use the parameters I took time to input, without me having to say "and I mean it" every time.)

oerc0122:
> > Do we want the close button changing to "do nothing" or one removing?
> 
> Were you thinking of two buttons with functions "accept changes and close" and "reject changes and close"? I'm fine with this. One note: overall I want the default behaviour to be to use the NEW values, while rejecting the new values should be an exception. So, if we just close the dialog without pressing any buttons, it should still use the new values. This is the behaviour in the main branch at the moment.
> 
> (It's my personal preference as a GUI user that the GUI should use the parameters I took time to input, without me having to say "and I mean it" every time.)

Done, there's only one button now.

ChiCheng45 (commented):
Looks good, just one thing so far. It can get stuck in a bad configuration state which doesn't go away. I think the error is not getting cleared. I think it should give an error when bad frequency cutoff and etc are chosen prior to running the job as well if possible.

Image

MBartkowiakSTFC:
> Looks good, just one thing so far. It can get stuck in a bad configuration state which doesn't go away. I think the error is not getting cleared. I think it should give an error when bad frequency cutoff and etc are chosen prior to running the job as well if possible.
> 

I think we had a similar problem with `QVectorsWidget`. One way to trigger the clearing/updating of a widget's error state is to call the `updateValue` method. I see that the widgets of `QVectorsWidget` are connected to its `updateValue`. Since the filter widget is similar (that is, it uses a helper dialog to change the parameters), maybe a similar solution would help here.

ChiCheng45 (commented):
A few more issues. 

When I have the DFTB trajectory loaded and set the bound frequency to 50 THz with bandstop or bandpass the GUI can get into a bad state.

The plot in the helper stops updating.

Image

The GUI gets into the purple state and stays there with bandstop.

Image

Some other issues, bound frequency box can be changed even though a bandpass or bandstop is not selected.

Image

ChiCheng45 (commented):
One more issue I've stumbled on.

The error state is not refreshed when the filter type is changed. So that the filter setting can be valid, but the widget says it is not, and vice versa.


Image

ChiCheng45 (approved):
Looks good to me.

MBartkowiakSTFC (approved):
I'm happy with the changes too. Looks good.

PR #1192, opened 2026-03-03
Use "core" driver for trajectories on data analysis workspaces

MBartkowiakSTFC:
**Description of work**
These changes are meant to speed up trajectory file access on data analysis platforms without changing anything on ordinary computers.

Since the filesystems we are looking at are likely to cache the file after it has been loaded initially, it may be that the load operation in the GUI will place the trajectory file in the cache and speed up the subsequent analysis as well.

**Fixes**
1. The changes are only applied if an environment variable MDANSE_FILE_IN_MEMORY is set to a non-zero value.
2. The GUI and the CLI will use `h5py.File(..., driver='core')` when loading the trajectory to read information.
3. The analysis jobs will still use the default driver, not to risk loading large files into memory completely for every subprocess.

**To test**
Set MDANSE_FILE_IN_MEMORY to 1.
Run the GUI and load some trajectories. There should be no change to the results. The memory requirements of running the GUI may increase with the number and size of trajectories.

### DIFF
--- a/MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
+++ b/MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
@@ -157,14 +157,16 @@ class Trajectory:
     such as the atom selection, atom transmutation and grouping.
     """
 
-    def __init__(self, filename, trajectory_format: ValidFormats | None = None):
+    def __init__(self, filename, trajectory_format: ValidFormats | None = None,
+                 hdf5_driver: str | None = None):
         self._filename = filename
+        self._hdf5_driver = hdf5_driver
         self._format = (
             trajectory_format if trajectory_format else self.guess_correct_format()
         )
 
         if self._format not in {"mock"}:
-            self._trajectory = self.open_trajectory(self._format)
+            self._trajectory = self.open_trajectory(self._format, self._hdf5_driver)
         self._min_span = None
         self._max_span = None
         self._grouping_level = GroupingLevels.ATOM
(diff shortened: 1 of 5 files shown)


MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -54,6 +55,27 @@
...
+    """
+    try:
+        env_var = os.environ["MDANSE_FILE_IN_MEMORY"]
+    except KeyError:
+        return False
+    try:
+        flag_value = int(env_var)
+    except (ValueError, TypeError):
+        return False
+    return flag_value > 0
```
```suggestion
    env_var = os.environ.get("MDANSE_FILE_IN_MEMORY", "0")
    return env_var.isdigit() and int(env_var)
```

MDANSE/Src/MDANSE/Scripts/mdanse.py
```diff
@@ -15,6 +15,7 @@
 #
 from __future__ import annotations
 
+import os
```
oerc0122: Not sure you need `os` here anymore.


```suggestion
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/TrajectoryModel.py
```diff
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import itertools
+import os
```
```suggestion
```

oerc0122 (changes requested):
Looks reasonable, but would it be best to pass a `bool` into trajectory (`use_core_driver`)?

MBartkowiakSTFC:
> Looks reasonable, but would it be best to pass a `bool` into trajectory (`use_core_driver`)?

Since there are several possible drivers (https://docs.h5py.org/en/stable/high/file.html#file-drivers), I thought that it would be OK for now to add a way to pick a specific one (in case it turned out that we need to use different ones in different parts of the code). But it is likely that "core" will turn out to be the only one we need, apart from the default one.

oerc0122 (approved):
Looks good

PR #1194, opened 2026-03-04
Update plot limits when changing axes

### LINKED ISSUE #1193: [BUG] Use it? does not appear to work as expected.
**Description of the error**
I have a sqw result of size (69, 1001). When I plot the results, it loads up with the default main axis as omega, the default Use it? setting is 0:69:1. When I switch to the main axis to q, the Use it? settings is still 0:69:1, this is fine, but when I try to change it to 69:70:1 then it fails to plot the results.

MBartkowiakSTFC:
**Description of work**
Limits of data slices in the plotter now depend on the choice of the main axis.

closes #1193

**Fixes**
1. SingleDataset uses the main axis information when checking slice limits.
2. Extra error output from the plotter when loading files.
3. Minor changes based on `ruff check` output.

**To test**
Plot a 2D data set.
Change the main axis. The limits in the "use it?" field should be applied correctly.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
@@ -181,7 +181,7 @@ class ViewerControls(QWidget):
             layout.addWidget(temp)
             self._buttons[button_name] = temp
         self._buttons["stop"].clicked.connect(self.stop_animation)
-        self._buttons["play"].clicked.connect(lambda: self.animate())
+        self._buttons["play"].clicked.connect(lambda: self.animate(step_size=1))
         self._buttons["fwd"].clicked.connect(lambda: self.animate(step_size=10))
         self._buttons["back"].clicked.connect(lambda: self.animate(step_size=-10))
         self._buttons["start"].clicked.connect(lambda: self.go_to_end(last_frame=False))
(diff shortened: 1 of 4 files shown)


ChiCheng45 (approved):
Looks good to me. The issue is fixed now.

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
```diff
@@ -181,7 +181,7 @@ def createButtons(self, orientation: Qt.Orientation):
             layout.addWidget(temp)
             self._buttons[button_name] = temp
         self._buttons["stop"].clicked.connect(self.stop_animation)
-        self._buttons["play"].clicked.connect(lambda: self.animate())
+        self._buttons["play"].clicked.connect(lambda: self.animate(step_size=1))
```
oerc0122, line 184: If you just want to use the default

```suggestion
        self._buttons["play"].clicked.connect(self.animate)
```

oerc0122 (approved):

MDANSE_GUI/Src/MDANSE_GUI/MolecularViewer/Controls.py
MBartkowiakSTFC: It's true that it works either way. But I'll keep this version, since it makes the new ruff complain less, without affecting the way this button works.

PR #1197, opened 2026-03-18
Only apply origin offset once in LAMMPS converter

MBartkowiakSTFC:
**Description of work**
The results of the 1st tutorial in MDANSE-Examples have changed since we created the tutorial.

Screenshot 2026-03-18 at 14 25 57

The additional spikes in the temperature were not there before.

This PR is a compromise, since the spikes in temperature are removed, but the animation of the trajectory slowly loses the synchronisation between the unit cell and atom positions.

Last frame of Mo trajectory:

Screenshot 2026-03-18 at 17 33 03

**Fixes**
1. Instead of updating the value of `origin` every step, we always shift the coordinates by the `first_origin` value from initial frame.

**To test**
Go through the tutorial 1 in MDANSE-Examples. The temperature results should be the same as in the tutorial description.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Converters/LAMMPS.py
+++ b/MDANSE/Src/MDANSE/Framework/Converters/LAMMPS.py
@@ -15,7 +15,7 @@
 #
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Literal
+from typing import Any, Literal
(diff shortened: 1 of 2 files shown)


ChiCheng45 (approved):
Looks good, issue is fixed.

PR #1199, opened 2026-03-19
Adjust unit conversion for LAMMPS velocities and forces

### LINKED ISSUE #1198: [BUG] lammps velocities look incorrect
**Description of the error**
When I dump lammps trajectory with the following command in lammps

```
dump           dumpTraj all custom 1 argon_8_long.txt id type x y z vx vy vz
``` 

then run a temperature calculation with different levels of interpolation. I find that the temperature with interpolation 0 which uses the save velocities are far too small. Here is the results on protos.

Image

The temperature of the MD simulation was set to 85K. So interpolation 3 and 5 appear to give the correct result, but it seems like there's some issue with the unit conversion when the velocities are saved.

MBartkowiakSTFC:
**Description of work**
The LAMMPS converter is able to read velocities and forces, but does not always convert them to the MDANSE internal units. This PR was created to correct this.

closes #1198

**Fixes**
1. Create conversion factors for velocity and force arrays in the LAMMPS converter,
2. Add missing cgs units to `units.json`.

**To test**
Convert a LAMMPS trajectory containing velocities. Calculate the temperature of the system using velocities and using position interpolation. The results should be similar.


oerc0122 (approved):
Looks good

ChiCheng45 (approved):
Looks good to me. Interpolation 0 and 3 agree with each other now.

Image

PR #1200, opened 2026-03-20
Adjust log level for SMARTS pattern match error

ChiCheng45:
**Description of work**
Adjusting the log level from error to warning since the failing bond order does not affect anything other than the SMARTS pattern matching in the atom selection. Cleared up the log messages a bit more.

**To test**
None


MBartkowiakSTFC (approved):
Looks good. I'm getting warnings instead of errors, as expected.

PR #1202, opened 2026-03-24
Improve QVectorsWidget warning highlighting

### LINKED ISSUE #1201: [BUG] q-vector generator widget warning highlight.
**Description of the error**
q-vector widget warning highlighting is not updating.

**Additional details**

Image

MBartkowiakSTFC:
**Description of work**
`QVectorsWidget` has an extra feature of setting and clearing errors if the inputs in the vector parameter table are invalid. This needed to be updated to account for the possibility of a warning being set in the underlying configurator.

closes #1201

**Fixes**
1. Reset the warning after clearing the input errors in the widget.

**To test**
Try to reproduce the problem from the issue. The warning should stay marked in the GUI.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/QVectorsConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/QVectorsConfigurator.py
@@ -69,7 +69,8 @@ class QVectorsConfigurator(IConfigurator):
         self._original_input = value
 
         trajConfig = self.configurable[self.dependencies["trajectory"]]
-        self.error_status = "NONE"
+        self.error_status = "OK"
+        self.warning_status = ""
         try:
             if not isinstance(value, tuple):
                 raise Exception(f"Q vectors setting must be a tuple {value}")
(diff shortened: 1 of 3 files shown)


ChiCheng45 (commented):
Its better but there is still the issue where the highlighting can get cleared when you change some other option e.g. weights.

Image

MBartkowiakSTFC:
> Its better but there is still the issue where the highlighting can get cleared when you change some other option e.g. weights.
> 

Thanks for catching this one. I'll check if I can fix it too.

MBartkowiakSTFC:
I think this should work now.

ChiCheng45:
One last thing.

image

The run button is disabled, which is right, but highlighted yellow. It's fine, but usually it's not highlighted yellow when other configurators are in an error state. Happy to approve if you think its ok.

image

MBartkowiakSTFC:
The **RUN!** button should be yellow whenever we have a warning in any of the widgets. It should be disabled when we have a widget with an error. So far, so good.

The q vectors widget creates a warning if one or more of the q-vector shells are empty, and an error if they are all empty. So, it is not possible for it to trigger this specific error without triggering this warning as well. I checked that invalid inputs in the parameter table trigger an error without a warning.

So, I think it's OK.

ChiCheng45 (approved):
Looks good.

PR #1203, opened 2026-03-24
Attach the 'use it?' field to Text output

MBartkowiakSTFC:
**Description of work**
In the MDANSE plotter, we are able to select specific rows or columns in a 2D array. However, the same input field is ignored in the `Text` "plotter". This PR makes it possible to pick a subset of data in the `Text` viewer and save it to a CSV file instead of saving all the rows and columns.

**Fixes**
1. Modify `Text` plotter to use the `curves_vs_axis` method.
2. Minor type hint corrections.

**To test**
Load a 2D dataset and plot it. Select two specific rows  or columns using the `Use it?` field.
Send the same dataset to the `Text` plotter. Try to show the same rows/columns in the text. (Disable/enable preview to check if that still works correctly.)

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
@@ -496,7 +496,8 @@ class SingleDataset:
         self,
         index_tuple: list[int],
         axis_lookup: list[str],
-    ) -> str:
+        skip_text: bool = False,
+    ) -> str | float:
         """Get a meaningful label for a subset of data.
 
         Used when plotting 1D arrays out of a multidimensional array.
@@ -504,14 +505,16 @@ class SingleDataset:
         Parameters
         ----------
         index_tuple : list[int]
-            indices of the 1D data array position in the ND array
+            indices of the 1D data array position in the ND array.
         axis_lookup : list[str]
-            Names of the axes to use
+            Names of the axes to use.
+        skip_text : bool, optional
+            If set to true, omits the text parts of the label. By default False.
 
         Returns
         -------
-        str
-            A string label for the plot legend.
+        str | float
+            A string label for the plot legend or a number for Text plotter.
 
         """
         if self._n_dim < 2:
@@ -540,13 +543,16 @@ class SingleDataset:
                 picked_value = round(picked_value, 1)
 
             label += f"{axis_label}={picked_value} {axis_unit}, "
+            if skip_text:
+                return float(picked_value)
         return label.rstrip(", ")
 
     def curves_vs_axis(
         self,
         x_axis_details: tuple[str, str],
         max_limit: int = 1,
-    ) -> list[np.ndarray]:
+        skip_label_text: bool = False,
+    ) -> dict[int, np.ndarray]:
         """Prepare a set of curves for plotting.
 
         Parameters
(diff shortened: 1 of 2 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -496,22 +496,25 @@ def generate_curve_label(
         self,
         index_tuple: list[int],
         axis_lookup: list[str],
-    ) -> str:
+        skip_text: bool = False,
```
oerc0122, line 500:
```suggestion
        *,
        skip_text: bool = False,
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Models/PlottingContext.py
```diff
@@ -540,13 +543,16 @@ def generate_curve_label(
...
+            if skip_text:
+                return float(picked_value)
         return label.rstrip(", ")
 
     def curves_vs_axis(
         self,
         x_axis_details: tuple[str, str],
         max_limit: int = 1,
-    ) -> list[np.ndarray]:
+        skip_label_text: bool = False,
```
oerc0122, line 556:
```suggestion
        *,
        skip_label_text: bool = False,
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Text.py
```diff
@@ -267,18 +275,43 @@ def process_2D_data(
...
+        multi_curves = np.vstack(
+            list(
+                dataset.curves_vs_axis(
+                    (best_unit, best_axis), max_limit=curves_limit, skip_label_text=True
+                ).values()
+            )
+        )
         # Add corner nil
-        xaxis = prepend(0.0, new_axes[axis_numbers[1]].flat)
+        xaxis = prepend(0.0, new_axes[axis_numbers[1 if flip_array else 0]].flat)
```
oerc0122:
```suggestion
        xaxis = prepend(0.0, new_axes[axis_numbers[flip_array]].flat)
```
Should work

oerc0122 (approved):
Looks good if it works. Do we have any tests for the text dumpers?

ChiCheng45 (commented):
Works well. Just one thing. Can we change the first value in the table to be empty instead of zero? Just slightly confusing on first reading.

Image

MBartkowiakSTFC:
> Works well. Just one thing. Can we change the first value in the table to be empty instead of zero? Just slightly confusing on first reading.
> 

What would be a good placeholder to put there instead of a zero? The main reason for it being there is not to end up with the top axis misaligned with the array elements below. This only happens when you import the data into some other software like Origin.

ChiCheng45:
> > Works well. Just one thing. Can we change the first value in the table to be empty instead of zero? Just slightly confusing on first reading.
> 
> What would be a good placeholder to put there instead of a zero? The main reason for it being there is not to end up with the top axis misaligned with the array elements below. This only happens when you import the data into some other software like Origin.

can it be some non-numeric character like '_'?

MBartkowiakSTFC:
> 
> can it be some non-numeric character like '_'?

I think this is done now.

ChiCheng45 (approved):
Thanks for that. Looks good.

PR #1205, opened 2026-03-27
Add switch for setting vector weights to 1

MBartkowiakSTFC:
**Description of work**
For users who want to override the reciprocal space sampling, we add an additional configuration flag.

**Fixes**
1. Added new vector variable `force_equal_weights`,
2. Updated instrument profiles to store the new value.

**To test**
Try running the same DCSF analysis with and without weights override.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING
 from qtpy.QtCore import Signal, Slot
 from qtpy.QtGui import QDoubleValidator, QIntValidator, QValidator
 from qtpy.QtWidgets import (
+    QCheckBox,
     QComboBox,
     QGridLayout,
     QLabel,
@@ -166,6 +167,10 @@ class InstrumentDetails(QWidget):
             instance = VectorWidget(self)
             instance.set_value([0, 0, 0])
             instance.value_changed.connect(self.update_values)
+        elif widget == "QCheckBox":
+            instance = QCheckBox("", self)
+            instance.setChecked(False)
+            instance.checkStateChanged.connect(self.update_values)
         qlabel = QLabel(label, self)
         layout.addWidget(qlabel, next_row, 0)
         layout.addWidget(instance, next_row, 1)
@@ -192,6 +197,10 @@ class InstrumentDetails(QWidget):
     @Slot()
     def update_values(self):
         for key, value in self._widgets.items():
+            if hasattr(value, "isChecked"):
+                new_val = value.isChecked()
+                self._values[key] = new_val
+                continue
             try:
                 new_val = value.text()
             except AttributeError:
@@ -256,6 +265,9 @@ class InstrumentDetails(QWidget):
         self._current_instrument = None
         for key, widget in self._widgets.items():
             new_value = getattr(instrument_instance, key, "Nothing!")
+            if hasattr(widget, "setChecked"):
+                widget.setChecked(new_value)
+                continue
             try:
                 widget.setText(str(new_value))
             except AttributeError:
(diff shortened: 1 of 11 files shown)


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
```diff
@@ -192,6 +197,10 @@ def reset_qvector_combobox(self):
     @Slot()
     def update_values(self):
         for key, value in self._widgets.items():
+            if hasattr(value, "isChecked"):
```
oerc0122, line 200: What is the question you actually want to ask here? Do you care whether it has `isChecked`, or do you care whether it's a `QCheckBox`?


```suggestion
            if isinstance(value, QCheckBox):
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
```diff
@@ -256,6 +265,9 @@ def update_panel(self, instrument_instance):
         self._current_instrument = None
         for key, widget in self._widgets.items():
             new_value = getattr(instrument_instance, key, "Nothing!")
+            if hasattr(widget, "setChecked"):
```
oerc0122, line 268: As above

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
```diff
@@ -166,6 +167,10 @@ def create_widget(self, entry):
             instance = VectorWidget(self)
             instance.set_value([0, 0, 0])
             instance.value_changed.connect(self.update_values)
+        elif widget == "QCheckBox":
+            instance = QCheckBox("", self)
+            instance.setChecked(False)
+            instance.checkStateChanged.connect(self.update_values)
```
oerc0122, line 173: Outside the scope of this PR, really, but if we were to rewrite this as a `match`, it would look more like this, which suggests that we spend a lot of time ignoring value, is this expected?

```suggestion
        label = attribute.strip(" _")
        
        match entry:
            case (_, "QComboBox", value):
                instance = QComboBox(self)
                instance.addItems(getattr(SimpleInstrument, value))
                instance.currentTextChanged.connect(self.update_values)
            case (_, "QLineEdit", "float"):
                instance = QLineEdit(self)
                instance.setValidator(QDoubleValidator())
                instance.setText("1.0")
                instance.setPlaceholderText("1.0")
                instance.textChanged.connect(self.update_values)
            case (_, "QLineEdit", "int"):
                instance = QLineEdit(self)
                instance.setValidator(QIntValidator())
                instance.setText("1")
                instance.setPlaceholderText("1")
                instance.textChanged.connect(self.update_values)
            case (_, "QLineEdit", _):
                instance = QLineEdit(self)
                instance.setText("text")
                instance.textChanged.connect(self.update_values)
            case (_, "VectorWidget", _):
                instance = VectorWidget(self)
                instance.set_value([0, 0, 0])
                instance.value_changed.connect(self.update_values)
            case (_, "QCheckBox", _):
                instance = QCheckBox("", self)
                instance.setChecked(False)
                instance.checkStateChanged.connect(self.update_values)
```

oerc0122 (approved):
Just some small style suggestions and a query. Otherwise looks reasonable codewise.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/InstrumentDetails.py
MBartkowiakSTFC: I care if it has `isChecked`, because I'm about to call it. If some other, unrelated-but-checkable widget appears one day, we will be ready.

MBartkowiakSTFC:
As for the `match` statement, we should definitely re-write the code to use it. But not right now, or at least not in this PR.

oerc0122:
> As for the `match` statement, we should definitely re-write the code to use it. But not right now, or at least not in this PR.

Indeed, I did start with "Not for this PR", I didn't expect it would be merged, should probably have changed it to Python from suggestion.

PR #1206, opened 2026-03-27
Update qvector defaults

MBartkowiakSTFC:
**Description of work**
Extends documentation of the q-vector generation scheme and adjusts the default number of vectors used for sampling.

**Fixes**
1. New section of documentation on reaching convergence.
2. Default number of vectors per shell changed to 250 for lattice vector generators.

**To test**
Please read the text and point out all the mistakes.


Doc/pages/H_conv.rst
```diff
@@ -270,3 +270,106 @@ had produced a severely underestimated VACF value at :math:`t=0`. Additionally,
...
+For isotropic, disordered systems, the calculation results will be averaged over
+vectors with similar :math:`|Q|` and output as a function of :math:`|Q|`. Spherical
+averaging is most commonly applied, as many systems can be assumed to be fully isotropic.
+:ref:`qvectors-reference-SphericalLatticeQVectors` is normally used for calculating
+the quasielastic neutron scattering results in isotropic systems.
+
+Generating the vectors on a sphere
+is equivalent to assuming that all the possible scattering geometries are possible,
+or rotating the simulated system to calculate the scattering results for different
+orientations of the system. This allows to approximate the correct scattering
```
oerc0122:
```suggestion
orientations of the system. This allows the code to approximate the correct scattering
```

Doc/pages/H_conv.rst
```diff
@@ -270,3 +270,106 @@ had produced a severely underestimated VACF value at :math:`t=0`. Additionally,
...
+experiment results which would otherwise require a significantly larger simulation.
+A physical sample in a real-life experiment typically contains atoms in
+amounts close to :math:`10^23` (the order of magnitude of the Avogadro number), while
```
oerc0122:
```suggestion
A physical sample in a real-life experiment typically contains in
the order of :math:`10^23` atoms (the Avogadro number), while
```

Doc/pages/H_conv.rst
```diff
@@ -270,3 +270,106 @@ had produced a severely underestimated VACF value at :math:`t=0`. Additionally,
...
+many MD trajectories contain less than :math:`10^6` atoms. Therefore, an actual experiment will
+sample many more configurations of atoms than it is possible to sample in a single MD simulation.
+Performing the calculation on multiple Q vectors maximises the amount the information
+that can be extracted from an MD trajectory. At the same time, the computational effort
+of the calculation increases with every additional vector sampled. Also,
```
oerc0122: Ends abruptly?

Doc/pages/H_conv.rst
```diff
@@ -270,3 +270,106 @@ had produced a severely underestimated VACF value at :math:`t=0`. Additionally,
...
+Performing the calculation on multiple Q vectors maximises the amount the information
+that can be extracted from an MD trajectory. At the same time, the computational effort
+of the calculation increases with every additional vector sampled. Also, 
+
+Random vector generation
+------------------------
+
+For 1D, 2D and 3D averaging (which could be relevant to calculations on layered,
+nematic and completely disordered systems, respectively), there exist three
+vector generators which apply the same approach in different number of dimensions:
```
oerc0122:
```suggestion
vector generators which apply the same approach in different numbers of dimensions:
```

Doc/pages/H_conv.rst
```diff
@@ -270,3 +270,106 @@ had produced a severely underestimated VACF value at :math:`t=0`. Additionally,
...
+and a sphere, respectively.
+
+You will most likely need the *lattice* versions of these generators, which follow
+the same approach, but apply additional processing steps afterwards.
+
+Lattice vector generators
+-------------------------
+
+Once the random vector distribution has been created, the lattice vector generators
+round the HKL coordinates of all the vectors to the nearest integer values.
```
oerc0122:
```suggestion
round the **hkl** coordinates of all the vectors to the nearest integer values.
```
? 

I don't know if we have an agreed format for `hkl`, but since they're vectors this would make more sense to me? Not that important though...

oerc0122 (changes requested):

oerc0122 (approved):

MDANSE/Src/MDANSE/Framework/QVectors/CircularLatticeQVectors.py
```diff
@@ -52,7 +52,7 @@ class CircularLatticeQVectors(LatticeQVectors):
             "default": (0.0, 5.0, 0.5),
         },
     )
-    settings["n_vectors"] = ("IntegerConfigurator", {"mini": 1, "default": 50})
+    settings["n_vectors"] = ("IntegerConfigurator", {"mini": 1, "default": 250})
```
ChiCheng45: I personally think it should be much higher like at least 10000.

MBartkowiakSTFC: 10000 would be good for a production run. I am not sure if it is a good idea to _start_ with a production run. If it turns out that the user made some mistake on the way, I would prefer them to notice it after 10 minuter rather than 3 hours. Then they can increase the number of vectors. If they end up spending 3 hours + 10 minutes on the calculation in the end, it's probably still OK.

There is no right decision here. My position is that there is no way to set a single value of `n_vectors` that will always work for every system, and the user will have to assume responsibility for their own results sooner or later. Your default value would make it clear to the user what numbers are required to get high-quality results.

Do we have any other options? Can we make it clear to the user that 250 is OK to start with, but not enough for a serious run?

ChiCheng45: Maybe 1000?

MBartkowiakSTFC: Actually, why not? I'll change it, and update the instrument profiles as well, since they are using low numbers at the moment.

ChiCheng45 (approved):
Looks fine. I've adjusted the notation in the documentation you added so it follows https://mdanse.readthedocs.io/en/protos/pages/notation.html.

PR #1207, opened 2026-03-27
Fix input parsing in QVectorsWidget

MBartkowiakSTFC:
**Description of work**
A few fixes to make "False" evaluate as `False`.

**Fixes**
1. Modified BooleanConfigurator,
2. Added checks to QVectorsWidget.

**To test**
Try running DCSF or DISF. Change the true/false values of `force_equal_weights`. The value should be parsed correctly. Invalid input should trigger a widget error.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/BooleanConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/BooleanConfigurator.py
@@ -17,6 +17,19 @@ from __future__ import annotations
 
 from MDANSE.Framework.Configurators.IConfigurator import IConfigurator
 
+BOOL_MAPPING = {
+    True: True,
+    "true": True,
+    "yes": True,
+    "y": True,
+    "1": True,
+    False: False,
+    "false": False,
+    "no": False,
+    "n": False,
+    "0": False,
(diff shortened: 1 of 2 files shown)


MDANSE/Src/MDANSE/Framework/Configurators/BooleanConfigurator.py
```diff
@@ -57,8 +57,9 @@ def configure(self, value):
...
         self._original_input = value
 
-        if value not in self._shortCuts:
+        value = value.lower() if isinstance(value, str) else value
+        if value not in BOOL_MAPPING:
             self.error_status = "Input is not recognised as a true/false value."
         else:
             self.error_status = "OK"
-            self["value"] = self._shortCuts[value]
+            self["value"] = BOOL_MAPPING[value]
```
oerc0122, line 65: Would this be better implemented as a `get_value` or similar I am not necessarily sure it's the best thing to make a new module level varaible and duplicate this code. 

I think, however, for the purposes of a quick fix, let's get this in and debate later.

oerc0122 (approved):
I think we get this in and create an issue to clean this up once we know what we want to do.

PR #1209, opened 2026-03-31
Change how lattice based q-vectors and their weight are generated 

### LINKED ISSUE #806: [BUG] Does the SphericalLatticeQVectors generate q-vectors which are similar to vectors generated from a uniform sphere
SphericalLatticeQVectors generates q-vectors from the reciprocal cell and bins them based on their inverse lengths. The DCSF and other functions are averaged over all the q-vectors in each bin. This attempts to create DCSF and other functions that are spherically averaged. It assumes that the q-vectors were sampled uniformly from a sphere.

I'm wondering if there are cases when SphericalLatticeQVectors does not do a good job at making an approximation to q-vectors uniformly sampled from a sphere. Consider the reciprocal lattice of a tetragonal system where one of the lengths is very long. Here is an illustration,

Image

You can see that certain angles will not be sampled very well. Even if the system size is very large so that the point density of the above image is increased, I think you'd still be sampling some angles more than others. Here is the same reciprocal lattice of the above, but with the cell lengths increased by 2.

Image

This issue will only affect anisotropic system like solids.

ChiCheng45:
**Description of work**
Change how q-vectors and weights are generated there are now two variables `n_vectors` which control the number of lattice q-vectors to be selected and `n_samples` which is used to determine the weights of the q-vectors and to search for the lattice points. In general, `n_samples` should be large 100000 or more and does not increase the compute time much so setting this even higher should be ok. Also probably needs to be large for shells with larger volumes. `n_vectors` should also be large but does increase the compute time by quite a bit so should be as large as possible for the given machine that is used. 

Here is the general scheme for sphericallatticeqvectors.

1. Get all lattice q-vectors in a shell (using the random search method similar to protos).
2. Select a subset  of the q-vectors using farthest point sampling
3. If force_equal_weights is false then determine the weight by estimating the Voronoi volume (gaussian weighted around the q-vector center) of the spherical shell the q-vector occupies. Else set the weights to one.

Basically, this scheme gives the user more control over how many q-vectors they want to generate so it's something between the new and old scheme. Additionally, only the new scheme with weights produces results which are in consistent with itself (three trajectories but with different unit cell which are longer in x, y, or z) and the sphericalqvectors generator, and is also more stable with respect to reruns in my tests seen below.

**Fixes**
Fixes #806 see proof of fix in the comments below.

**To test**
DISF and DCSF and other scattering jobs should work as usual. Check using the linear, circular, and spherical lattice q-vector generators.


ChiCheng45:
Here are my tests using this basic system, which diffuses anisotropically. I generated the trajectory using this script, which can be converted using the ASE converter.

```python
import numpy as np

# -----------------------
# Parameters
# -----------------------
n_atoms = 1600
box = np.array([80.0, 40.0, 40.0])
# box = np.array([40.0, 80.0, 40.0])
# box = np.array([40.0, 40.0, 80.0])
n_steps = 2000

# Diffusion strengths (standard deviations of noise)
diff_x = 1.6
diff_y = 0.4
diff_z = 0.1

# -----------------------
# Initialize positions
# -----------------------
positions = np.random.rand(n_atoms, 3) * box

# -----------------------
# Output file
# -----------------------
output_file = "anisotropic_diffusion_box_x.xyz"

with open(output_file, "w") as f:
    for step in range(n_steps):

        f.write(f"{n_atoms}\n")
        lattice = f"{box[0]} 0 0 0 {box[1]} 0 0 0 {box[2]}"
        f.write(f'Lattice="{lattice}" Properties=species:S:1:pos:R:3 Time={step}\n')

        for i in range(n_atoms):
            x, y, z = positions[i]
            f.write(f"Ar {x:.6f} {y:.6f} {z:.6f}\n")

        # -----------------------
        # Brownian motion update
        # -----------------------
        noise = np.zeros_like(positions)
        noise[:, 0] = np.random.normal(scale=diff_x, size=n_atoms)
        noise[:, 1] = np.random.normal(scale=diff_y, size=n_atoms)
        noise[:, 2] = np.random.normal(scale=diff_z, size=n_atoms)

        positions += noise

print(f"Saved: {output_file}")
```

Traj X
image

Since this is a strongly anisotropic it should be difficult to converge results with respect to the number of q-vectors. Here I used sphericalqvector generator with the DISF calculations.

image

Will focus on the shell at around 20.0 with a width of 1.0, which has a total of 2556 q-vectors. Notice that due to the non-cubic unit cell that is used, the distribution of q-vectors is different along the different axes.

image
image

Here are my findings so far using the DISF calculation with the sphericallatticeqvector generator.

New scheme
- 'shells': [20.0, 20.0, 0.5]
- 'force_equal_weights': False
- 'n_samples': 100000
- 'width': 1.0 

(total number of vectors is 2556 for this shell)

100 vectors (3.9% of available vectors)
image

200 vectors (7.8% of available vectors)
image

400 vectors (15.6% of available vectors)
image

800 vectors (31.2% of available vectors)
image

1600 vectors (62.6% of available vectors)
image

New scheme
- 'shells': [20.0, 20.0, 0.5]
- 'force_equal_weights': True
- 'n_samples': 100000 (actually has no effect here since force weight is set to true)
- 'width': 1.0 

(total number of vectors is 2556 for this shell)

100 vectors (3.9% of available vectors)
image

200 vectors (7.8% of available vectors)
image

400 vectors (15.6% of available vectors)
image

800 vectors (31.2% of available vectors)
image

1600 vectors (62.6% of available vectors)
image

ChiCheng45:
Tests with the old scheme v2.0.0rc2.

- Traj X
- 'shells': [20.0, 20.0, 0.5]
- 'width': 1.0

100 vectors  (3.9% of available vectors)
image

200 vectors  (7.8% of available vectors)
image

400 vectors  (15.6% of available vectors)
image

800 vectors  (31.2% of available vectors)
image

1600 vectors  (62.6% of available vectors)
image

ChiCheng45:
Tests with the January scheme.

 January scheme
- 'shells': [20.0, 20.0, 0.5]
- 'width': 1.0
- 'force_equal_weights': False

150 vectors (112, 115, 110, 107, 102, 96, 101, and 98 unique vectors)
image

300 vectors (205, 213, 203, 215, 215, 219, 201, and 211 unique vectors)
image

600 vectors (407, 381, 398, 412, 389, 383, 394, and 391 unique vectors)
image

1350 vectors (803, 793, 796, 784, 761, 782, 806, and 825 unique vectors)
image

3500 vectors (1570, 1576, 1581, 1583, 1544, 1580, 1568, and 1570 unique vectors)
image

 January scheme
- 'shells': [20.0, 20.0, 0.5]
- 'width': 1.0
- 'force_equal_weights': True

150 vectors (107, 107, 106, 112, 108, 99, 107, and 107 unique vectors)
image

300 vectors (203, 193, 221, 198, 217, 208, 209, and 190 unique vectors)
image

600 vectors (398, 388, 395, 381, 393, 389, 391, and 393 unique vectors)
image

1350 vectors (778, 825, 779, 798, 758, 785, 795, and 807 unique vectors)
image

3500 vector (1581, 1587, 1569, 1557, 1557, 1541, 1589, and 1595 unique vectors)
image

ChiCheng45:
Comparison between Traj X, Traj Y, and Traj Z. I've made the same trajectory as the above, but with different boxes.

Traj X
image

Traj Y
image

Traj Z
image


Diffusion constants are the same, so the system is the same, and therefore, spherically averaged results should also be the same.


SphericalQVectors
- 'shells': [20.0, 20.0, 0.5]
- 'n_vectors': 51200
- 'width': 1.0
image
image


New scheme
- 'shells': [20.0, 20.0, 0.5]
- 'force_equal_weights': False
- 'n_samples': 100000
- 'n_vectors': 3200
- 'width': 1.0
image
image


New scheme
- 'shells': [20.0, 20.0, 0.5]
- 'force_equal_weights': True
- 'n_samples': 100000
- 'n_vectors': 3200
- 'width': 1.0
image
image


Old scheme
- 'shells': [20.0, 20.0, 0.5]
- 'width': 1.0
- 'n_vectors': 3200 
image
image

Jan scheme
- 'shells': [20.0, 20.0, 0.5]
- 'width': 1.0
- 'n_vectors': 100000
- 'force_equal_weights': False
image
image


Jan scheme
- 'shells': [20.0, 20.0, 0.5]
- 'width': 1.0
- 'n_vectors': 100000
- 'force_equal_weights': True
image
image



With Traj Z you get more points in the $q_z$ direction, and the smallest diffusion occurs along the $z$-axis. 
image
All methods, therefore, oversample this direction. Only the new scheme with weights corrects for this issue.

MBartkowiakSTFC:
I have been testing the new scheme; the results are looking good.

From a practical point of view, should we include n_samples in the instrument profile settings? At the moment I think that we should, but let's do it in a separate PR.

ChiCheng45:
> I have been testing the new scheme; the results are looking good.
> 
> From a practical point of view, should we include n_samples in the instrument profile settings? At the moment I think that we should, but let's do it in a separate PR.

Let's do it in this PR. I put in some values for the instruments.

MBartkowiakSTFC:
Screenshot 2026-04-16 at 12 40 05

One more question: the vector preview plot shows the "total" and "unique" vector counts per shell. In this vector generation scheme "total" and "unique" vectors seem to be the same every time. At the same time, we probably need to keep the vector statistics plotter the way it is now, so it still works for the files from version 2.0.0. What is your opinion: do we need to do anything about this plot, and is there anything we can do?

ChiCheng45:
> One more question: the vector preview plot shows the "total" and "unique" vector counts per shell. In this vector generation scheme "total" and "unique" vectors seem to be the same every time. At the same time, we probably need to keep the vector statistics plotter the way it is now, so it still works for the files from version 2.0.0. What is your opinion: do we need to do anything about this plot, and is there anything we can do?

Thanks, I have updated it so that it is now "used" and "found". "Used" is the number of q-vectors used in the calculation, its value is either the total number of q-vectors found in the shell or the total number of q-vectors requested by the user. "Found" is the total number of q-vectors found in the shell.

image

The older MDA files should still work as usual, but labels are changed from unique and total to used and found, which I think should still make sense.

MBartkowiakSTFC (approved):
I tried out the new vector plot labels on some old files. It seems to work fine.

Looks good.

PR #1210, opened 2026-03-31
Fix the vector dialog getting stuck

### LINKED ISSUE #1204: [BUG] Preview vector distribution can get into a stuck state.
**Description of the error**
By carrying out a few specific steps. Preview vector distribution can get into a stuck state where it no longer updates. It looks like it specifically occurs when you have a shell which starts from zero.

**Additional details**

Image

MBartkowiakSTFC:
**Description of work**
Adds the smallest possible change needed to prevent the vector dialog from getting stuck in an invalid state.

closes #1204

**Fixes**
1. Added a check for the rounding order being calculated as a very small number.

**To test**
Use default vector shell settings (including 0). Lower the shell width until an error is triggered. Increase the width again. The vector dialog should show valid values again.


ChiCheng45 (approved):
Seems to be fixed now.

PR #1213, opened 2026-04-13
Read H5MD time correctly and update test results

### LINKED ISSUE #1212: [ENHANCEMENT] Add support for fixed time format in H5MD trajectories
At the moment MDANSE can use H5MD trajectories which contain explicit time information. However, the H5MD standard defines also time storage based on a fixed time step size https://www.nongnu.org/h5md/h5md.html#fixed-step-and-time-storage

The `time` method should be extended to recognise which time format is used in the trajectory, and to load it correctly in each case.

N/A

**Additional context**
This will be important for software like MDMC.

MBartkowiakSTFC:
**Description of work**
Adds a mechanism for getting the time axis out of a scalar dataset in an H5MD file.

It turns out that the code was getting time values in steps of 1 ps if the time information was missing. For this reason, the reference results in the unit tests had to be adjusted.

closes #1212

**Fixes**
1. `H5MDTrajectory` recreates the time axis based on time step information and time offset.
2. Reference files for unit tests have been updated.

**To test**
Load the `Ar_mdmc_h5md.h5` trajectory. The GUI should show a valid time axis with ca. 150 fs per step.
Calculate the temperature of the system. It should be of the order of 100 K.


ChiCheng45 (approved):
Looks good to me.

oerc0122 (approved):

PR #1215, opened 2026-04-14
Reduce log output, remove unused code

MBartkowiakSTFC:
**Description of work**
Some of the log output is not useful to the user. Also, ASE converter outputs the cell parameters in a loop, which is unnecessary.

This PR removes some superfluous log messages.

**Fixes**
1. Remove several log calls,
2. Downgrade log calls from `info` to `debug` if the information is not relevant to the user,
3. Remove `GeneralInputs` and `InputFactory`, which contain a lot of log calls and are ultimately not used.

**To test**
Run an ASE converter job. There should be less log output.
No functional changes are expected.


MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Text.py
```diff
@@ -428,12 +428,10 @@ def clear(self, figure: QTextBrowser = None):
             optionally, new QTextBrowser to be used
 
         """
-        LOG.debug("Text.clear stared")
```
oerc0122, line 431: At least this ominous message has gone. Does it have paralysing gaze?

MDANSE_GUI/Src/MDANSE_GUI/ElementsDatabaseEditor.py
```diff
@@ -410,8 +410,8 @@ def write_to_database(self, item: QStandardItem):
         column = idx.column()
         row_name = self.verticalHeaderItem(row).text()
         column_name = self.horizontalHeaderItem(column).text()
-        LOG.info(f"data:{data}, text:{text}, row:{row}, column:{column}")
-        LOG.info(f"column name={column_name}, row name={row_name}")
+        LOG.debug(f"data:{data}, text:{text}, row:{row}, column:{column}")
```
oerc0122, line 413: These should really be converted to "%s" format, but I'm not that fussed.

oerc0122 (approved):
Looks good.

MDANSE_GUI/Src/MDANSE_GUI/ElementsDatabaseEditor.py
MBartkowiakSTFC: I'm happy to make a separate PR later specifically for formatting the log strings. There will be many that need fixing.

PR #1218, opened 2026-04-17
Enable deleting entries from the second dict in ChainMap

### LINKED ISSUE #1216: [BUG] Custom chemical elements can't be renamed or removed
**Description of the error**
Once a copy of a chemical element has been created, it seems impossible to rename it or delete it.

Normally, it has been possible to change the name of the element.  The element list would then be updated.

A deleted element disappears from the table in the GUI, but reappears after restarting the software.
The rename operation does not work.

The accompanying error message is:
`KeyError: "Key not found in the first mapping: 'Am243(copy)'"`

**Suggested fix**
This seems to be an error introduced when atom database started using the `ChainMap`.

**Additional details**
This error only occurred if MDANSE has been restarted after adding new elements. When a new element is first created, it can be renamed or deleted. To reproduce the error, copy an element and then exit MDANSE. Once MDANSE has been restarted, the copy will be impossible to rename or remove.

MBartkowiakSTFC:
**Description of work**
Since `AtomsDatabase` uses a `ChainMap` now, the custom atoms are in the first dictionary when they are created, but in the second dictionary once the database has been reloaded. The code deleting and renaming the atoms has to account for this. 

closes #1216

**Fixes**
If the atom entry is not in the first dictionary in `AtomsDatabase`, allow access to the second one (but not the ones beyond).

**To test**
Create new atoms in the element editor. Rename and delete them.
Create new atoms, then restart the GUI. Rename and delete them.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/Databases.py
+++ b/MDANSE/Src/MDANSE/Chemistry/Databases.py
@@ -749,8 +749,11 @@ class AtomsDatabase(_Database):
         """
         try:
             del self._data[symbol]
-        except KeyError as err:
-            raise AtomsDatabaseError(f"Atom {symbol} does not exist.") from err
+        except KeyError:
+            try:
+                del self._data.parents[symbol]
+            except KeyError as err:
+                raise AtomsDatabaseError(f"Atom {symbol} does not exist.") from err
 
     def remove_property(self, label: str):
         """Remove an atom property from the database.
@@ -787,7 +790,10 @@ class AtomsDatabase(_Database):
                 f"Cannot rename atom from {old_key} to {new_key} as {new_key}"
                 " already exists.",
             )
-        self._data[new_key] = self._data.pop(old_key)
+        try:
+            self._data[new_key] = self._data.pop(old_key)
+        except KeyError:
+            self._data[new_key] = self._data.parents.pop(old_key)
 
     def rename_atom_property(self, old_key: str, new_key: str):
         """Rename the atom property in the atom database.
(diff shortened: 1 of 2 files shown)


MDANSE/Src/MDANSE/Chemistry/Databases.py
```diff
@@ -749,8 +749,11 @@ def remove_atom(self, symbol: str):
         """
         try:
             del self._data[symbol]
-        except KeyError as err:
-            raise AtomsDatabaseError(f"Atom {symbol} does not exist.") from err
+        except KeyError:
+            try:
+                del self._data.parents[symbol]
+            except KeyError as err:
+                raise AtomsDatabaseError(f"Atom {symbol} does not exist.") from err
```
oerc0122, line 756: Does this look cleaner? I'm not too sure. It'd handle arbitrary length `ChainMap`s, but we don't have those.
```python
        for d in self._data.maps:
            with suppress(KeyError):
                del d[key]
                break
        else:
            raise AtomsDatabaseError(f"Atom {symbol} does not exist.") ```

oerc0122 (approved):
I feel there should be a flatter way, but that's the best of my thoughts and it still looks ugly. Happy for this.

MDANSE/Src/MDANSE/Chemistry/Databases.py
MBartkowiakSTFC: I think that, at least at the moment, our `ChainMap` is expected to have three layers. The deepest one is the original atom database, and the user is not supposed to modify that one. So, just to be on the safe side I would probably not go through _all_ the layers. Having said that, the GUI is supposed to prevent you from editing atoms that belong to the original file, so it probably would not matter either way.

oerc0122: I'm not convinced it's any clearer in intent either. I think what you've got is fine.

MBartkowiakSTFC: Thanks. I'd better fix the failing unit test, then.

PR #1219, opened 2026-04-20
Add checks for uneven time steps in frames and LAMMPS

### LINKED ISSUE #1217: [BUG] Wrong time step length when importing certain LAMMPS trajectories
**Description of the error**
When importing LAMMPS trajectories, MDANSE uses the time step difference between frames 0 and 1 as reference. LAMMPS often inserts one additional, significantly shorter steps into the trajectory file after frame 0, resulting in a timestep sequence like 0,166,1000,2000,... 
MDANSE will convert this to 0,166,332,498,... leading to wrong results.

Either correct time step sequence of 0,166,1000,2000,... or deleting the outlier and getting time steps 0,1000,2000,...

Trajectories with effective time steps like 0,166,332,... leading to wrong results whenever simulation time is involved.

**Suggested fix**
Either calculating the actual time for every frame by multiplying its time step with the time step length or (probably easier) ignoring/deleting the outlier on import.

MBartkowiakSTFC:
**Description of work**
Recently, it was reported that invalid trajectories can be passed through the LAMMPS converter, leading to analysis problems later. The trajectories in question contained frames corresponding to different time intervals, and the resulting .mdt files could not be easily fixed using `TrajectoryEditor`.

This PR is meant to address the issue. LAMMPS converter will log an error if the time step between consecutive frames changes in the input file. An .mdt file containing frames with different time steps will produce a warning in the GUI. The reason why the conversion does not fail immediately is that the changes in this PR make it possible to remove the invalid frames using `TrajectoryEditor` or to omit them by setting the initial frame index to a non-zero value.

closes #1217

**Fixes**
1. `HDFTrajectoryConfigurator` shows a warning if the time steps in the input trajectory are not equal,
2. `FramesConfigurator` takes the time values from the trajectory instead of calculating them assuming an even step,
3. LAMMPS converter logs an error (but does **not** fail) if the time step in the input trajectory changes between frames.

**To test**
Try re-running analysis on known, correct trajectories using different values of `start, stop, step` in the frames configurator. The results should not change.
Convert a LAMMPS trajectory with additional frames which do not match the normal writeout interval. An error should be logged.
Load the trajectory and select `TrajectoryEditor`. A warning should be shown in the GUI. Remove the invalid frames; the time step in the output trajectory should be correct.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Configurators/FramesConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/FramesConfigurator.py
@@ -97,12 +97,8 @@ class FramesConfigurator(RangeConfigurator):
 
         self["n_frames"] = self["number"]
 
-        self["time"] = trajConfig["md_time_step"] * self["value"]
+        self["time"] = trajConfig["time_axis"][self["value"]]
 
-        # case of single frame selected
-        try:
-            self["time_step"] = self["time"][1] - self["time"][0]
-        except IndexError:
-            self["time_step"] = 1.0
+        self["time_step"] = trajConfig["md_time_step"]
 
         self["duration"] = self["time"] - self["time"][0]
--- a/MDANSE/Src/MDANSE/Framework/Configurators/HDFTrajectoryConfigurator.py
+++ b/MDANSE/Src/MDANSE/Framework/Configurators/HDFTrajectoryConfigurator.py
@@ -15,9 +15,13 @@
 #
 from __future__ import annotations
 
+import numpy as np
+
 from MDANSE import PLATFORM
 from MDANSE.Framework.Configurators.InputFileConfigurator import InputFileConfigurator
 from MDANSE.MolecularDynamics.Trajectory import Trajectory
+
+TIME_STEP_TOL = 1e-8
 
 
 class HDFTrajectoryConfigurator(InputFileConfigurator):
@@ -57,6 +61,8 @@ class HDFTrajectoryConfigurator(InputFileConfigurator):
         if value == self._original_input:
             return
         self._original_input = value
+        self.error_status = "OK"
+        self.warning_status = ""
 
         InputFileConfigurator.configure(self, value)
         try:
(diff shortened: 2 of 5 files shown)


ChiCheng45:
@MBartkowiakSTFC 

Everything seems to work, For

"A warning should be shown in the GUI. Remove the invalid frames; the time step in the output trajectory should be correct."

So the user will have to remove the invalid frame by hand if it's in an awkward place? e.g.

0, 1, 2, 3, 4, 5, 6, 7, ..., 10000, 10000.5, 10001, ...

I guess it's at the start, they can just use the trajectory editor to remove a few frames, and it will work.

ChiCheng45 (approved):

MBartkowiakSTFC:
> 
> So the user will have to remove the invalid frame by hand if it's in an awkward place? e.g.
> 
> 0, 1, 2, 3, 4, 5, 6, 7, ..., 10000, 10000.5, 10001, ...
> 
> I guess it's at the start, they can just use the trajectory editor to remove a few frames, and it will work.

From my perspective, there is probably something fundamentally wrong with your trajectory if you ended up having extra frames somewhere in the middle of it. The user will most likely need to do much more than just remove a few frames.

So, the main goal was not to accept the bad trajectory silently but to inform the user instead. At the same time, not highlighting an error here was meant to leave the option open to run the Trajectory Editor.

I'm happy to merge the current changes, but if we come up with a better idea later, then we can make further changes.

ChiCheng45:
> > So the user will have to remove the invalid frame by hand if it's in an awkward place? e.g.
> > 0, 1, 2, 3, 4, 5, 6, 7, ..., 10000, 10000.5, 10001, ...
> > I guess it's at the start, they can just use the trajectory editor to remove a few frames, and it will work.
> 
> From my perspective, there is probably something fundamentally wrong with your trajectory if you ended up having extra frames somewhere in the middle of it. The user will most likely need to do much more than just remove a few frames.
> 
> So, the main goal was not to accept the bad trajectory silently but to inform the user instead. At the same time, not highlighting an error here was meant to leave the option open to run the Trajectory Editor.
> 
> I'm happy to merge the current changes, but if we come up with a better idea later, then we can make further changes.

I think it's good to merge.

PR #1220, opened 2026-04-21
Improve vector statistics output

MBartkowiakSTFC:
**Description of work**
Improvements to the way the vector distribution is presented.
New vector generation mechanism made it necessary to adjust the violin plot parameters.
Average |q| is written into the output files and can be compared to the nominal |q|.

**Fixes**
1. Weights used for generating multiple copies of vectors for the violin plot are now scaled by a constant before converting to `int`, reducing the fraction of weights that would otherwise be rounded to 0.
2. `average_q` is a new dataset written into .mda files, containing the weighted average of |q| of all the vectors in the shell.

**To test**
Run DCSF or DISF analysis. Check that the new dataset (`average_q`) is written into the output files and can be plotted.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/QVectors/IQVectors.py
+++ b/MDANSE/Src/MDANSE/Framework/QVectors/IQVectors.py
@@ -81,6 +81,35 @@ def truncated_normal_distribution(
     )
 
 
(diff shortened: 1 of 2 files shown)


ChiCheng45 (approved):
Looks good to me.

PR #1221, opened 2026-04-21
Adjust fpsampling in q-vector generator

ChiCheng45:
**Description of work**
Adjust the fpsampling method used so q-vectors are normalized before selecting them. This should avoid selecting q-vectors on the edges of the shell. Distances between q-vectors are now determined based on the angle between them.

**To test**
DCSF and DISF calculations with the lattice-based q-vector generator should work as usual.


ChiCheng45:
@MBartkowiakSTFC there was some issue with the changes. should be sorted now.

MBartkowiakSTFC (approved):
Looks good. Definitely a better agreement between the average and nominal q values.

PR #1222, opened 2026-04-22
Handle different formats of vector statistics in the plotter

MBartkowiakSTFC:
**Description of work**
Some DCSF or DISF result files can fail to plot the vector statistics if empty shells are included in the output. This PR aims to support both the old files (where the empty shells are skipped) and the new ones (where the empty shells are shown with vector count 0).

**Fixes**
1. Try to match the length of the q axis to the length of the vector count dataset.

**To test**
Load old files with empty shells and plot vector statistics.
Generate a DCSF file with empty vector shells and plot vector statistics.
Check that the vector preview still plots correctly if some of the vector shells are empty.


ChiCheng45 (approved):
Looks good. Vectory summary with empty shell fixed now.

PR #1223, opened 2026-04-22
Speed up scattering job loading in GUI

ChiCheng45:
**Description of work**
Changed the default q-vectors setting so that the q-vectors are generated more quickly. Blocked signals when the q-vector table background colours are changed to stop q-vectors from being regenerated.

Selecting a scattering job that uses the q-vector generator should be faster.

**To test**
Run scattering jobs with different q-vector settings. Jobs should work as normal.


MBartkowiakSTFC (approved):
The time needed to run the GUI unit tests has decreased a lot. Looks good.

Issue #1154, opened 2026-02-03
[BUG] Visual bug in plot holder when scrolling


ChiCheng45:
**Description of the error**
Strange visual bug seen on Mac. This occurs when I use the left/right scrolling with the Mac track pad.

**Additional details**

![Image](https://github.com/user-attachments/assets/dd1d9e01-2847-4127-9464-f551ed2b3d5b)

MBartkowiakSTFC:
So far I can confirm that the same problem occurs on an M1 Mac, also after updating PyQt6 to a later version.

I have not found a relevant Qt bug report so far, but my working assumption is that this is not an MDANSE-specific problem.

ChiCheng45:
> So far I can confirm that the same problem occurs on an M1 Mac, also after updating PyQt6 to a later version.
> 
> I have not found a relevant Qt bug report so far, but my working assumption is that this is not an MDANSE-specific problem.

Yea it looks like this is a Qt issue.

https://qt-project.atlassian.net/browse/QTBUG-134434

PR #1224, opened 2026-04-22
Increase the PyQt6 version to 6.8.1

MBartkowiakSTFC:
**Description of work**
We are switching from `PyQt6==6.7.0` to `6.8.1` as a compromise between supporting platforms with dated libraries and Python versions, and trying to stay up to date and using software that is still supported.

**Fixes**
1. Allowed PyQt6 version changed from 6.7.0 to 6.8.1.

**To test**
Please update the PyQt6 version on your platform and check if the GUI still runs correctly.

On a Mac I noticed at least one new glitch: in the trajectory view tab, scrolling up and down within the tab makes the 3D view move outside of its area:

Screenshot 2026-04-22 at 16 49 51

### DIFF
--- a/MDANSE_GUI/pyproject.toml
+++ b/MDANSE_GUI/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
     "matplotlib",
     "qtpy",
     "vtk>=9.4.0",
-    "PyQt6<=6.7.0",
+    "PyQt6<=6.8.1",
     "PyYAML",
     "tomlkit"
 ]


ChiCheng45:
Do you still see a bug with an even newer version of pyqt6? I don't see the issue on Windows.

MBartkowiakSTFC:
So far, here are my test results by version:
6.10.0 - 6.11.0 - hangs when opening any VTK view,
6.9.1 - still has this glitch.

MBartkowiakSTFC:
By the way, the highest version of PyQt6 that I can run on my computer, 6.9.1, is still affected by #1154.

ChiCheng45:
I see, I will also have a look to see if the issue occurs on my Intel Mac when I get home. 

Also I wonder if the visual bug in #1154 and above goes away if we set 

https://github.com/ISISNeutronMuon/MDANSE/blob/a092ce8fff82e3c9507eede10fdae8cffaf0629f/MDANSE_GUI/Src/MDANSE_GUI/main.py#L82

to something other than "Fusion". Apparently, according to https://qt-project.atlassian.net/browse/QTBUG-123038 the close button bug in #1154 is specifically when the stylesheet is set to Fusion.

MBartkowiakSTFC:
I checked it. On my machine I have three possible styles: "macOS", "Windows" and "Fusion". The bug from #1154 is present in both "Windows" and "Fusion". "macOS" fails to start, apparently because in "macOS" the tabs don't have a `close` button, and our code relies on it being there.

ChiCheng45:
So on my Intel Mac 6.8.1, 6.9.1 work fine. 6.10.0 and 6.11.0 hangs when I open the trajectory viewer tab. I wonder if there is a general issue with vtk pyqt6 and macOS.

MBartkowiakSTFC:
This discussion here mentions some general problems with QVTKRenderWindowInteractor and software that creates its own 3D/OpenGL context:
https://discourse.vtk.org/t/qvtkrenderwindowinteractor-and-qwebengineview-conflict/15408

In any case, if PyQt6 6.8.1 works, we can probably use it. There are no issues that would make the software unusable. But we have to start thinking about a long-term plan, if later versions cause problems. Probably we'll start by reporting it to VTK and PyQt developers.

ChiCheng45:
@oerc0122 can you test different pyqt6 versions on linux?

MBartkowiakSTFC:
So far I managed to test MDANSE_GUI on a laptop with an Intel GPU, running MXLinux Infinity, kernel 6.12, and Python 3.13.

PyQt6 versions 6.7.0, 6.9.1, 6.10.1 and 6.11.0 all worked correctly, without hanging, crashing or showing any of the glitches mentioned above.

MBartkowiakSTFC:
This has been replaced by #1226.

PR #1227, opened 2026-04-29
Create MockTrajectory from an input JSON string

MBartkowiakSTFC:
**Description of work**
This does not affect the normal MDANSE workflow, but will be useful when using MDANSE as a library in other software.

**Fixes**
1. Allow passing a JSON string to `MockTrajectory.from_json` as an alternative to passing a file name of a file containing the JSON string.

**To test**
All tests should pass.


ChiCheng45 (approved):
Looks good to me.

PR #1228, opened 2026-05-05
Replace CASTEP_MD parser with castep_outputs

oerc0122:
**Description of work**
Replace custom parser with library designed for purpose.

**Fixes**
N/A

**To test**
All conversion tests should work as normal.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Converters/CASTEP.py
+++ b/MDANSE/Src/MDANSE/Framework/Converters/CASTEP.py
@@ -116,11 +116,11 @@ class CASTEP(Converter):
         # Read the information in the frame
         time_step = frame["time"]
 
-        unit_cell = UnitCell(np.vstack(frame["h"]))
-        coords = np.vstack(tuple(arr[1] for arr in frame["R"]))
+        unit_cell = UnitCell(frame["h"])
+        coords = frame["R"]
         variables = {
-            "velocities": np.vstack(tuple(arr[1] for arr in frame["V"])),
-            "gradients": np.vstack(tuple(arr[1] for arr in frame["F"])),
+            "velocities": frame["V"],
+            "gradients": frame["F"],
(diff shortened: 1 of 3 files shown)


MBartkowiakSTFC:
One observation so far is that this implementation of the converter is more strict than the original MDANSE CASTEP converter. The conversion will fail if the header is present, but the empty line after the `END header` is missing.

```
2026-05-06 10:23:38,369 - CRITICAL - process[24450] - IJob 541 - Job failed with traceback: Traceback (most recent call last):
IJob._runner[mode](self, prog=prog_bar)
File "MDANSE/Src/MDANSE/Framework/Jobs/IJob.py", line 389, in _run_singlecore
^^^^^^^^^^^^^^^^^^^^
File "MDANSE/Src/MDANSE/Framework/Converters/CASTEP.py", line 111, in run_step
frame = next(self.frames)
^^^^^^^^^^^^^^^^^
File "MDANSE/Src/MDANSE/Framework/Parsers/CASTEP_MD.py", line 58, in frames
yield from map(self.parse_frame, parser)
File "MDANSE/Src/MDANSE/Framework/Parsers/CASTEP_MD.py", line 63, in parse_frame
accum["time"] = frame["time"] * AUT
~~~~~~~~~~~~~~^~~~~
TypeError: can't multiply sequence by non-int of type 'float'
```

oerc0122:
I have a PR in-place in `castep_outputs` (oerc0122/castep_outputs#264) to fix that, but yes for now it strictly follows the definition.

MBartkowiakSTFC:
I installed `castep_outputs==0.3.3`, but on my machine the conversion still fails with the same error posted above, both for input files without the header and those that have the header but don't have an empty line afterwards.

oerc0122:
This is now blocked by oerc0122/castep_outputs#270

PR #1229, opened 2026-05-05
Modernise pyprojects to use dependency-groups and uv workspaces

oerc0122:
**Description of work**
Modernise pyproject.toml to exploit new (Oct. 2024; pip 25.1) features.

**Fixes**
For `pip install`:
- this should make no difference (besides PyPI not being able to also install lint/test) and changing to `pip install --group dev` to get lint/test groups.

For `uv`:
- This will automatically sync the `dev` group installing those packages.
- This will install both MDANSE/MDANSE_GUI (with deps) as editable in a shared venv which will be automatically updated. (`uv sync --all-packages`)

**To test**
Install with pip/uv

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install MDANSE
         working-directory: MDANSE
-        run: python -m pip install ".[test]"
+        run: python -m pip install --group test .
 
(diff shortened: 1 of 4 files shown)


MBartkowiakSTFC (approved):
I tried out the `pip` installation options (with and without the dev group). It worked as expected.

Issue #1196, opened 2026-03-10
[ENHANCEMENT] Diffraction pattern calculations (single crystal)


MBartkowiakSTFC:
**Is your feature request related to a problem? Please describe.**
Some of our users would like to calculate single crystal diffraction patterns from their MD trajectories. They are mainly interested in seeing the diffuse scattering in the results.

**Describe the solution you'd like**
1. Calculate the diffraction pattern from individual frames and then average over time.
2. Symmetry considerations and folding can be added once the basic calculation has been implemented.

**Describe alternatives you've considered**
Apparently, tools like https://github.com/jkulda/MP_tools exist for calculating diffuse scattering.

**Additional context**
This analysis should support both neutron and x-ray scattering.

glezm1971-cloud:
If you are interested in this, I think that this could be a source worthy to check: https://github.com/tproffen/DiffuseCode

There is also a book (https://doi.org/10.1093/acprof:oso/9780199233694.001.0001), but hopefully with the code and the manuals (which are quite complete) you won't need it.

MBartkowiakSTFC:
Our users agree that the calculation itself has been implemented in #1233 

They do recommend follow-up work on improving the visualisation of the results. We will have a separate issue for that.

PR #1233, opened 2026-05-19
3D static structure factor

ChiCheng45:
**Description of work**
An implementation of the 3D static structure factor. Additionally, it should reduce memory usage for multiprocess jobs if the output data is large, which can occur for this job on Windows.

Tasks remaining
 - [x] Fix when a large trajectory is used. When a large number of atoms are used, memory usage is very high. 
 - [x] Allow the options to change q-vector basis and etc.
 - [x] Change the current implementation so that it uses FFT to calculate the result? Perhaps use something like https://github.com/flatironinstitute/finufft?
 - [x] Check performance and memory usage. Fix if excessive.
 - [x] Update documentation.
 
 **To test**
Check that the GUI greys out static structure factor 3d when finufft is not installed. After installation of finufft, check that the 3D static structure factor works with a few system and q-vector options.


ChiCheng45:
Here is the result for SrTiO3. Scattering lengths are not implemented yet, and I will need to adjust and fix the grid spacing so it depends on the unit cell.

image

image

image

ChiCheng45:
I've updated so that the grid depends on the reciprocal lattice. I've also added partial terms to the results. Here is the results for the SrTiO3.

Total 
image

TiTi
image

Note the because of the choice we made in #963, the each point above averages the results over $-q$ and $q$.

MBartkowiakSTFC:
> I've updated so that the grid depends on the reciprocal lattice. I've also added partial terms to the results. Here is the results for the SrTiO3.
> 

This is already looking realistic.

Probably the next thing that would be good to have is reciprocal lattice units instead of 1/nm in the plot axes. Since this is tetragonal I4/mcm (but with c dimension on the first axis), peaks like 010 or 001 should not be visible, but 011, 020 and 002 should be there. I think that's what we already see in the plot above, but it would be good for the axis labels to reflect this.

ChiCheng45:
@MBartkowiakSTFC what do you reckon about adding [finufft](https://github.com/flatironinstitute/finufft) as a depedency? Seems to be a good use case for this problem, and it's massively faster than the current method. For a 100,000 atom system with 300,000 or more q-vectors its like going from hours to minutes.

direct (c1ef616343cc1615a5cb9e0cd9d76b6cf6c73284)
```python
    def run_step(self, index):
        frame_idx = index // self.q_vec_n_chunks
        chunk_idx = index % self.q_vec_n_chunks

        frame = self.configuration["frames"]["value"][frame_idx]
        coords = self.trajectory.configuration(frame)["coordinates"]
        start = chunk_idx * self.q_vec_chunk_size
        stop = min((chunk_idx + 1) * self.q_vec_chunk_size, self.n_q_vecs)

        q_vectors, q_idxs = q_vectors_in_cube(self.inverse, self.max_hkl, start, stop)

        rho = {}
        for element, idxs in self._indicesPerElement.items():
            selectedCoordinates = np.take(coords, idxs, axis=0)
            rho[element] = np.sum(
                np.exp(1j * np.dot(selectedCoordinates, q_vectors)), axis=0
            )

        return index, (q_idxs, rho)

    def combine(self, index, results):
        q_idxs, rho = results
        for pair_str, (label_i, label_j) in self.labels:
            np.add.at(
                self._outputData[f"ssf3d/{pair_str}"],
                q_idxs,
                (rho[label_i] * rho[label_j].conj()).real,
            )
```


non uniform fft
```python
    def run_step(self, index):
        frame = self.configuration["frames"]["value"][index]
        coords = self.trajectory.configuration(frame)["coordinates"]

        rho = {}
        for element, idxs in self._indicesPerElement.items():
            pts = 2 * np.pi * coords[idxs] @ self.inverse
            rho[element] = finufft.nufft3d1(
                pts[:, 0],
                pts[:, 1],
                pts[:, 2],
                np.ones(len(pts), dtype=np.complex128),
                n_modes=(
                    2 * self.max_hkl[0] + 1,
                    2 * self.max_hkl[1] + 1,
                    2 * self.max_hkl[2] + 1
                ),
                eps=1e-3,
            )

        return index, rho

    def combine(self, index, rho):
        for pair_str, (label_i, label_j) in self.labels:
            self._outputData[f"ssf3d/{pair_str}"] += (
                rho[label_i] * rho[label_j].conj()
            ).real
```

MBartkowiakSTFC:
> @MBartkowiakSTFC what do you reckon about adding [finufft](https://github.com/flatironinstitute/finufft) as a depedency? Seems to be a good use case for this problem, and it's massively faster than the current method. For a 100,000 atom system with 300,000 or more q-vectors its like going from hours to minutes.
> 

Since other packages for the same purpose (e.g. MP_tools mentioned in #1196) use finufft, we were going to look at it sooner or later.

It's a <3MB download, we might as well use it.

oerc0122:
As we chatted about, if this becomes an optional dependency until we add more use-cases I'd be happier, but I'm also not opposed to just using it directly. It seems well resourced, well used and likely to stay alive in the long term, so yeah.

MBartkowiakSTFC:
Screenshot 2026-07-21 at 14 03 39
I tried it out on a water trajectory as well. It produced rings, as expected.

MBartkowiakSTFC (approved):
The output prediction and vector cross-product input both work well. I'm happy with the results and the performance, too.

PR #1235, opened 2026-05-20
Load atoms based on HDF5 chunks in DISF

MBartkowiakSTFC:
**Description of work**
As an attempt to improve the efficiency of MDANSE calculations, this PR introduces a prototype function for grouping atom indices into sets, based on the chunking of the underlying HDF5 dataset in the trajectory file.

The grouping scheme is meant to achieve the following goals:
1. Include multiple atoms in a single calculation "step",
2. Only use atoms from a single chunk in one step,
3. Always have at least as many steps as the number of CPU cores,
4. Limit the number of atoms in a chunk not to exceed a memory limit.

If this approach turns out to be useful, some of the parameters (e.g. memory limit per process) will have to be added to the analysis inputs.

**Fixes**
1. Dynamic Incoherent Structure Factor calculation has been rewritten to load many atoms in a single step,
2. `atomic_trajectory` has been copied and extended into `atomic_trajectory_many`,
3. New functions grouping atom indices have been added to `TrajectoryUtils`,
4. Keyword arguments have been added to trajectory classes to adjust the dataset cache size.

**To test**
Run a DISF calculation for a trajectory using the protos branch and then this branch. Check that the results are still the same. Compare the execution time.


MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
```diff
@@ -59,3 +66,157 @@ def atomic_trajectory(
...
+
+    """
+    # i is time, j is the XYZ coordinate
+    trajectory = np.einsum("inj,ijk->ink", config, rcell)
+    sdxyz = trajectory[1:, :] - trajectory[:-1, :]
+    sdxyz -= np.cumsum(np.round(sdxyz), axis=0)
+    trajectory[1:, :] = trajectory[:-1, :] + sdxyz
+    if not box_coordinates:
+        trajectory = np.einsum("inj,ijk->ink", trajectory, cell)
+    return trajectory
```
ChiCheng45, line 106: Looks good, but I've been wondering if we calculate DISF with the sphericallatticeqvectors then do we need to run atomic_trajectory or atomic_trajectory_many? I think not because 

```math
\exp(-iq \cdot [r_{i}(0) - r_i(t) + n(t) ]) = \exp(-iq \cdot [r_{i}(0) - r_i(t) ])
```

where $r_i(t)$ is the coordinate folded into the unit cell. If I remember correctly the atomic_trajectory was the slower part of the DISF calculation so we might be able to get more of a speed up if the user uses sphericallatticeqvectors. Of course, if they use sphericalqvectors then we will need to do atomic_trajectory.

I think this also might be why MDMC S(q,w) calculation was faster because it was not doing the atomic_trajectory calculation.

MBartkowiakSTFC: OK, I can add a check for the vector generator type, and skip the atomic_trajectory if it used a lattice vector generator.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -224,36 +234,54 @@ def run_step(self, index):
...
-                n_configs * np.sum(qvec_weights)
+            rho = np.exp(
+                1j * np.einsum("inj,jk,k->ikn", series, qVectors, np.sqrt(qvec_weights))
             )
+            res = np.hstack(
+                [
+                    correlate(rho[:, :, n], rho[:n_configs, :, n], mode="valid")
+                    for n in range(rho.shape[2])
+                ]
+            )
```
oerc0122: Is this a function which we're likely to use in other places? Might be worth breaking out into its own function.

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
```diff
@@ -59,3 +66,157 @@ def atomic_trajectory(
...
+        Number of available CPU cores or worker processes, by default 1
+    memory_scale_factor : int, optional
+        Multiplier of the expected memory requirements of the job, by default 10
+    size_limit : int, optional
+        Limit of memory per worker process, by default CHUNK_SIZE_LIMIT
+
+    Returns
+    -------
+    list[list[int]]
+        _description_
```
oerc0122: Needs description?

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
```diff
@@ -59,3 +66,157 @@ def atomic_trajectory(
...
+    """
+    selected_indices = sorted(traj_instance.atom_indices)
```
oerc0122: I think it would make sense for `selected_indices` to just be an unordered `set` here, it's only used in a `set.intersection` later, and since an `intersection` is commutative this might even be simpler.

oerc0122:
```suggestion
    selected_indices = set(traj_instance.atom_indices)
```

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
```diff
@@ -59,3 +66,157 @@ def atomic_trajectory(
...
+        )
+    min_chunk_count = n_proc
+    initial_sets = [
+        sorted(
+            set(range(chunk_limits[n], chunk_limits[n + 1])).intersection(
+                selected_indices
+            )
+        )
+        for n in range(len(chunk_limits) - 1)
+    ]
```
oerc0122, line 165:
```python
from itertools import pairwise
...
    initial_sets = [
        sorted(
            selected_indices.intersection(range(*chunk_block))
        )
        for chunk_block in pairwise(chunk_limits)
    ]
```

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
```diff
@@ -59,3 +66,157 @@ def atomic_trajectory(
...
+        List of index sets after splitting into smaller sets.
+    """
+    result = []
+    for ind_set in index_sets:
+        set_len = len(ind_set)
+        if set_len < max_size:
+            result.append(ind_set)
+            continue
+        result.append(ind_set[: set_len // 2])
+        result.append(ind_set[set_len // 2 :])
```
oerc0122: May be slightly off, but this looks like:

```python
from more_itertools import chunked_even
...
    result = [chunk 
              for ind_set in index_sets 
              for chunk in chunked_even(ind_set, max_size)]
```

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
```diff
@@ -59,3 +66,157 @@ def atomic_trajectory(
...
+        sorted(
+            set(range(chunk_limits[n], chunk_limits[n + 1])).intersection(
+                selected_indices
+            )
+        )
+        for n in range(len(chunk_limits) - 1)
+    ]
+    return balance_index_groups(
+        initial_sets, max_size=max_chunk_size, min_count=min_chunk_count
+    )
```
oerc0122, line 169: I'm wondering whether most of these operations would be better as generators rather than lists.

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -312,6 +325,27 @@ def charges(
 
         return charge.astype(np.float64)
 
+    def chunk_size(self, dataset_type: TrajDataArray = TrajDataArray.POSITION) -> int:
+        if dataset_type == TrajDataArray.POSITION:
+            data_key = self.KEYS["position"]
+        elif dataset_type == TrajDataArray.VELOCITY:
+            data_key = self.KEYS["velocity"]
+        elif dataset_type == TrajDataArray.FORCE:
+            data_key = self.KEYS["force"]
```
```suggestion
        data_key = self.KEYS[dataset_type.name.lower()]
```

MDANSE/Src/MDANSE/Trajectory/H5MDTrajectory.py
```diff
@@ -312,6 +325,27 @@ def charges(
...
+        try:
+            dataset = self._h5_file[data_key]
+        except KeyError:
+            LOG.error("Dataset %s was not in the trajectory file", data_key)
+            return -1
+        if not hasattr(dataset, "chunks") or dataset.chunks is None:
+            LOG.warning("Dataset %s is not chunked, and was expected to be", data_key)
+            return -1
+        chunk_shape = dataset.chunks
```
```suggestion
        if (chunk_shape := getattr(dataset, "chunks", None)) is None:
            LOG.warning("Dataset %s is not chunked, and was expected to be", data_key)
            return -1
```

MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
```diff
@@ -221,6 +239,27 @@ def charges(
         n_req = len(range(*indices.indices(self.chemical_system.number_of_atoms)))
         return np.zeros(n_req, dtype=np.float64)
 
+    def chunk_size(self, dataset_type: TrajDataArray = TrajDataArray.POSITION) -> int:
+        if dataset_type == TrajDataArray.POSITION:
+            data_key = "/configuration/coordinates"
+        elif dataset_type == TrajDataArray.VELOCITY:
+            data_key = "/configuration/velocities"
+        elif dataset_type == TrajDataArray.FORCE:
+            data_key = "/configuration/gradients"
```
oerc0122:
```suggestion
        data_key = self.KEYS[dataset_type.name.lower()]
```

MDANSE/Src/MDANSE/Trajectory/MdanseTrajectory.py
```diff
@@ -221,6 +239,27 @@ def charges(
...
+        try:
+            dataset = self._h5_file[data_key]
+        except KeyError:
+            LOG.error("Dataset %s was not in the trajectory file", data_key)
+            return -1
+        if not hasattr(dataset, "chunks"):
+            LOG.warning("Dataset %s is not chunked, and was expected to be", data_key)
+            return -1
+        chunk_shape = dataset.chunks
```
oerc0122: Also as above

ChiCheng45 (commented):
Here is the timing for a 2048 atom, 10001 frame argon trajectory with

```
- q_vectors = ('SphericalLatticeQVectors', {'seed': 1, 'shells': [10.0, 10.0, 1.0], 'n_samples': 500000, 'n_vectors': 100, 'force_equal_weights': False, 'width': 1.0})
- running_mode = ('multicore', 8)
``` 

Protos: 6m 49s
This branch b280aa1a97c0576a55e57ec19680494b549e9d84:  3m 6s
This branch 491cd7f9b2c2d1c21a4bb718ac5e17f095ba8e8c (don't use atomic trajectory):  3m 21s

So apparently, my suggestion didn't make much difference for this system. Sorry about that.

I had to set the seed in `def spherical_vectors(q: float, q_width: float, n_vecs: int)` because that part of the code's seed is not set by MDANSE.

Anyway, for some reason, I get some slight differences between the three results. Could be numerical noise.

Protos: test1
This branch b280aa1a97c0576a55e57ec19680494b549e9d84:  test3
This branch 491cd7f9b2c2d1c21a4bb718ac5e17f095ba8e8c (don't use atomic trajectory):  test2
Image

All vectors and weights are the same.

ChiCheng45 (commented):
I also get an error when I select one atom.

```
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 544, in run
    IJob._runner[mode](self, prog=prog_bar)
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 392, in _run_singlecore
                  ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\DynamicIncoherentStructureFactor.py", line 275, in run_step
    res = np.hstack(
          ^^^^^^^^^^
  File "C:\Users\...\AppData\Local\miniconda3\envs\MDANSE3\Lib\site-packages\numpy\_core\shape_base.py", line 365, in hstack
    return _nx.concatenate(arrs, 1, dtype=dtype, casting=casting)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: need at least one array to concatenate
```

MBartkowiakSTFC:
> I also get an error when I select one atom.
> 

Thanks for spotting this one! I will fix it.
I thought that the "one atom" unit tests would have picked this up.

MBartkowiakSTFC:
I tried running the MDMC argon-openmm-mdanse.py script with different MDANSE branches. For 10 refinement steps in the script I got the following timing
protos: 19 minutes
this branch: 22 minutes

At the same time, an equivalent script using the MDMC S(q,w) instead of MDANSE finished in 4 minutes.

I will do more careful checks and some profiling to see where it goes wrong. I suspect that my trajectory was not long enough and the overheads of starting a subprocess on a Mac dominated my results.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -224,36 +234,54 @@ def run_step(self, index):
...
                 continue
             qVectors = self.configuration["q_vectors"]["value"][q]["q_vectors"]
             qvec_weights = self.configuration["q_vectors"]["value"][q]["weights"]
-
-            rho = np.exp(1j * np.dot(series, qVectors)) * np.sqrt(qvec_weights[None, :])
-            res = correlate(rho, rho[:n_configs], mode="valid").T[0] / (
-                n_configs * np.sum(qvec_weights)
+            rho = np.exp(
+                1j * np.einsum("inj,jk,k->ikn", series, qVectors, np.sqrt(qvec_weights))
             )
```
ChiCheng45: The weights are inside the np.exp function should be outside.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
ChiCheng45: I did some line profiling, and apparently, this part of the code is quite slow. Apparently, it can be with higher-dimensional arrays. I did some playing around and found that

```python
            fast_len = scipy.fft.next_fast_len(rho.shape[0])
            v = scipy.fft.fft(rho, n=fast_len, axis=0)
            w = scipy.fft.fft(rho[:n_configs], n=fast_len, axis=0)
            res = scipy.fft.ifft(np.sum(v * w.conj(), axis=1), axis=0)[:self._nFrames]
```

was faster. Interestingly, this is like MDANSE 1.5 again, but when I was testing the original MDANSE 1.5 correlation function, it was slower than scipy.signal.correlate. Not sure why, but the above is faster than both, I think, because here I sum before the IFFT, the 1.5 function does the sum after the IFFT. Also, because we do the windowing thing, we only need to pad zeros on the smaller windowed array. Also, apparently, scipy FFT is supposed to be better.

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
MBartkowiakSTFC: Generators would work here, true. At the same time, I feel inclined to keep the current version because this code is only executed once before the start of the actual calculation, and the effect on the overall performance will not be significant.

MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
MBartkowiakSTFC: I made this change. The code uses `pairwise` now.

ChiCheng45 (approved):
Looks good to me.I get a good speed up from 1m 4s to 20s on the SrTiO3 trajectory.

Results are the same as before (needed to merge protos on to this branch so the seed got fix properly).

Image

PR #1238, opened 2026-05-27
Fix for the sphericalqvector with a set seed and update mdanse to use new numpy approach to random number generation

### LINKED ISSUE #1237: [BUG] Setting seed with sphericalqvectors does not fix q-vector with respect to rerun jobs
**Description of the error**
When the seed is set to 1 in the job's parameters using sphericalqvectors the q-vectors that are used are not fixed with respect to job reruns.

**Additional details**
Here are two reruns with the same job parameters with the seed set to 1.

Image

Image

ChiCheng45:
**Description of work**
Updated `def spherical_vectors(q: float, q_width: float, n_vecs: int)` so that it uses the random number generator, which should have its seed set by MDANSE and not randomly.

Updated MDANSE so that the new approach to random number generation is used.

**Fixes**
Fixes #1237

**To test**
Run a few of the same jobs with the sphericalqvector generator with a seed set to something other than zero, but the same. Check that the q-vectors in the vector plots are the same. Do this with sphericallatticeqvector generator, check that the weights are the same. 

Check setting the seed with the other q-vector generator circularlatticeqvector, circularqvector, linearlatticeqvector, and the linearqvector continue to work with the seed set and also continue to be random with the seed set to zero.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/QVectors/CircularLatticeQVectors.py
+++ b/MDANSE/Src/MDANSE/Framework/QVectors/CircularLatticeQVectors.py
@@ -63,10 +63,7 @@ class CircularLatticeQVectors(LatticeQVectors):
     )
 
     def _generate(self):
-        if self._configuration["seed"]["value"] != 0:
-            rng = np.random.default_rng(self._configuration["seed"]["value"])
-        else:
-            rng = np.random.default_rng()
+        rng = np.random.default_rng(self._configuration["seed"]["value"] or None)
 
         nvecs_per_shell = self._configuration["n_vectors"]["value"]
         n_samples = self._configuration["n_samples"]["value"]
--- a/MDANSE/Src/MDANSE/Framework/QVectors/CircularLatticeQVectors.py
+++ b/MDANSE/Src/MDANSE/Framework/QVectors/CircularLatticeQVectors.py
@@ -15,7 +15,6 @@
 #
 from __future__ import annotations
 
-import random
 from functools import partial
 
 import numpy as np
@@ -65,8 +64,9 @@ class CircularLatticeQVectors(LatticeQVectors):
 
     def _generate(self):
         if self._configuration["seed"]["value"] != 0:
-            np.random.seed(self._configuration["seed"]["value"])
-            random.seed(self._configuration["seed"]["value"])
+            rng = np.random.default_rng(self._configuration["seed"]["value"])
+        else:
+            rng = np.random.default_rng()
 
         nvecs_per_shell = self._configuration["n_vectors"]["value"]
         n_samples = self._configuration["n_samples"]["value"]
@@ -83,7 +83,7 @@ class CircularLatticeQVectors(LatticeQVectors):
         self._configuration["q_vectors"] = {}
 
         for q in self._configuration["shells"]["value"]:
-            samples = circle_of_vectors(q, width, n_samples, rot_mat=rot_mat)
+            samples = circle_of_vectors(q, width, n_samples, rot_mat=rot_mat, rng=rng)
             lattice_hkl_vectors, _ = self.lattice_vectors_with_weights(
                 samples,
                 self._unit_cell,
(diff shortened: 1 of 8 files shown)


MBartkowiakSTFC:
Just to check with you: this fix means that we're going back to the legacy RNG in numpy:
https://numpy.org/doc/stable/reference/random/legacy.html
It probably won't cause any problems for another few years, but let's just think for 5-10 minutes if there is a way to use a fixed seed using what is apparently numpy's preferred random number generation mechanism.
In principle there is a keyword argument for setting a seed here:
https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng
Would this be an option?

oerc0122:
I think we either want to attach an RNG base object to the `IQVectors` and make the generation functions methods, or make RNG a passable arg into the generation functions.

ChiCheng45:
> Just to check with you: this fix means that we're going back to the legacy RNG in numpy: https://numpy.org/doc/stable/reference/random/legacy.html It probably won't cause any problems for another few years, but let's just think for 5-10 minutes if there is a way to use a fixed seed using what is apparently numpy's preferred random number generation mechanism. In principle there is a keyword argument for setting a seed here: https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng Would this be an option?

I see. I wasn't aware of the new approach. Anyway, I think there are a few places in MDANSE that still use the legacy approach e,g np.random.uniform, etc so will need to fix them as well.

MBartkowiakSTFC:
> 
> I see. I wasn't aware of the new approach. Anyway, I think there are a few places in MDANSE that still use the legacy approach e,g np.random.uniform, etc so will need to fix them as well.

Should we just merge your changes here and then create an issue for moving away from the legacy random number generation altogether?

ChiCheng45:
> > I see. I wasn't aware of the new approach. Anyway, I think there are a few places in MDANSE that still use the legacy approach e,g np.random.uniform, etc so will need to fix them as well.
> 
> Should we just merge your changes here and then create an issue for moving away from the legacy random number generation altogether?

I'm updating this PR it to use the new approach as it shouldn't be too many changes.

MBartkowiakSTFC (approved):
I tested it for linear, circular and spherical lattice vectors. I got the same results for the same random seed in consecutive runs. Looks good.

MDANSE/Src/MDANSE/Framework/QVectors/CircularLatticeQVectors.py
```diff
@@ -65,8 +64,9 @@ class CircularLatticeQVectors(LatticeQVectors):
 
     def _generate(self):
         if self._configuration["seed"]["value"] != 0:
-            np.random.seed(self._configuration["seed"]["value"])
-            random.seed(self._configuration["seed"]["value"])
+            rng = np.random.default_rng(self._configuration["seed"]["value"])
+        else:
+            rng = np.random.default_rng()
```
oerc0122:
```suggestion
        rng = np.random.default_rng(self._configuration["seed"]["value"] or None)
```

MDANSE/Src/MDANSE/Framework/QVectors/CircularQVectors.py
```diff
@@ -118,7 +121,9 @@ class CircularQVectors(IQVectors):
 
     def _generate(self):
         if self._configuration["seed"]["value"] != 0:
-            np.random.seed(self._configuration["seed"]["value"])
+            rng = np.random.default_rng(self._configuration["seed"]["value"])
+        else:
+            rng = np.random.default_rng()
```
oerc0122: As above and elsewhere.

MDANSE/Src/MDANSE/Framework/QVectors/LatticeQVectors.py
```diff
@@ -56,7 +59,7 @@ def fpsampling(
     if n_vecs >= n_points:
         return np.arange(n_points)
     elif n_points == 1:
-        return np.array([np.random.randint(0, n_points)])
+        return np.array([rng.integers(low=0, high=n_points)])
```
oerc0122: Is this really a special case? 
```suggestion
        return rng.integers(low=0, high=n_points, size=1)
```

MDANSE/Src/MDANSE/Framework/QVectors/LatticeQVectors.py
oerc0122: Equality comparison with 0 is still risky.

oerc0122 (approved):
Looks good, only other suggestion would be to typehint `rng` in functions as you have for `SphericalQVectors` elsewhere, but not essential.

PR #1240, opened 2026-06-08
HDF5 parameters for chunking and cache size

MBartkowiakSTFC:
**Description of work**
As mentioned in #1162, on certain platforms (especially using cloud storage) it may be necessary to use different chunking scheme, HDF5 drivers or HDF5 cache size. This PR is meant to let the users control the HDF5 parameters so we can try out different parameter combinations easily.

**Fixes**
1. Added "frames per chunk" to the "atoms per chunk" in the trajectory output.
2. Added HDF5 cache parameters to the trajectory input widget.
3. Suppressed `ChemicalSystem` creation in subprocesses.
4. Removed `ChemicalSystem` from `Configuration` classes.

**To test**
All tests must pass.
Manual testing of CLI and GUI is recommended.

PR #1245, opened 2026-06-16
Chunk-aware analysis

MBartkowiakSTFC:
**Description of work**
The main goal of this PR is to speed up MDANSE analysis runs.

This work is based on the changes from #1240 

**Fixes**
So far, the following changes have been implemented:

1. Grouping atom indices into batches belonging to the same chunk in the underlying trajectory file. This reduces the number of read operations performed on the file.
2. Use [scipy.interpolate.make_interp_spline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.make_interp_spline.html#make-interp-spline) to calculate the velocity by interpolating the atom positions. **This change has been removed in order not to reduce the precision in the case of order=2 and all the frames being present in the trajectory.**
3. All the analysis types which used `read_atomic_trajectory` have been modified to use `read_atomic_trajectory_many`.

**To test**
All analysis types must be checked, given the extent of the changes. Ideally we should be comparing both the results and the performance before and after the changes.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/CurrentCorrelationFunction.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/CurrentCorrelationFunction.py
@@ -33,7 +33,6 @@ from MDANSE.Mathematics.Signal import (
     get_spectrum,
 )
 from MDANSE.MLogging import LOG
-from MDANSE.MolecularDynamics.Analysis import mean_square_displacement_many
 from MDANSE.MolecularDynamics.TrajectoryUtils import group_atom_indices
 from MDANSE.MolecularDynamics.UnitCell import UnitCell
 from MDANSE.util_types import ComplexArray
--- a/MDANSE/Src/MDANSE/Framework/Jobs/CurrentCorrelationFunction.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/CurrentCorrelationFunction.py
@@ -29,10 +29,12 @@ from MDANSE.Framework.Jobs.IJob import IJob
 from MDANSE.Framework.QVectors.IQVectors import IQVectors
 from MDANSE.Mathematics.Arithmetic import assign_weights, get_weights, weighted_sum
 from MDANSE.Mathematics.Signal import (
-    differentiate,
+    differentiate_many,
     get_spectrum,
 )
 from MDANSE.MLogging import LOG
+from MDANSE.MolecularDynamics.Analysis import mean_square_displacement_many
+from MDANSE.MolecularDynamics.TrajectoryUtils import group_atom_indices
 from MDANSE.MolecularDynamics.UnitCell import UnitCell
 from MDANSE.util_types import ComplexArray
 
@@ -127,6 +129,8 @@ class CurrentCorrelationFunction(IJob):
     def initialize(self):
         """Initialize the input parameters and analysis self variables."""
         super().initialize()
+
+        self.n_proc = self.configuration["running_mode"].get("slots", 1)
 
         self.numberOfSteps = self.configuration["q_vectors"]["n_shells"]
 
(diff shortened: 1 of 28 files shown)


MDANSE/Src/MDANSE/Mathematics/Signal.py
```diff
@@ -203,6 +179,17 @@ def differentiate(a, dt=1.0, order=1):
...
 
+def differentiate_many(input_array, dt: float = 1.0, order: int = 1):
+    time_axis = np.arange(len(input_array)) * dt
+    if order == 1:
+        spline = make_interp_spline(time_axis, input_array, k=order, axis=0)
+    else:
+        spline = make_interp_spline(
+            time_axis, input_array, k=order, axis=0, bc_type=None
+        )
+    return spline(time_axis, nu=1)
```
ChiCheng45: Looks fine, results are close to protos, and it's much faster. 

Just noting again that when calculating velocities with spline, will mean that order=2 will no longer reproduce velocities from the MD when something like the Velocity Verlet is used (assuming all frames were dumped by the user).

Image

Image

I don't mind this too much since it's still close, and the user should dump velocities in most cases.

MBartkowiakSTFC: Thanks for pointing this out. The spline approach was faster and more accurate in my tests on a trajectory where some frames have been skipped. I'll think some more about the best way to proceed, now that you pointed out that we lose accuracy for trajectories where all the frames have been written out.

MBartkowiakSTFC: In the end I decided to go back to the original velocity calculation, not to change the results. But it would be good to add splines as an additional option in the long run.

In any case, it was possible to speed up the `differentiate` function and now `differentiate_many` using the old equation is about as fast as the splines.

MDANSE/Src/MDANSE/MolecularDynamics/Analysis.py
```diff
@@ -54,7 +68,36 @@ def mean_square_displacement(coords: FloatArray, n_configs: int) -> FloatArray:
     return msd
```
ChiCheng45: Should we just remove this now? Everything else uses mean_square_displacement_many, except for one test.

MDANSE/Src/MDANSE/Framework/Jobs/CurrentCorrelationFunction.py
```diff
@@ -29,10 +29,12 @@
 from MDANSE.Framework.QVectors.IQVectors import IQVectors
 from MDANSE.Mathematics.Arithmetic import assign_weights, get_weights, weighted_sum
 from MDANSE.Mathematics.Signal import (
-    differentiate,
+    differentiate_many,
     get_spectrum,
 )
 from MDANSE.MLogging import LOG
+from MDANSE.MolecularDynamics.Analysis import mean_square_displacement_many
```
ChiCheng45: Unused import can be removed.

MDANSE/Src/MDANSE/Trajectory/FileTrajBase.py
```diff
@@ -354,38 +356,38 @@ def read_atomic_trajectory(
             coords,
             direct_cells,
             inverse_cells,
-            box_coordinates=box_coordinates,
+            fractional_coordinates=fractional_coordinates,
         )
```
ChiCheng45: I think there is one place that uses read_atomic_trajectory, which is in i_reader.py. Can we make it use read_atomic_trajectory_many but with one index? That way we can get rid of read_atomic_trajectory and atomic_trajectory and make the _many versions of the function the standard way to do things.

Don't mind leaving it in as well.

ChiCheng45:
Results look good. Here are some timings I did using an Argon trajectory I had, looks great.


| Job  | This branch | Protos |
| ------------- | ------------- | ------------- |
| Average structure | 24s  | 15m 57s |
| PCF | 28s  | 16m 22s  |
| VCF | 23s  | 15m 53s  |
| MSD | 26s  | 17m 18s  |
| EISF | 2m 39s  | 18m 20s  |
| RMSD | 22s | 16m 56s |
| RMSF | 20s  | 16m 50s  |
| Temperature | 14s  | 15m 55s  |
| CCF | 38s  | 4m 28s  |
| VHFS | 12s | 5m 40s  |

ChiCheng45 (approved):

MBartkowiakSTFC:
> Results look good. Here are some timings I did using an Argon trajectory I had.

Thanks for timing it. For a 300 MB / 5000 frames water trajectory containing velocities, my CCF was running for 4 minutes (compared to 11 minutes 30 on protos branch). Which is still a good speedup.

PR #1247, opened 2026-06-23
Unit test fixes

MBartkowiakSTFC:
**Description of work**
Unit tests were all passing with scipy 1.17.1, but now one test fails with Python>=3.12 and scipy==1.18.0. This should be investigated, but for now let's use the last version that did not trigger errors.

**Fixes**
1. Limits scipy version to 1.17.1.
2. Makes all platforms use "spawn" method of creating processes. (Hoping to fix random timeout on Linux.)

**To test**
All tests must pass.


MBartkowiakSTFC:
While it is possible that using `spawn` could fix the problems with tests timing out, there are visible performance losses on Linux when not using `fork`. The new plan is to switch only the tests to `spawn` and wait to see if the timeout problem does not reoccur.

PR #1248, opened 2026-06-24
Rewrite the bug report template

MBartkowiakSTFC:
**Description of work**
The bug report template we are using could be improved. This PR proposes a modified structure. Comments welcome.

**Fixes**
1. Changed the template for a GitHub bug report.

**To test**
Please check if the new template makes sense and prompts the users for all the information we need.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,23 +1,29 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Create a bug report to help us improve MDANSE
 title: "[BUG]"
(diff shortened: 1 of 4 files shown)


oerc0122 (commented):
One small thing I would suggest would be to add instructions of how to get this information for the average user. Can/should we expose Python version and CPU info in the GUI?

MBartkowiakSTFC:
> One small thing I would suggest would be to add instructions of how to get this information for the average user. Can/should we expose Python version and CPU info in the GUI?

Good point. I can add more details to the version information for both CLI and GUI.

oerc0122 (approved):
Looks good to me.

PR #1250, opened 2026-06-25
Change velocity unit of trj files to ang/ps

### LINKED ISSUE #1249: [BUG] Velocity unit in trj files (DFTB converter)
**Description of the error**
The temperature from velocities does not match the temperature from atom coordinates for the DFTB water trajectory (or Forcite nylon trajectory).

The temperature calculated by interpolating atom coordinates should be close to the formally correct value calculated from velocities.

This is the temperature of the DFTB water based on the atom coordinates as a function of time.
Image

For the DFTB water, the velocities are three orders of magnitude larger than expected based on the position changes between frames.

This is the temperature calculated from atom velocities at each frame:

Image

**Suggested fix**
We need to check carefully what units are used in the trj file. Changing the velocity unit in the parser from "Ang/fs| to "Ang/ps" improved the temperature results, but there was still roughly a factor 2 between the velocities and interpolated atom positions.

Image

**Additional details**
N/A

MBartkowiakSTFC:
**Description of work**
Assumes that velocity in trj files (e.g. from Forcite) is given in units of Ang/ps. This seems to work file for the trajectories of water and nylon from MDANSE-Examples.

Closes #1249 

**Fixes**
1. Changed "ang/fs" to "ang/ps".

**To test**
Calculate the temperature of water or nylon from the example trajectories. It should be less than 500K.


oerc0122 (approved):
If this fixes the issue, looks good to me.

PR #1254, opened 2026-07-01
Make ase,mdtraj and MDAnalysis optional extras

oerc0122:
**Description of work**
Remove explicit dependency on optional external projects.

**Fixes**
- Remove dependency on ase, mdtraj and mdanalysis
- Disable jobs if not present.
- Add `visible` label to `Configurable` s to enable visibility without working. 
- These should also have a `requires_extras` tuple to display a useful error message to users for disabled jobs.

**To test**
Try GUI, ensure jobs are disabled. Ensure tests pass.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
@@ -239,6 +239,21 @@ class Action(QWidget):
                     f"traceback {traceback.format_exc()}"
                 )
                 return
+
+            if not job_instance.enabled:
+                widget = QLabel(f"""<h1>Job {job_name} not available.</h1>
+
+                <h2>Please install MDANSE with the following extras enabled: {", ".join(job_instance.requires_extras)}.</h2>
+
+                <h2>To do this, install using, e.g.:</h2>
+
+                <h2><b>pip install "mdanse[{",".join(job_instance.requires_extras)}]"</b></h2>
+                """)
+                self.layout.addWidget(widget)
+                self._widgets_in_layout["err"] = widget
+                self._widgets.append(widget)
+                return
+
             job_instance.build_configuration()
             settings = job_instance.settings
             self._job_instance = job_instance
--- a/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
@@ -241,9 +241,11 @@ class Action(QWidget):
                 return
 
             if not job_instance.enabled:
-                widget = QLabel(f"""<h1>Job {job_name} not available.</h1>
+                widget = QLabel(f"""<h1>{job_name} requires an external package.</h1>
 
-                <h2>Please install MDANSE with the following extras enabled: {", ".join(job_instance.requires_extras)}.</h2>
+                <h2>Please install MDANSE with the following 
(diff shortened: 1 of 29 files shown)


MBartkowiakSTFC:
Could we change the behaviour so the corresponding jobs are still listed in the GUI, but are not enabled if their underlying Python package is not installed? This would prevent the situation where a new user never realises that they had the option of installing external packages. What do you think?

oerc0122:
This is exactly what Chi said, so it seems like a popular option. Do we want to change `enabled` to grey out the option rather than not list in general?

MBartkowiakSTFC:
At the moment the `enabled` variable in MDANSE also controls if the job is visible. We probably need a distinction between `enabled` and `visible`.

On the GUI side, it should be possible to run `.setEnabled(False)` on the specific `QStandardItem` in the JobTree.

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
```diff
@@ -239,6 +239,21 @@ def update_panel(self, job_name: str) -> None:
                     f"traceback {traceback.format_exc()}"
                 )
                 return
+
+            if not job_instance.enabled:
+                widget = QLabel(f"""<h1>Job {job_name} not available.</h1>
```
MBartkowiakSTFC: @oerc0122 @ChiCheng45 What phrasing do we prefer for this message?

I would consider emphasising in the text that the whole thing is optional, so the users don't wonder why we did not include the required dependency to start with. Something like:
```
                widget = QLabel(f"""<h1>{job_name} requires an external package.</h1>

                <h2>The following optional dependencies of MDANSE are needed here: {", ".join(job_instance.requires_extras)}.</h2>

                <h2>If you want to use {job_name}, please install the additional dependencies, e.g. by running:</h2>

                <h2><b>pip install "mdanse[{",".join(job_instance.requires_extras)}]"</b></h2>
                """)
```

Anyway, I'm also happy with the current version (especially since it tells the user what to do to enable the job).

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Visualisers/Action.py
```diff
@@ -239,6 +239,21 @@ def update_panel(self, job_name: str) -> None:
...
+
+                <h2>Please install MDANSE with the following extras enabled: {", ".join(job_instance.requires_extras)}.</h2>
+
+                <h2>To do this, install using, e.g.:</h2>
+
+                <h2><b>pip install "mdanse[{",".join(job_instance.requires_extras)}]"</b></h2>
```
oerc0122: I would probably combine the two to make:

```suggestion
                widget = QLabel(f"""<h1>{job_name} requires an external package.</h1>

                <h2>Please install MDANSE with the following 
                optional dependencies enabled: 
                {", ".join(job_instance.requires_extras)}.</h2>

                <h2>To do this, install using, e.g.:</h2>

                <h2><b>pip install "mdanse[{",".join(job_instance.requires_extras)}]"</b></h2>
```

MBartkowiakSTFC: Just to document our discussion: we agreed to use this suggestion, and to add an installation option that installs all optional dependencies.

MBartkowiakSTFC (approved):
I tried the GUI out without mdtraj, mdanalysis or ase, and then again, after installing them with `[all]`. Everything worked as expected.

PR #1258, opened 2026-07-03
Extend plot creator

MBartkowiakSTFC:
**Description of work**
One of the weakest spots of the current plotting workflow in MDANSE_GUI is the one-way communication between "Plot Creator" and "Plot Holder". This PR is meant to give the users a chance to see that their actions in "Plot Creator" tab had some effect on "Plot Holder", without the need to switch between the two.

**Fixes**
1. Plot creator buttons have more accurate labels,
2. Plot creator contains text informing the user about the current plotting tab in "Plot Holder".
3. Plot creator contains a preview of the datasets in the current plot in "Plot Holder".

**To test**
Try to create some plots and text views out of .mda files. Check if you find that the information shown in "Plot Creator" helps you understand what is happening in the "Plot Holder" tab.

### DIFF
--- a/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
+++ b/MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
@@ -161,6 +161,18 @@ class TabbedWindow(QMainWindow):
         self._tabs["Instruments"]._visualiser.instrument_details_changed.connect(
             self._tabs["Actions"].update_action_after_instrument_change
         )
+        self._tabs["Plot Holder"]._visualiser.current_tab_index.connect(
+            self._tabs["Plot Creator"]._visualiser.new_target_plot_index
+        )
+        self._tabs["Plot Holder"]._visualiser.current_tab_count.connect(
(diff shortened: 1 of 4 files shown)


ChiCheng45 (commented):
Looks good; seems to do the job. I think it's fine to go in now, but here are a few suggestions from me and one small visual issue I noticed.

Can the appear in tab bit be the name of the tab instead? As sometimes I have closed some tabs and the numbers will be different, etc. Also might be worth removing the "new" in the plot or text view names.
Image
Image

Can we move some stuff around so that it becomes something like this? It's just in my mind that they should be next to each other.
Image

Small visual issue where the width of the different subsections changes when I click new empty plot or text view.

Animation47

MBartkowiakSTFC:
I think that all these changes can be achieved with a bit of effort. I'll let you know when it's ready to be tested again.

Also, I can probably remove the pop-up dialog informing the users that the plot has been created in the next tab. With these changes it should be pretty clear that the plot can be viewed in the next tab.

ChiCheng45 (approved):
Looks great.

PR #1260, opened 2026-07-06
Improve index grouping to prevent zero-size groups

### LINKED ISSUE #1257: [BUG] Issues running DISF with a large number of atoms and frames
**Background information**
Your OS: Windows 11 
Python version: 3.11.15
MDANSE version: 2.0.1.post13+git.0c6b336a
MDANSE_GUI version:  2.0.1.post13+git.0c6b336a
CPU architecture: AMD64

**What were you doing?**
I am trying to run DISF on a trajectory with 425000 frames and 14400 atoms. 

**What went wrong?**
The following errors were obtained.

```
2026-07-03 14:15:19,219 - CRITICAL - process[25088] - IJob 556 - Job failed with traceback: Traceback (most recent call last):
File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\IJob.py", line 534, in run
self.initialize()
File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\Framework\Jobs\DynamicIncoherentStructureFactor.py", line 119, in initialize
self.grouped_indices = group_atom_indices(
^^^^^^^^^^^^^^^^^^^
File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\MolecularDynamics\TrajectoryUtils.py", line 169, in group_atom_indices
return balance_index_groups(
^^^^^^^^^^^^^^^^^^^^^
File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\MolecularDynamics\TrajectoryUtils.py", line 214, in balance_index_groups
index_sets = split_index_sets(index_sets, max_size)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\MolecularDynamics\TrajectoryUtils.py", line 189, in split_index_sets
return [
^
File "C:\Users\...\PycharmProjects\MDANSE\MDANSE\Src\MDANSE\MolecularDynamics\TrajectoryUtils.py", line 189, in 
return [
^
File "C:\Users\...\AppData\Local\miniconda3\envs\MDANSE3\Lib\site-packages\more_itertools\more.py", line 4558, in chunked_even
for _ in islice(map(buffer.append, iterator), n, None, n):
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Step for islice() must be a positive integer or None.
```

**Suggested fix (optional)**
I think this issue is that the max chunk size variable below ends up being zero.

```python
def group_atom_indices(
    traj_instance: Trajectory,
    n_proc: int = 1,
    memory_scale_factor: int = 10,
    size_limit: int = CHUNK_SIZE_LIMIT,
) -> list[list[int]]:
    """Create groups of atom indices coaligned with the chunk size in the input file.

    ...
    """
    selected_indices = set(traj_instance.atom_indices)
    total_dimensions = traj_instance.variable("position").shape
    chunk_size = traj_instance.chunk_size(array_name="position")
    if chunk_size < 0:
        max_chunk_size = size_limit // (
            total_dimensions[0] * 3 * 8 * memory_scale_factor
        )
        chunk_limits = np.array([0, total_dimensions[1]])
    else:
        predicted_memory = (
            total_dimensions[0] * chunk_size * 3 * 8 * memory_scale_factor
        )
        downscale_factor = np.ceil(predicted_memory / size_limit)
        max_chunk_size = chunk_size // downscale_factor
        chunk_limits = np.concatenate(
            [np.arange(0, total_dimensions[1], chunk_size), [total_dimensions[1]]]
        )
    min_chunk_count = n_proc
    initial_sets = [
        sorted(selected_indices.intersection(range(*chunk_block)))
        for chunk_block in pairwise(chunk_limits)
    ]
    initial_sets = [index_set for index_set in initial_sets if len(index_set)]
    return balance_index_groups(
        initial_sets, max_size=int(max_chunk_size), min_count=min_chunk_count
    )
```

It also looks like the function `def group_atom_indices(...)` is quite slow to complete as well.

**Additional details (optional)**
Test on 2.0.1 with 3 frames and 2 correlation frames works fine.

MBartkowiakSTFC:
**Description of work**
The `group_atom_indices` function is split into smaller functions that can be tested better in the unit tests.

Additional checks are introduces to make sure that a usable grouping is returned also for large trajectories.

closes #1257

**Fixes**
1. Unit tests added for `group_atom_indices` and related functions.
2. An environment variable 'MDANSE_RAM_PER_PROCESS` can be used to increase the amount of memory per process that we aim for when grouping atoms. The default value is 512 (give in MB). **Note** there is no mechanism to guarantee that the right amount of memory will be used in the end. Also, it requires us to set the memory scaling factor correctly in the code of each analysis job, based on how much additional memory will be allocated by the job per atom.

**To test**
All tests should pass.
Try running the DISF analysis for a large trajectory. Check if memory use can be tuned using the `MDANSE_RAM_PER_PROCESS` environment variable.

### DIFF
--- a/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
@@ -116,15 +116,18 @@ class DynamicIncoherentStructureFactor(IJob):
         )
         n_proc = self.configuration["running_mode"].get("slots", 1)
 
+        self._nFrames = self.configuration["frames"]["n_frames"]
+
         self.grouped_indices = group_atom_indices(
-            self.trajectory, n_proc=n_proc, memory_scale_factor=vectors_per_shell
+            self.trajectory,
+            self._nFrames,
+            n_proc=n_proc,
+            memory_scale_factor=8 * vectors_per_shell,
         )
         self.numberOfSteps = len(self.grouped_indices)
 
         self._using_lattice_vectors = self.configuration["q_vectors"]["is_lattice"]
         self._nQShells = self.configuration["q_vectors"]["n_shells"]
-
-        self._nFrames = self.configuration["frames"]["n_frames"]
 
         self._instrResolution = self.configuration["instrument_resolution"]
 
--- a/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
+++ b/MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
@@ -120,9 +120,9 @@ class DynamicIncoherentStructureFactor(IJob):
 
         self.grouped_indices = group_atom_indices(
             self.trajectory,
-            self._nFrames,
+            self.configuration["frames"]["number"],
             n_proc=n_proc,
-            memory_scale_factor=8 * vectors_per_shell,
+            memory_scale_factor=6 * vectors_per_shell,
         )
         self.numberOfSteps = len(self.grouped_indices)
 
--- a/MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
+++ b/MDANSE/Src/MDANSE/MolecularDynamics/TrajectoryUtils.py
@@ -15,6 +15,7 @@
 #
 from __future__ import annotations
 
+import os
 from itertools import pairwise
 from typing import TYPE_CHECKING
 
@@ -24,7 +25,10 @@ from more_itertools import chunked_even
 if TYPE_CHECKING:
     from MDANSE.MolecularDynamics.Trajectory import Trajectory
 
-CHUNK_SIZE_LIMIT = 2**29  # 512 MB memory limit per process for now
+mb_ram_per_process = os.environ.get("MDANSE_RAM_PER_PROCESS", "512")
+CHUNK_SIZE_LIMIT = (
+    int(mb_ram_per_process) * 2**20
+)  # 512 MB memory limit per process for now
 
 from MDANSE.util_types import FloatArray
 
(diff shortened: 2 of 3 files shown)


ChiCheng45 (commented):
Works great, and `set MDANSE_RAM_PER_PROCESS=200000` seems to reduce the number of steps and speed things up. However, I found that using `set MDANSE_RAM_PER_PROCESS=200000` with eight cores ended up only using something like 500MB per process. Is there any way to get the number a bit closer?

Image

MBartkowiakSTFC:
Tests confirmed that a more complicated function will have to be used to predict the memory requirements of DISF. So far I adjusted the scaling factor so the memory prediction is close to correct for a 40000-frame, 34000-atom trajectory. But if I change frames from "in steps of 1, 20000 correlation frames" to "in steps of 10, 2000 correlation frames", the actual memory used by the calculation is about 70% of the predicted value.

After this PR we can start switching to more specialised, job-specific memory prediction calculations.

MDANSE/Src/MDANSE/Framework/Jobs/DynamicIncoherentStructureFactor.py
```diff
@@ -116,16 +116,19 @@ def initialize(self):
         )
         n_proc = self.configuration["running_mode"].get("slots", 1)
 
+        self._nFrames = self.configuration["frames"]["n_frames"]
+
```
ChiCheng45, line 120: I didn't find a difference in the number of steps of the job with a change in the "in steps of" option. `self.configuration["frames"]["n_frames"]` is the number of correlation frames; I think this needs to be `self.configuration["frames"]["number"]`.

MBartkowiakSTFC: I changed it now to use `self.configuration["frames"]["number"]`.

The current calculation tends to overestimate the required memory. In some cases for D2O (40000 frames, 34000 atoms, 256 atoms per trajectory chunk) I would end up using 2.5GB where the calculation expected 3.5GB.

ChiCheng45 (approved):
Looks good.

PR #1264, opened 2026-07-08
Set visibility to true on jobs with invisible parents

MBartkowiakSTFC:
**Description of work**
Some jobs like PairDistributionFunction are not visible in the GUI because they inherit `visible = False` from their parent class. This PR makes them visible.

**Fixes**
1. Set `visible = True` on PDF, VACF, PACF, DOS, PPS and coordination number jobs.

**To test**
Run the GUI. Check if any jobs are missing.


ChiCheng45 (approved):
Looks good.

oerc0122 (approved):
Whoops. Didn't spot this.

MBartkowiakSTFC:
> Whoops. Didn't spot this.

It took me a few days to realise that we used to have more analysis jobs before.

PR #1268, opened 2026-07-14
Chunked trajectory writer

MBartkowiakSTFC:
**Description of work**
This PR makes it possible to write a subset of atoms into the output trajectory, instead of a complete frame. It is a step towards lowering the RAM requirements as suggested in #893.

This PR is built on top of #1240.

**Fixes**
1. Added `write_array_fragment` to `TrajectoryWriter`.
2. Modified `TrajectoryEditor` and `TrajectoryFilter` to use `write_array_fragment` instead of `dump_configuration`.

**To test**
Run TrajectoryFilter and TrajectoryEditor using the protos branch and this branch. The results of each run should be the same for both branches. The code from this PR is expected to be faster and to require less RAM.


ChiCheng45 (approved):
Looks good to me, I don't see any issues.

PR #1270, opened 2026-07-21
Prevent InstrumentInfo from using an empty vector type

MBartkowiakSTFC:
**Description of work**
This PR stops InstrumentInfo from logging errors every time a new instrument is picked.

It seems that an instance of SimpleInstrument is passed to the visualiser before it has been fully configured, and the q vector type is set to `''`.  This PR adds a check for this.

**Fixes**
1. Ignore `SimpleInstrument` in `InstrumentInfo` if the q vector type has not been set.

**To test**
Run the GUI and go to the Instrument tab. Change the current instrument. There should be no error in the log.


ChiCheng45 (approved):
Looks good to me.

PR #1272, opened 2026-07-22
Improve the plot metadata preview.

MBartkowiakSTFC:
**Description of work**
The dataset path in plot preview is sometimes not displayed correctly. We get
Screenshot 2026-07-22 at 15 41 54
while the correct preview should be
Screenshot 2026-07-22 at 16 01 37

**Fixes**
`html.escape` is applied to the dataset path, as well as the values of the `attrs` dictionary. At the same time it is **not** applied to keys, allowing us to keep the formatting that displays each key in **bold** typeface.

**To test**
Generate analysis results using molecule grouping. Select the datasets in the plot creator and check if the dataset path is displayed correctly.


ChiCheng45 (approved):
Looks good to me.

PR #1275, opened 2026-07-30
Property viewer only updates when property viewer tab is selected and optimisations for a large number of atoms.

ChiCheng45:
**Description of work**
Previously, the property viewer would update even when the property viewer tab was not selected. This causes a slowdown when loading large trajectories and changing the frame for the first time, even when the property viewer tab is not open. Subsequent frame changes after the first slow one are fast.

Optimisation to reduce the amount of time selecting trajectories or analysis jobs for trajectories with a large number of atoms.

**To test**
Check that the property tab updates as usual. Load a large trajectory e.g. 500K atoms (to save disk space, you can test with a small number of frames), and change the frame; an update to the property viewer should not happen anymore, so it should be fast.

### DIFF
--- a/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
+++ b/MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
@@ -148,27 +148,36 @@ class ChemicalSystem:
             list of atom text labels from trajectory, by default None
 
         """
-        self._atom_indices = [
-            self.add_atom(self._database.get_atom_property(symbol, "atomic_number"))
-            for symbol in element_list
-        ]
+        # reduce calls to get_atom_property and create_rdkit_atom
+        # which can be expensive with a large number of atoms
+        atm_nums = {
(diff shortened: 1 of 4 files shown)


MDANSE/Src/MDANSE/Chemistry/ChemicalSystem.py
```diff
@@ -148,27 +148,36 @@ def initialise_atoms(
...
+            elem: self.create_rdkit_atom(atm_num) for elem, atm_num in atm_nums.items()
+        }
+
+        self._atom_indices = []
+        self._rdkit_dummy_atms = set()
+        for symbol in element_list:
+            atm_num = atm_nums[symbol]
+            idx = self.rdkit_mol.AddAtom(rdkit_atms[symbol])
+            self._atom_indices.append(idx)
+            if atm_num is None or atm_num == 0:
```
oerc0122: Can this reasonable by `if not atm_num`?

oerc0122 (approved):
Looks good if it does what it says., one ignorable suggestion.

PR #1282, opened 2026-08-14
Add sliding progress bar on job startup

MBartkowiakSTFC:
**Description of work**
A minor change to the GUI behaviour: when a job is starting and has not completed a single step, show a scrolling bar in the "Running Jobs" table with the text "STARTING". This is meant mainly to reassure the user that the specific job is being prepared.

**Fixes**
1. Add a `QTimer` to `JobEntry`, to control the progress bar movement in the starting phase.
2. Extend `ProgressDelegate` to handle negative values of progress.

**To test**
Start a converter or an analysis run.
Go to the "Running Jobs" tab and check if the "Starting" message is shown on the progress bar before the run starts.
Try pausing and resuming the run before and after the "Starting" message has disappeared. All the buttons from the context menu should work the same as before.


ChiCheng45 (approved):
Looks good to me.

PR #1283, opened 2026-08-19
Use read_atomic_trajectory_many in other analysis jobs

ChiCheng45:
**Description of work**
The execution time for some jobs are dominated by the file reading as the default chunking has 128 atoms and 1 frame per chunk. This make job which require all frames per step particularly slow, the 1 frame chunking is better for the trajectory conversion and some other analysis jobs e.g. PDF. A performance update was made in #1235 which speed up the DISF calculation, we should be able to apply a similar update to a lot of the other jobs.

**To test**
Tests should pass. Try a few jobs and compare timings against protos, this version should be faster.


MBartkowiakSTFC:
This may have some overlap with #1245

ChiCheng45:
Effectively a duplicate of #1245

PR #1292, opened 2026-09-01
Rework Units to put more work on class and use named tuple

oerc0122:
**Description of work**
- Rework Units to use the classes more effectively
- Replace anonymous 9-tuples with named tuple
- Remove `kg` from `units.json` since `kg` and `g` are strictly the same unit.
- Add detection for ambiguous units
- Make `get_trailing_digits` slightly more robust.
- Move some functions from `Units` into `IOUtils`.

**Fixes**
N/A

**To test**
Fiddling with units repeatedly, units editor and unit tests should pass.


ChiCheng45 (commented):
If I create a new unit kg in the MDANSE GUI.

```
"kg": {
    "factor": 1.0,
    "dimension": [
        1,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
    ]
}
```

Then this causes MDANSE crash with the error

```
MDANSE.Framework.Units.UnitError: More than one possible unit for kg.
```

I have to delete my cache to get it working again. This will also be an issue for users who have added new units, and their cache will contain a kg unit from previous versions.

PR #1265, opened 2026-07-08
Initial settings rework

oerc0122:
**Description of work**
Rework settings and session to allow local settings.

- Minor changes to use `@property` more extensively
- New `Settings` class outside of GUI (in `MDANSE/Core/Settings.py`)
- New `GUISettings` class to ease setting GUI parameters.
- Replace `forbidden_files` with sets
- Replace `LocalSession` and `FileSession` with single `Session` class
- Rename `TabbedWindow` to `MDANSEMainWindow`
- Improve distinction between `qt_settings` and `mdanse_settings` 

#### Usage
- `Settings` -- Singleton class which contains two paired dictionaries (`_settings`, `_defaults`):
  - `_settings` is dynamic and contains the custom settings which are saved
  - `_defaults` should mostly be static and values in `_defaults` are not saved (they are assumed to be inferred). 
  - `settings` (the public interface) defines a `Mapping[str, ChainMap(_settings, _defaults)]` style interface so updates to `settings` are recorded in `_settings` and `_defaults` is left untouched. 

  It  has several methods for getting values:
  - `get_opt(grp, key)` get the value of the parameter
  - `get_opt_w_default(grp, key, default)` get the value and if it does not exist, create a new **default** value.
  - `get_default(grp, key)` get the value of the parameter from `_defaults`

  And several for setting values:
  - `set_opt(grp, key, value)` update the parameter in `_settings` (and create the parameter if it does not exist)
  - `set_item(grp, key, value: Option)` set the parameter in `_defaults` for creating dynamic properties.
- `@Settings.parametrise` -- Define some parameters with saved settings on the class. 
  
  For `x = val`, `x` will be the property on the class and can be accessed directly via `.` notation. 
  Assigning to this value will update the value in settings (but currently not in the GUI settings editor). 
  If `val` is an `Option` these settings should be defined there, if not provided or `val` is just a value they will be inferred (`Option(value=val, group=type(class).__name__, name=x, comment=None)`).
  ```python
  @Settings.parametrise(
      favourite=Option("pike", group="fish")
  )
  class Test:
      def __init__(self):
          print(self.favourite) # => "pike"
          self.favourite = "trout"
          print(Settings.get_opt("fish", "favourite")) # "trout"
  ```

- `Option` defines the parameter in detail. 
- `Settings` -- contains two paired dictionaries (`_settings`, `_defaults`):
  - `_settings` is dynamic and contains the custom settings which are saved
[...]

PR #1295, opened 2026-09-03
Fix the q-vector viewer when the reciprocal space units are changed.

### LINKED ISSUE #1236: [BUG] 3D q-vector plot is flat in the Plot Holder with unit change
**Description of the error**
When I plot the vector positions, the 3D q-vector plot is flat when any reciprocal space unit is used other than 1/ang.

1/ang
Image

1/nm
Image

1/pm
Image

ChiCheng45:
**Description of work**
When the q-vectors shells are being viewed and the reciprocal space units are changed, all but one of the axes are changed, so the plot would be distorted. The axes and titles have also been updated and the box ratios of the q-vector plotting have been set to 1.

image

**Fixes**
Fixes #1236

**To test**
Plot the q-vectors shell and change the reciprocal space units; the 2D and 3D q-vectors plots should update correctly.


MBartkowiakSTFC:
Looks good!

On my laptop I had a minor problem of the axis labels of the top row overlapping with the plot titles in the bottom row:
vector_plot
But I expect that this is a combination of the screen resolution and matplotlib settings, and there is no easy way to prevent it from happening. I could easily fix it by setting `figure.subplot.hspace` to 0.3 (instead of the default 0.2).
vector_plot_better

ChiCheng45:
> But I expect that this is a combination of the screen resolution and matplotlib settings, and there is no easy way to prevent it from happening. I could easily fix it by setting `figure.subplot.hspace` to 0.3 (instead of the default 0.2). 

Yeah, that sounds good. Could do it in this PR. Where exactly is this hspace setting set in MDANSE?

MBartkowiakSTFC:
> > But I expect that this is a combination of the screen resolution and matplotlib settings, and there is no easy way to prevent it from happening. I could easily fix it by setting `figure.subplot.hspace` to 0.3 (instead of the default 0.2).
> 
> Yeah, that sounds good. Could do it in this PR. Where exactly is this hspace setting set in MDANSE?

It is one of the `matplotlib` settings:
Screenshot 2026-09-03 at 14 41 43

I expect that another user with a smaller screen will need it set to a different value. Do you think that instead of changing a value in this PR, we should encourage users to customise their matplotlib settings?

Also, independent of if we want to change the settings, or tell the users to change them themselves, it may be better not to do it in this PR, but wait until #1265 has been merged. Just in case it affects the way we set these parameters.

ChiCheng45:
> I expect that another user with a smaller screen will need it set to a different value. Do you think that instead of changing a value in this PR, we should encourage users to customise their matplotlib settings?
> 
> Also, independent of if we want to change the settings, or tell the users to change them themselves, it may be better not to do it in this PR, but wait until #1265 has been merged. Just in case it affects the way we set these parameters.

I figured it would be good to set some default values that should work for most users.

MBartkowiakSTFC:
The current mechanism of setting matplotlib parameters is located in `MDANSE_GUI.Widgets.PlotSettingsDialog`.

Normally, it imports `rcParams`, which should contain the default values, potentially with the changes applied by the user outside of MDANSE (as described here https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file). Next, it checks the `~/.mdanse/matplotlib.txt` file for any changes that had been applied by the user - these are the changes that the user made manually in MDANSE_GUI to values different from the defaults.

This structure does not make it easy to insert a parameter value that is our own preference. You may want to initialise a `matplotlib.txt` in the user's MDANSE settings if none exists, but this will only work for users who've never changed their matplotlib settings. What do you think?

ChiCheng45:
> The current mechanism of setting matplotlib parameters is located in `MDANSE_GUI.Widgets.PlotSettingsDialog`.
> 
> Normally, it imports `rcParams`, which should contain the default values, potentially with the changes applied by the user outside of MDANSE (as described here https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file). Next, it checks the `~/.mdanse/matplotlib.txt` file for any changes that had been applied by the user - these are the changes that the user made manually in MDANSE_GUI to values different from the defaults.
> 
> This structure does not make it easy to insert a parameter value that is our own preference. You may want to initialise a `matplotlib.txt` in the user's MDANSE settings if none exists, but this will only work for users who've never changed their matplotlib settings. What do you think?

Not sure; maybe it's better to come up with a solution in another PR? Anyway, the overlapping labels are seen with on protos as well, depending on the screen size.

image

MBartkowiakSTFC (approved):
Looks good, plots definitely look better this way.

PR #1298, opened 2026-09-04
Add a QSystemTrayIcon to MDANSE_GUI

### LINKED ISSUE #276: MDANSE_GUI tray icon and send to background on window close.
In some cases we may want to leave the GUI running in the background during a MDANSE job.

Similarly to other programs it might be good to allow the MDANSE_GUI to be sent to the background but leave it running. The MDANSE_GUI window will close but will be still running. The GUI will need to be closed from the tray icon. On window close it may be worth asking the user whether it want to send the process to the background or to close it entirely.

**Additional context**
See for example programs like Zoom which do not exit on window closing.

### LINKED ISSUE #504: [ENHANCEMENT] MDANSE_GUI close; running job behaviour.
When the GUI is closed job continues to run in the background.

When MDANSE_GUI is closed with jobs running a popup should appear asking whether the user wants to cancel all jobs. If the user doesn't want to cancel all jobs then we should send MDANSE_GUI to the system tray (#276) or stop MDANSE_GUI from closing.

MBartkowiakSTFC:
**Description of work**
Prevents the GUI from exiting while tasks are still running. Adds a system tray icon that can be used to restore the main window.

closes #276
closes #504

**Fixes**
1. Create a QSystemTrayIcon when MDANSE_GUI is started. This can be prevented by starting with `mdanse_gui --no-systray`.
2. If the system tray icon has been requested, replace the "Exit" in main window's dropdown menu with "Hide window".
3. Only allow the GUI to exit if no jobs are running and no trajectories are loading.
4. Additionally, if the system tray icon exists, only allow the GUI to exit if the main window is hidden.
5. MDANSE_GUI produces system tray notifications when jobs finish. (Should we add an option of switching it off, in case it gets too annoying?)

**To test**
Please test both the `mdanse_gui` and `mdanse_gui --no-systray` cases.
Check that the GUI can exit only if no trajectories are being loaded and no jobs are running (valid for runs with and without the system tray icon).
Check that the GUI can be closed through the system tray icon context menu only if the main window is hidden.
Check that the main window can be restored after closing if the system tray icon is present.


MDANSE_GUI/Src/MDANSE_GUI/main.py
```diff
@@ -66,6 +66,11 @@ def build_parser():
         action="store_true",
         help="Do not display splash screen on startup.",
     )
+    parser.add_argument(
+        "--no-systray",
+        action="store_true",
+        help="Do not create a system tray icon.",
```
oerc0122: Feel like this could be better described. Does this also prevent closure with running jobs?

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
```diff
@@ -114,10 +115,16 @@ def __init__(
         parent=None,
         title="MDANSE",
         settings=None,
-        app_instance=None,
+        app_instance: QApplication | None = None,
+        create_systray_icon: bool = True,
+        systray_icon: QIcon | None = None,
```
oerc0122, line 119: Does this work if it's `None`?

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
```diff
@@ -190,6 +198,53 @@ def __init__(
...
+            else:
+                self.tray_menu.addAction(label, func_slot)
+        self.system_tray_icon.setContextMenu(self.tray_menu)
+        self.system_tray_icon.setVisible(True)
+
+    @Slot(object)
+    def systray_message_job_finished(self, job_details: tuple[str, str]):
+        self.system_tray_icon.showMessage(
+            "MDANSE run finshed", f"{job_details[0]} {job_details[1]}"
+        )
```
oerc0122: Think this adds a bit of clarity and undestanding.
```suggestion
        name, status = job_details
        self.system_tray_icon.showMessage(
            "MDANSE run finshed", f"{name} {status}"
        )
```

MDANSE_GUI/Src/MDANSE_GUI/TabbedWindow.py
```diff
@@ -203,6 +258,45 @@ def changeEvent(self, event):
...
+                return
+            if self.system_tray_icon.isVisible():
+                self.hide()
+                return
+        elif self.can_be_closed:
+            return super().closeEvent(event)
+        else:
+            event.ignore()
+            self.block_gui_shutdown()
+
```
oerc0122, line 307: The logic here is a little confusing, but that might just be me.

oerc0122 (commented):
Looks good and works on my i3 setup. Just a couple of things.

ChiCheng45 (commented):
Looks good works on windows. Can we add a double click of the tray icon to restore the GUI window?

I also wondering if the current users might not be aware that MDANSE will be still open after this change. Do we need some notification which is only shown once to let them know that MDANSE is still running? I think its probably fine without one.

oerc0122:
I also think we might want to allow settings to save configuration as to whether it stays alive/goes to the tray. Though we might want to wait until the settings rework hopefully makes that easier.

MDANSE_GUI/Src/MDANSE_GUI/main.py
MBartkowiakSTFC: The GUI should prevent you from closing independent of whether the system tray icon exists or not.

MBartkowiakSTFC:
BTW it is apparently impossible to generate a double-click event for a system tray icon on a Mac, if the system tray icon already has a context menu: https://doc.qt.io/qt-6/qsystemtrayicon.html#ActivationReason-enum
But I hope that testing on Windows and Linux will show that it works.

ChiCheng45:
> BTW it is apparently impossible to generate a double-click event for a system tray icon on a Mac, if the system tray icon already has a context menu: https://doc.qt.io/qt-6/qsystemtrayicon.html#ActivationReason-enum But I hope that testing on Windows and Linux will show that it works.

Double-click is working on Windows now. One other issue: if the mdanse_gui window is open or minimized and I right-click the tray icon to quit, then MDANSE doesn't exit. It only exits when I close the GUI and then quit via the tray icon.

MBartkowiakSTFC:
> > BTW it is apparently impossible to generate a double-click event for a system tray icon on a Mac, if the system tray icon already has a context menu: https://doc.qt.io/qt-6/qsystemtrayicon.html#ActivationReason-enum But I hope that testing on Windows and Linux will show that it works.
> 
> Double-click is working on Windows now. One other issue: if the mdanse_gui window is open or minimized and I right-click the tray icon to quit, then MDANSE doesn't exit. It only exits when I close the GUI and then quit via the tray icon.

I think the current version give the system tray Quit entry more power, so it can exit when the GUI window is visible too.

ChiCheng45:
> > > BTW it is apparently impossible to generate a double-click event for a system tray icon on a Mac, if the system tray icon already has a context menu: https://doc.qt.io/qt-6/qsystemtrayicon.html#ActivationReason-enum But I hope that testing on Windows and Linux will show that it works.
> > 
> > 
> > Double-click is working on Windows now. One other issue: if the mdanse_gui window is open or minimized and I right-click the tray icon to quit, then MDANSE doesn't exit. It only exits when I close the GUI and then quit via the tray icon.
> 
> I think the current version give the system tray Quit entry more power, so it can exit when the GUI window is visible too.

Thanks looks good. Last one, I think: double-clicking works when the window is closed to the system tray, but the window stays minimised when double-clicking the system tray icon. I find that other apps like Spotify, NVIDIA driver, etc., unminimise after double-clicking the system tray icon.

PR #1304, opened 2026-09-09
Improved plot colouring and CoM mixed atom colours

### LINKED ISSUE #1133: [ENHANCEMENT] Improve plot colouring options in single plot
I'm trying to make a plot for the liquid argon van Hove function, works fine, but with the current colouring scheme makes the colours quite similar and is quite difficult to tell the difference. 

Image

Perhaps we could connect the colour that is used for multiple line single plots to the colormap setting and have the color bar included?

ChiCheng45:
**Description of work**
Updated the plot holder so that it now has two colour and one colour map options.

image

For Grid, Grouped, and Single plots, two colours are used to create a colour range, which is used to colour each curve. Previously, it would colour the plot curve from the main colour to gray.

image

I have also added an optional dependency to use mixbox to generate the colour range using pymixbox, which generates paint-like blending between the colours.

image

On the heatmap plot, different heatmaps can now be used for the different datasets.

image

The CoM Trajectory job has now been updated so that the CoM particle colour is a mixture of the molecules' atom colours.

image

image

With pymixbox, which gives better mixing.
image

For the opposite case.
image 

image

With pymixbox
image

**Fixes**
Closes #1133

**To test**
Test the CoM trajectory job and the plotter with different colour and colour map settings. Test with and without pymixbox installed.


ChiCheng45:
Currently, the non-Pymixbox implementation uses some basic averaging rules to produce the colours, so intermediate colours end up not being the expected ones; see the CoM trajectory above. I could also try to implement a more complicated procedure so it does something more similar to pymixbox, but I think just using pymixbox would be better.

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -973,16 +979,22 @@ def create_average_atom(
...
+            for i in temp:
+                r, g, b = i[0].split(";")
+                colours.append((int(r), int(g), int(b)))
+
+            weights = []
+            for element_name, element_count in atom_dictionary.items():
+                # weight color mixture by the atom volume
+                weights.append(
+                    float(database.get_atom_property(element_name, "vdw_radius") ** 3 * element_count)
+                )
```
```suggestion
            # weight color mixture by the atom volume
            weights = [
                float(
                    database.get_atom_property(element_name, "vdw_radius") ** 3 * element_count
                ) 
                for element_name, element_count in atom_dictionary.items()
            ]
```

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -973,16 +979,22 @@ def create_average_atom(
...
+
+            rgb = mix_colors(colours, weights)
+            total = f"{rgb[0]};{rgb[1]};{rgb[2]}"
```
```suggestion
            total = ";".join(map(str, rgb))
```

MDANSE/Src/MDANSE/MolecularDynamics/Trajectory.py
```diff
@@ -973,16 +979,22 @@ def create_average_atom(
...
+
         values[property] = total
     is_dummy = 1
     for element_name, _ in atom_dictionary.items():
```
```suggestion
    for element_name in atom_dictionary:
```

MDANSE_GUI/Src/MDANSE_GUI/Tabs/Plotters/Grid.py
```diff
@@ -154,6 +154,14 @@ def plot(
         grid_size = self.grid_size(self.n_curves)
         gs = self._figure.add_gridspec(*grid_size)
 
+        colours = {}
+        for databundle in plotting_context.datasets().values():
+            colours[databundle.row] = self.colours(
+                databundle.colour_1,
+                databundle.colour_2,
+                min(self._plot_limit, databundle.dataset.n_curves),
+            )
```
oerc0122:
```suggestion
        colours = {databundle.row: self.colours(
                databundle.colour_1,
                databundle.colour_2,
                min(self._plot_limit, databundle.dataset.n_curves),
            )
        for databundle in plotting_context.datasets().values()}
```

MBartkowiakSTFC:
I have not found any problems so far. I still need to try out the centre of mass colour mixing.

One idea: should the datasets start with "Colour 2" different from "Colour 1"? At the moment the users is always required to set the "Colour 2" manually to be able to tell the difference between the curves. To stick to our principle of reducing the number of clicks needed to get a result, we should come up with a setting that gives the curves different colours from the start. Then the user is still free to customise further, but there is a chance that the original setting already works for them.

ChiCheng45:
> One idea: should the datasets start with "Colour 2" different from "Colour 1"? At the moment the users is always required to set the "Colour 2" manually to be able to tell the difference between the curves. To stick to our principle of reducing the number of clicks needed to get a result, we should come up with a setting that gives the curves different colours from the start. Then the user is still free to customise further, but there is a chance that the original setting already works for them.

Could do; I stuck to making it the same colour, as I am not really sure what the best colour for colour 2 should be, perhaps the inverse of colour 1?

MBartkowiakSTFC:
> > One idea: should the datasets start with "Colour 2" different from "Colour 1"? At the moment the users is always required to set the "Colour 2" manually to be able to tell the difference between the curves. To stick to our principle of reducing the number of clicks needed to get a result, we should come up with a setting that gives the curves different colours from the start. Then the user is still free to customise further, but there is a chance that the original setting already works for them.
> 
> Could do; I stuck to making it the same colour, as I am not really sure what the best colour for colour 2 should be, perhaps the inverse of colour 1?

We definitely want these colours to be different, so either the inverse of colour1, or you could cycle the RGB values so that if colour 1 is RGB, then colour 2 is BRG. It is not easy to predict which will work better, so I'm happy with either for now.

ChiCheng45:
> > > One idea: should the datasets start with "Colour 2" different from "Colour 1"? At the moment the users is always required to set the "Colour 2" manually to be able to tell the difference between the curves. To stick to our principle of reducing the number of clicks needed to get a result, we should come up with a setting that gives the curves different colours from the start. Then the user is still free to customise further, but there is a chance that the original setting already works for them.
> > 
> > 
> > Could do; I stuck to making it the same colour, as I am not really sure what the best colour for colour 2 should be, perhaps the inverse of colour 1?
> 
> We definitely want these colours to be different, so either the inverse of colour1, or you could cycle the RGB values so that if colour 1 is RGB, then colour 2 is BRG. It is not easy to predict which will work better, so I'm happy with either for now.

Could do, just one thing I do find that if colour 1 and 2 are always different, then the plots can be quite noisy and difficult to read if there are multiple plots on Single.

same colour

image

vs

some randomly selected colours.

image

Obviously, this is fine with a grouped plot.

image

MBartkowiakSTFC:
> 
> Could do, just one thing I do find that if colour 1 and 2 are always different, then the plots can be quite noisy and difficult to read if there are multiple plots on Single.
> 

I'm afraid we will never be able to come up with an automatic colour selection scheme that will cover all the cases correctly. Sooner or later the user will have to invest some effort into customising the plot. My assumption was that if the colours 1 and 2 are the same, then the user will _always_ have to change the colours, just to be able to tell the curves apart. If it turns out that with the other scheme the user still has to change colour 2 manually, then we haven't really lost anything.

Anyway, let's not spend too much time on this. I'm happy with any scheme of generating colour 2, as long as it is not identical to colour 1. The main point of this PR is that the user can change the colours, and this is already working fine.

ChiCheng45:
> > Could do, just one thing I do find that if colour 1 and 2 are always different, then the plots can be quite noisy and difficult to read if there are multiple plots on Single.
> 
> I'm afraid we will never be able to come up with an automatic colour selection scheme that will cover all the cases correctly. Sooner or later the user will have to invest some effort into customising the plot. My assumption was that if the colours 1 and 2 are the same, then the user will _always_ have to change the colours, just to be able to tell the curves apart. If it turns out that with the other scheme the user still has to change colour 2 manually, then we haven't really lost anything.
> 
> Anyway, let's not spend too much time on this. I'm happy with any scheme of generating colour 2, as long as it is not identical to colour 1. The main point of this PR is that the user can change the colours, and this is already working fine.

Alright. Anyway, I updated it so that colour 2 is the inverse of colour 1, which is probably the best/easiest.

MBartkowiakSTFC:
Additional testing shows that I cannot plot individual q-vector shells with the current version. Plotting fails with this error output:
```
  File "MDANSE_GUI/InputWidgets/QVectorsWidget.py", line 269, in set_shell
    self.set_upper_limit(len(vec_dict))
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "MDANSE_GUI/InputWidgets/QVectorsWidget.py", line 249, in set_upper_limit
    self.set_shell(current_shell, update_limit=False)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MDANSE_GUI/InputWidgets/QVectorsWidget.py", line 299, in set_shell
    self.plot_widget.set_plotter(self.plotter_type)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "MDANSE_GUI/Tabs/Visualisers/PlotWidget.py", line 260, in set_plotter
    self.plot_data()
    ~~~~~~~~~~~~~~^^
  File "MDANSE_GUI/Tabs/Visualisers/PlotWidget.py", line 345, in plot_data
    self._plotter.plot(
    ~~~~~~~~~~~~~~~~~~^
        self._plotting_context,
        ^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        toolbar=self._toolbar,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "MDANSE_GUI/Tabs/Plotters/Vectors3D.py", line 140, in plot
    color=databundle.colour,
          ^^^^^^^^^^^^^^^^^
AttributeError: 'PlotArgs' object has no attribute 'colour'. Did you mean: 'colour_1'?
```

ChiCheng45:
> Additional testing shows that I cannot plot individual q-vector shells with the current version. Plotting fails with this error output:
> 
> ```
> Traceback (most recent call last):
>   File "MDANSE_GUI/InputWidgets/QVectorsWidget.py", line 269, in set_shell
>     self.set_upper_limit(len(vec_dict))
>     ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
>   File "MDANSE_GUI/InputWidgets/QVectorsWidget.py", line 249, in set_upper_limit
>     self.set_shell(current_shell, update_limit=False)
>     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "MDANSE_GUI/InputWidgets/QVectorsWidget.py", line 299, in set_shell
>     self.plot_widget.set_plotter(self.plotter_type)
>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
>   File "MDANSE_GUI/Tabs/Visualisers/PlotWidget.py", line 260, in set_plotter
>     self.plot_data()
>     ~~~~~~~~~~~~~~^^
>   File "MDANSE_GUI/Tabs/Visualisers/PlotWidget.py", line 345, in plot_data
>     self._plotter.plot(
>     ~~~~~~~~~~~~~~~~~~^
>         self._plotting_context,
>         ^^^^^^^^^^^^^^^^^^^^^^^
>     ...<2 lines>...
>         toolbar=self._toolbar,
>         ^^^^^^^^^^^^^^^^^^^^^^
>     )
>     ^
>   File "MDANSE_GUI/Tabs/Plotters/Vectors3D.py", line 140, in plot
>     color=databundle.colour,
>           ^^^^^^^^^^^^^^^^^
> AttributeError: 'PlotArgs' object has no attribute 'colour'. Did you mean: 'colour_1'?
> ```

Thanks, I think this should be fixed now.

MBartkowiakSTFC (approved):
I tried plotting different things in the main plotter and in the dialogs (resolution, trajectory filter, q vectors). I think it's working fine. Also, plots with default settings look better than before.

Issue #1004, opened 2025-10-02
[BUG] Adjust display settings for high-DPI screens


MBartkowiakSTFC:
**Description of the error**
On some displays, the default settings make the fonts come out too small.

**Describe the expected result**
We want the GUI to be immediately legible on all screens.

**Describe the actual result**
So far, we have had reports from Linux/Wayland users with high-DPI screens that the initial settings were making the text to small to read.

**Suggested fix**
Two options:
1. Qt environment variables can be set to take this kind of display device into account.
2. If this does not help, an extra style sheet should be added just for changing the font size.

**Additional details**
N/A

Issue #414, opened 2024-04-17
[BUG] Problems starting MDANSE_GUI on openSUSE with Wayland


MBartkowiakSTFC:
**Description of the error**
The standard installation of MDANSE_GUI from pypi results in a situation where the mdanse_gui script fails to create the main window on an openSUSE Linux machine with Wayland.

**Describe the expected result**
The GUI should start on every platform without extra tuning.

**Describe the actual result**
Apparently, an additional environment variable is needed:
`QT_QPA_PLATFORM=xcb  mdanse_gui`
This makes MDANSE_GUI start correctly.

**Suggested fix**
Normally, Qt has a number of plugins it can use, but for MDANSE_GUI it seems that xcb is the only option.

**Additional details**
N/A

PR #1306, opened 2026-09-11
Check platform on GUI startup and output warnings about known issues

MBartkowiakSTFC:
**Description of work**
The main purpose of this PR is to check the Qt platform before creating any widgets, so that warnings can be output if the platform is connected to known problems. This is connected to #1004 and #414

**Fixes**
1. Use `QApplication` class methods to get information about the Qt platform before creating the GUI.
2. Connect the `QApplication.aboutQt` method to an entry in the main GUI menu. 

**To test**
On a Linux computer with Wayland, check if the warnings are output in the terminal before the GUI fails due to missing support for Wayland.


jamescrake-merani (approved):
Just tested on my machine, and the warning comes up as expected. If I could make one suggestion though: most applications which don't support Wayland will just open as an X11 windows without necessarily telling the user that they failed to launch as Wayland. Additionally, some applications only support Wayland experimentally, and Wayland support has to be manually enabled (e.g. when Jetbrains IDEs provided Wayland, it was marked experimental, and opt in only) otherwise it falls back to X11.

MDANSE looks fine on my display running under the XWayland compatibility layer. The one time where it can be an issue is when using fractional scaling which is something X11 always struggled with, so in that case the window will appear blurry.

With all of that in mind, I would suggest just forcing MDANSE to run on X11 by default on Linux until you can get VTK running properly on Wayland.

oerc0122 (approved):
Seems to work as desired. Looks good.

